PR #6, opened 2013-10-26
Okhin - Adding some personnalisation

okhin:
Added two new routes, one to memorize snippets whcih have been clicked by users (the /go one) and one to query the database of savec snippets, and their association with keywords (the /lsearch) to be used as a local search engine.

For now the personnalisation is not extremely smart, but results appears in the search page under the engine 'local'. I keep a score in database (number of time a snippets has been used for a specific keyword), but I'm not using it right now. Need to find a way to have a score from some search engines.

Everything is saved in a sqlite database whcih needs to be initilaised first by running this script:

from searx.webapp import init_db
init_db()


ibrokemypie:
bit of a necro, but why not just keep it a fork? of course many people want the privacy, but other people want the personalization, should be up to the individual.

PR #59, opened 2014-03-25
patch displays how many commits an instance is behind the master repo

stef:
possibly needs some polishing. however ghcheck.py should be working autonomously, can be called from shell even.


ghost:
So you will be constantly checking github.com. I don't think this is a good idea to display which security holes have been patched and which ones have not been (you don't put it this way, but you can easily get this info by looking at the commits and compare), and not every instance runs in git. I'm currently working on getting searx packaged for Guix (a system comparable to Nix) and we remove every version control possibility of packaged software there (ie: a software in the install process is received from a git source, the .git dir in the source is removed).
If this gets merged this should be optional, not enabled in the default config.

I see this was old, sorry for bumping old threads.

unixfox:
@dalf it's already implemented right? Well the searx instance display its commit number so I guess it's safe to close this PR.

dalf:
This PR is about something else: 
* the new URL `/updates` displays how many commits an instance is behind the master repo
* more or less [this](https://github.com/searx/searx/compare/v0.18.0...master) but from the last local git commit on the master branch.

----

The searx docker image displays the current local commit id:
https://github.com/searx/searx/blob/b661c3f5d4c3883454b7c403f4602dc93e8b8fd2/manage.sh#L125-L134
https://github.com/searx/searx/blob/b661c3f5d4c3883454b7c403f4602dc93e8b8fd2/Dockerfile#L69-L71

PR #137, opened 2014-12-06
[enh] implement image cache using flask-images, close #112

### LINKED ISSUE #112: implement image-proxy (fetch and mininize images)
currently images are provided in the result-site, without data compression or ssl-encryption.

My Idea is that searx fetch the result images, minimize it if possible and store the images for a short time localy.

pointhi:
I have implemented an image caching system.

If enabled, searx is automatic loading the images, and resizing it to the required size. Furthermore the images are provided by searx, so the website admins cannot see what users watch their images.

thx, pointhi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 flask
 flask-babel
+flask-images
 requests
 lxml
 pyyaml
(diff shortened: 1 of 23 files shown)


pointhi:
now, this feature is mostyl implemented by @asciimoo. Thanks

PR #143, opened 2014-12-09
Use human readable date for infobox

Cqoicebordel:
For DoB and DoD, wikipedia use a non standard ISO format, not easily readable.
Now the date is displayed in an human readable form, using the language setting as locale if available. If not, it uses the default locale.

It should be reworked when full internationalization will be in place, but it's a nice first step.


asciimoo:
Seems, it works, thanks!

PR #160, opened 2014-12-20
[enh] add result_templates/code.html

pointhi:
I have implemented a code highlighter based on pygments as said there: https://github.com/asciimoo/searx/pull/158#issuecomment-67746828

Code highlighting is currently only enabled for a few file-endings. The rest is printed as plain-text as before.

furthermore, I have done some pep8 stuff, and improving the oscar template


asciimoo:
Okay, it seems, i successfully rewrote the static files and template handling and restructured the static content to be able to handle common static content and result templates, so now this feature could be simplified to use a single template file (eg. `templates/result_templets/code.html`) and css (eg. `static/css/pygments.css`)

pointhi:
ok, I have merged the master to make that pull-request up-to-date

In future I could improve ordering of static-files. It shouldn't be a big problem do it.

PR #163, opened 2014-12-27
[enh] update translations

pointhi:
I have update the translations on transifex, and also inside searx (this pull request)

Also I have added a turkish translation, done by Caner Başaran on transifex.

@asciimoo  It is nice if you translate missing words, but please do not use only a translator :).
For Example: en:"Leecher" --> de"Zecke" I think nobody in germany/austria would name a p2p leecher "Zecke" (en:"Tick"), even when it is a good paraphrasing ^^


asciimoo:
Thanks the corrections

PR #164, opened 2014-12-28
improve Favicons

pointhi:
because of the favicon detection the oscar template was broken.

I have changed all favicons path to the subdir `/icons` like in the oscar_template, because I think it is much cleaner using a subfolder.

furthermore I have added some new icons to the oscar template


asciimoo:
great, thanks!

PR #165, opened 2014-12-28
Moar engines

Cqoicebordel:
Twitter corrections + Pubdate
Digg, Google Play Apps, Movies and Music

I put digg in social and news, but maybe it's only news. Tell me and I'll correct it.

### DIFF
--- a/searx/engines/500px.py
+++ b/searx/engines/500px.py
@@ -35,9 +35,9 @@ def request(query, params):
 # get response from search-request
 def response(resp):
     results = []
-    
+
     dom = html.fromstring(resp.text)
-    
+
     # parse results
     for result in dom.xpath('//div[@class="photo"]'):
         link = result.xpath('.//a')[0]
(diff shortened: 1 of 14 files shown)


asciimoo:
New engines and pep8 \o/
Thanks!
The bug sounds strange, never noticed the same - some test would be good..

Cqoicebordel:
After some tests, I still have this bug, and I don't know what causes it.

I thought it was the "tail" issue in lxml as described in http://lxml.de/tutorial.html at "Elements contain text", and actually, I still do, but I don't understand why nobody else has it.

Anyway, there are three options here : 
- Fuck it : I'm the only one to have it, so it's not important.
- Let's do something, just in case : let's change every `join` to `tostring(..., method='text')`, it should be cleaner, and we never know.
- It's really important, let's investigate and find what the hell is going on.

You know that I'm a newbie in almost every tech used in this project, so it's possible that I fucked up somewhere. 
But I can't imagine where...

Issue #154, opened 2014-12-17
Non exclusif search with !bang


Cqoicebordel:
I'm still trying to solve the issue with the google engine, and the images result (BTW, I call dibs on the coding ;) ) : 
When in the result of google we find images, google send only the links to the thumbnails, so we deactivated the images in the results. But google finding image in itself is useful.

So the one way of doing things is to have a "system" result, indicating that google found images, and proposing to relaunch the search including images (google images only or category image, to be choosen). So we must modify the query to add the !goi (or !images), so that the user can see, learn and modify the query appropriately. Three case scenarios : 
1. The original query was `!go foobar`. In that case, adding `!goi` or `!images` will work.
2. The orginal query was `!general foobar`. In that case too, adding `!goi` or `!images` will work.
3. The probably most frequently original query was `foobar` in the categorie general. In that case, adding `!goi` or `!images` won't work, because it will deactivate the general category.

So my questions/propositions are :
1. Should we implement a non exclusive (non deactivating categories) bang, maybe with '?' prefix ?
2. Should we modify the current situation by removing the exclusivity in '!' ? (I think not)
3. Should we do nothing ? (why not, it's not that important)
4. Another way ?

asciimoo:
Yap, sounds useful, prefix should be `-` or `?`

PR #166, opened 2015-01-03
Add non exclusif bang

Cqoicebordel:
Allow to perform a search while adding an engine (or a category) without adding it "officially" to the request.
'?' is used to add an engine without modifying anything else to the request. For example, you can perform a search in the 'general' category, and if you add '?tw' the result from Twitter will be added to the originals results.

Solve #154
'?' was chosen because '-' can be used to eliminate search terms from the majority of search engines.


asciimoo:
Thanks

PR #167, opened 2015-01-03
Minor fixes

Cqoicebordel:
Few digg and vimeo bug fixes.

Note that the Vimeo bug was due of a HTTP 301 redirect from https to http from Vimeo, which wasn't handled by searx.
I tried to add an allow_redirects in the GET params, but as the engines rewrite the response(), it wasn't working.
I think it could be useful to handle this kind of redirects. Maybe by extending the original response() function instead of rewriting it. But I have no idea of doing this...

### DIFF
--- a/searx/engines/digg.py
+++ b/searx/engines/digg.py
@@ -44,6 +44,9 @@ def response(resp):
 
     search_result = loads(resp.text)
 
+    if search_result['html'] == '':
+        return results
+
     dom = html.fromstring(search_result['html'])
 
     # parse results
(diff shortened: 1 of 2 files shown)


asciimoo:
It's strange, because requests automatically handles HTTP redirects: http://requests.readthedocs.org/en/latest/user/quickstart/#redirection-and-history - requires deeper debugging.
Anyway, thanks the fixes!

Cqoicebordel:
Yes I agree, but I got a resp.text equal to None, and resp.response equal to 301. And in browser, https was indeed redirected to http.
Anyway, I have no more clues on my side.

PR #168, opened 2015-01-05
Integrated media in results + Deezer Engine

Cqoicebordel:
New "embedded" item for the results, allow to give an iframe to display the media directly in the results.
Note that the attributes src of the iframes are not set, but instead data-src is set, allowing to only load the iframe when clicked.

Deezer engine based on public API (no key).

EDIT : I forgot two things : I added some infos in a few engines (content and or pubdates), and I didn't yet took care of i18n. I'm fuzzy about the steps to follow to do that. So wait a bit until I take care of that to merge this pull request.


Cqoicebordel:
Ok, I think It's ok here.
Of course, some things should be taken care of over at Transifex, but I leave that to the admin ;)

pointhi:
the fuzzy mechanism of the translation generation has generated some new gabage translations. Probably changing the line:

``` shell
pybabel update -i messages.pot -d $SEARX_DIR'/translations/' -l $f
```

https://github.com/asciimoo/searx/blob/master/utils/update-translations.sh#L12

_to_

``` shell
pybabel update -N -i messages.pot -d $SEARX_DIR'/translations/' -l $f
```

for deactivating the fuzzy mechanism. _(this change is not tested by me yet)_

http://babel.edgewall.org/wiki/Documentation/cmdline.html#update

Cqoicebordel:
Good catch.
I had to remove by hand the fuzzy lines, but beyond that, I recreated the translations, and everything went fine.

asciimoo:
That's a user friendly feature, but i'm not sure that embedding (more) third party content by default is a good idea - maybe it would fit better to searxs philosophy if it could be enabled by the users, like the autocompletition.

Cqoicebordel:
I agree that it was questionnable.
But the way I did it, nothing is included until you click on "show xxx". No request is send to youtube/dailymotion/etc. Nothing. No service can know that something is included until you click on "show xxx".
So the way I see it, it's exactly the same thing as clicking on the result link, but without leaving searx. 
Even better : youtube uses the "nocookie" domain name www.youtube-nocookie.com, so it's even less tracking that clicking on the link.

The difference with autocomplete is that autocomplete needs more requests, and so, more tracking. That's not the case here.
But an easy way to put it as an option, is to modify the theme.

asciimoo:
ah, okay, i got it (the iframe `data-src` attribute hack), cool, thanks!

PR #170, opened 2015-01-10
Little fixes, add searx-autocompletion, more informations about torrents

pointhi:
- rss-link in meta-section of html, which can be detected by browser (inside result-page)
- searx-autocompleter, which can handle the !bang syntax
- better visualisation of torrent results (filesize, number of files, .torrent file)

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -52,7 +52,7 @@ def searx_bang(full_query):
 
             # check if query starts with engine name
             for engine in engines:
-                if engine.startswith(engine_query):
+                if engine.startswith(engine_query.replace('_', ' ')):
                     results.append('!{engine}'.format(engine=engine.replace(' ', '_')))
 
             # check if query starts with engine shortcut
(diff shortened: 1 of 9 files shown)


Cqoicebordel:
That seems nice to me !
A few comments though (and feel free to hate me for those ;) ) :
- I feel like there is a lot of similitude between the parse query in autocomplete, and the one in query. Is there a way to refactor this ?
- For the Kickass engine, nice again ! To be honest, I was lazy on that one... But what I stumbled upon when looking to implement other torrent engines is that almost none of them provide magnet links. So it could be good to make it optionnal too.
- You use two <p> in torrent result, with both the content class. I haven't tested it, but doesn't it make it take a lot of place ? (read the next one before testing anything on your side)
- One of the pull request I wanted to send was adding the magnet link to the tool bar (archive/show media/show map etc.). Since you are in there you could add it yourself, and add the torrent one too.

That's all. I stop bothering you now ;)

(PS : Nice for the autoescape too !)

asciimoo:
wow, super useful, especially the autocompletition! Thanks!
@Cqoicebordel yes, the code around query parsing is really messy in whole searx, needs a refactor

PR #172, opened 2015-01-11
Torrents w3c+ux+rss

Cqoicebordel:
A few fixes across Torrent and RSS
Magnets and Torrents files optionnal and in tool bar
Lots of w3c fixes
Send the right URL in RSS


asciimoo:
Thanks!

Issue #171, opened 2015-01-10
Oscar as default template


Cqoicebordel:
This one is for a discussion :  
I feel that the Oscar template is way better than the two others. Maybe not graphically, but at least by the UX, and by the tools provided.

So my question : Shouldn't it be the default template for a new searx user ?

dimqua:
It may be better to improve the two others instead. For instance, courgette theme does not support DDG definitions.

asciimoo:
@Cqoicebordel currently it shouldn't, mainly because it doesn't fully functional without javascript. With javascript it's way more usable than the current default theme, but using noscript you can't even select categories.

@dimqua yap, both other themes requires improvements/fixes..

I'll finish soon with the static content refactor and then all the 3rd party client side libs will be available for all themes

pointhi:
@asciimoo the truth with js-compalibity is: you can select categories, but it isn't visualised in the theme without js yet. Probably I can add a little css-hack to fix that.

Cqoicebordel:
@asciimoo It could also be useful to do a JS-free template, to allow those without JS activated to use a nice version of searx.
But graceful degradation should indeed be implemented. If @pointhi, you know how to do it, it would be perfect :)

asciimoo:
@pointhi it would be great
javascript independence is always a good goal especially in the case of searx.

@Cqoicebordel agree!

asciimoo:
Okay, experimentally the default theme changed to oscar.
@pointhi what do you think about removing the javascript warning? And maybe a less harsh logo?

pointhi:
I think removing the javascript-warning except in the preferences wouldn't be a problem. Also, the warning should appear if the screensize is to small, because the menue on top doesn't work wihout it.

Because of the logo, I'm not sure. I think the magnifying glass in the logo should remain. What's your idea for a logo? The same as in the other themes, or somthing new?

In the next time, I would add a feature to change the look of the oscar-theme, using a few themes found on http://bootswatch.com/. Probably we could make searx more colourful with it :).

asciimoo:
i don't know, a lighter gray color?
Some minor issues:
- The category selection is buggy in console based browsers (tested with links2 and w3m)
- Long suggestion overflows the suggestion box

pointhi:
Due to refactoring I think we should also improve the oscar template. I'm not a designer, but I know some basic stuff. But color choosing, main graphic elements are not my strong side so I hope we can produce something greate together (I would implement the client side after doing the planing)

[...]

PR #174, opened 2015-01-15
add no javascript support to oscar-template

pointhi:
regarding to  #171, I have add nojs support to the oscar-template.

Now oscar should work without problems if js isn't activated. Implied, the browser is not too old.


asciimoo:
Awesome, thanks!

PR #175, opened 2015-01-15
Fix startpage

### LINKED ISSUE #169: Startpage engine doesn't support russian queries

Cqoicebordel:
Fix issue with unicode caracters in startpage : we shouldn't urlencode them if we are using POST.
Should fix #169. @dimqua can you confirm ?


asciimoo:
Nice fix, thanks, we need to check that is it affects more engines

Cqoicebordel:
I think it affected only startpage : it's the only one using POST as far as I can tell.

PR #176, opened 2015-01-16
Courgette theme - UIUX

Cqoicebordel:
Lots of tiny modifications in Courgette :
- Use courgette files instead of default's ones
- W3C fixes
- Whitespaces
- Less the CSS (including the Makefile modification)
- Add answers
- Add RSS in head to allow detection by browsers
- Unite UI (pretty url at the bottom, color of non result links, span instead of p...)


asciimoo:
Awesome, thanks!

PR #178, opened 2015-01-17
Thumbnails

Cqoicebordel:
- Modify engines to create/fetch an URL for the thumbnails
- Modify themes to show thumbnails instead of full images.

In Courgette, the result is not very beautiful. Should we change it ?


pointhi:
I would recommend checking if `template_src != null`, and showing `img_src` if necessary.

asciimoo:
Awesome! Thanks both the mods and the review!

PR #182, opened 2015-01-17
[enh] image-proxy : handle ETag and date related headers, add hash to URL

dalf:
- Handle ETag and date related header
- Add sha256 hash (URL and secret)
- Proxify image inside infoboxes


asciimoo:
Nice improvements, thanks - the template corrections too!

PR #183, opened 2015-01-18
improve oscar-template

pointhi:
regarding to #171:
- ~~allowing different bootstrap-themes inside oscar~~
- deactivate js-warning if possible


asciimoo:
could you please split it to two pull requests? Js-warning deactivation is great, but the theme topic is questionable, requires deeper review

Cqoicebordel:
I have an issue too with the theming : 
We need to be able to propose different skin/color scheme within a theme, without having to modify the app side of searx (webapp.py).
I was planning to do a color picker for courgette today, to change the theme, but I stumbled upon the issue of cookies.

We need a way to create and access cookies of local (theme) preference inside the theme. Maybe by extending a searx cookie class ?

asciimoo:
@pointhi thanks!

@Cqoicebordel yap, a sophisticated cookie handler could give bigger flexibility to themes

PR #188, opened 2015-01-19
General cookies

Cqoicebordel:
Allow a theme to have access to any cookie, and to store any cookie.
Modify Courgette to have a color picker.


Cqoicebordel:
I toyed with the idea of only allowing in the "cookies" var, the cookies whose names began with the theme name, for privacy issue.
But, in the end, it's all searx. It's searx writing the cookie, and reading it, so there is no privacy issue.

asciimoo:
What about handling the template assignment of the `request.cookies` inside the `render()` function instead of passing as a parameter?

Cqoicebordel:
Of course ! Much better !

asciimoo:
Great, thanks! =]

Issue #189, opened 2015-01-19
detect spelling error using provided suggestions


pointhi:
currently, we are recive suggestions from a few search-engines, but the strings are containing most time more as the actual search query (what suggestions should). But this suggestions have the correct spelling, even if the query is wrong. So I think we should use that suggestions, to check the query for spelling errors, and displaying a spelling-correction on the result-page if required.

For example, using the [Damerau–Levenshtein distance algorithm](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance) which is a fast spelling-correction algorithm as described in the faroo-blog (http://blog.faroo.com/2012/06/07/improved-edit-distance-based-spelling-correction).

Furthermore, a searx-instance can build an index which is containing all word from received suggestions, so the spelling-correction is also working for image-search,... as soon as the word is used with an engine with is providing suggestions

PR #190, opened 2015-01-20
add Spell checker

pointhi:
I have add a spell checker, like descibed int #189.

It is currently only using provided suggestions. I think we could improve them if we are also using the autocompleter-results. But I'm not sure how big the performance impact would be, if there are several thousands words in the database, so we have to test this first.

I'm sure we can improve the current code, to use same techiques like [farroo](http://blog.faroo.com/2012/06/07/improved-edit-distance-based-spelling-correction/) with has a runtime of O(1), using hashmaps. But I haven't figured out yet, how they have done that. So currently, we are only using the suggestions, which are including a very small subset of possible words.

Later we could use wordlists, generated from wikipedia or from other sources, to create a full offline spellchecker, and probably also autocompleter

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -58,4 +58,10 @@ if settings.get('server', {}).get('https_rewrite'):
     from searx.https_rewrite import load_https_rules
     load_https_rules(https_rewrite_path)
 
+# load https rules only if https rewrite is enabled
+if settings.get('server', {}).get('suggestion_wordlist', False):
+    # loade https rules
+    from searx.spellchecker import import_wordlist
+    import_wordlist(settings['server']['suggestion_wordlist'])
+
 logger.info('Initialisation done')
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -10,6 +10,7 @@ server:
     useragent_suffix : "" # suffix of searx_useragent, could contain informations like an email address to the administrator
     image_proxy : False # Proxying image results through searx
     default_locale : "" # Default interface locale - leave blank to detect from browser information or use codes from the 'locales' config section
+    spell_suggestion : True # using suggestions to check spell of search-query
 
 engines:
   - name : wikipedia
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -11,6 +11,8 @@ server:
     image_proxy : False # Proxying image results through searx
     default_locale : "" # Default interface locale - leave blank to detect from browser information or use codes from the 'locales' config section
     spell_suggestion : True # using suggestions to check spell of search-query
+    suggestion_distance : 2 # maximum levenshtein distance for spell-checking
+    suggestion_wordlist : False # IN DEVELOPMENT: using wordlist for spelling correction, ATTENTION: For example: '/usr/share/dict/words' require more than 1GB RAM on every thread if used!
 
 engines:
   - name : wikipedia
--- a/searx/spellchecker.py
+++ b/searx/spellchecker.py
@@ -15,9 +15,20 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 (C) 2015- by Thomas Pointhuber
 '''
 
+from searx import settings
+from searx import logger
+
 from sets import Set
+from os.path import isfile
 import operator
 import re
+
+
+logger = logger.getChild("spellchecker")
+
+# wordlist, which is containing all variants of possible words incl. correct words
+# TODO: using database
+wordlist = {}
 
 
 # Source:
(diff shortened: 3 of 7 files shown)


Cqoicebordel:
Nice work !

I would suggest three tiny things : 
- Add an option to allow to choose the levenshtein distance from settings.yml. Something like spell_suggestion is false or is the value of distance should be fine.
- Like said in the blog post, you should break the calcul of levenshtein distance when the max distance threshold is reached. Could means better perfs.
- As a static list of word, you could use the list that most linux have, `/usr/share/dict/words`

pointhi:
ok, I have add the possiblility to change levenshtein distance, and add a first implementation to use a wordlist.

I have implemented the wordlist like described by faroo, so it is really fast. The problem is the memory consumption, because  we have no database yet (around 1GB for `/usr/share/dict/words`(around 3 Million words which have at least 1. correct spelling correction) inside every thread).

I add a warning to this option becausse of that. (My computer was going frozen when I was loading the german wordlist, because of the big RAM consumption)

a dynamic wordlist would be more pratical for most use (using auto-suggestion results to fill it with new words), but does also require a database to work with multible threads correctly.

pointhi:
except spell suggestion we can also use it to detect the language of an result, if we are storing the language of every word. So we could implement #206 without knowing the language of the specified url.

For that we could load for example the top 500 or 1000 words of a desired language into the index, and then process language-detection from it. I would run tests how much word-language-checks/spelling-checks the current implementation can handle per second.

Cqoicebordel:
There is a much better way to detect language. You need to store the statistical prevalence of trigrammes of letters per languages (it works also with less letters than 3, but it's less exact). This way, you could have a score, that is independant of words and new requests.  
Because, for example in Google, if my memory is correct, they have daily around 75% of their requests they never saw before. So with content new, and never saw before also.  
So the word approach is less than perfect, because we have only a tiny content to detect those words, the words are fairly new, but the statisticall approach allow new words.

I can give more details, I already succefully coded this kind of detection a long time ago.

dalf:
Some links : 
- Language detection : https://code.google.com/p/language-detection/ ( slides : http://fr.slideshare.net/shuyo/language-detection-library-for-java )
- Chronium Language Detector 2 : https://code.google.com/p/cld2/ (with Python bindings : https://code.google.com/p/chromium-compact-language-detector/ )
- http://fr.slideshare.net/shuyo/short-text-language-detection-with-infinitygram-12949447

Cqoicebordel:
:+1: Exactly that !

pointhi:
trigrammes are looking good. (I wasn't knowing about that, textprocessing is new for me, but very interesting :))

I found a library which could do the required work: https://pypi.python.org/pypi/langdetect/1.0.1. On my computer it processed between 50-100 senteces/second. This isn't fast enought if we want process every result.

I have also tested my current implementation of the spellchecker. I can process about 2000 Words/Second with the current implementation. There is much optimisation potential to make it faster for cases like correct words or low edit distance (currently every possiblity is calculated even if we could abort the calculation). I don't know if an external database like MySQL with hash index is improving or decreasing the lookup-speed. But without an database it wouldn't be practical because of the RAM-requirements.

I'm not sure, but if I'm understanding trigrammes correctly, we could using hasmaps like inside the spellchecker to speedup detection(if not done yet)

Cqoicebordel:
Yes, with trigrammes we could use hashmaps and be around O(1), because there is no `n`.  
But I'm not sure using a DB with everything in it would be a good idea, even with an hash index, because the lookup time is not in O(1) but depends of `n` and would likely be O(log(n)).

I was recommending using the system list of words, because mine is relatively small (~100K lines). 
But you are right, having the words+the possible errors loaded or saved in base would be too much. We can't possibly ask a searx admin to have a few GB to spare for us (disk or RAM). Per languages.
Or maybe with an option. But even...

I have no idea how to do this properly anymore.

For the trigrammes, if you are interested in the basis, there is that science paper I used. It's not easy to understand (like any scientific paper...), but the math formula is the basis : 
http://swiss.knife.free.fr/StatisticalLID-II.PDF
And if you need help in text processing, don't hesitate to ask, I have done a lot of those (but not in python ;) )

pointhi:
I think It's clear that enhanced textprocessing operations are optional for an admin (and by default deactivated), because of the big cpu/storage requirements.

But If we implement text processing, it should be clear that speed is an important part, and then we have the known memory vs cpu problem. But I think we could implement greate features with text-processing:
- In my opinion, the basic part/algorithm of the spell-suggestion is the best combination of speed to memory we could easily achieve. (There are some improvements possible, but working only with deletions and hashmaps is a good/basic-algorithm).
- Language detection with trigrammes are interresting, but to slow to parse every result. A combination of a domainname-parser (.de domains are mostly german,...) with fast trigrammes for unknow cases can be a good combination. Probably we can also use language-detection for the search-query, to select a search-language (Automatic Language Modus)
- Question Answering is also an interesting part. But I haven't figure out yet, how I could implement a fast/good parser for a specific subset of questions (Geodata, Math, Weather).
- Autocompletion is a part where I don't know if we could achieve them without an very big database. The easiest idea would be using the spell-checker. But then, the autocompletion would be really random :). Probably using a Radix tree, build from older (autocompletion-) results.

@Cqoicebordel thats greate that you have experience with text processing.

Cqoicebordel:
For the spell checker, I had a thought a few seconds ago : Hunspell.  
I don't know how it works, but it's really fast, and the dictionnaries files are very light. And it's Open Source. 
It should be speedy enough for suggestions, and even auto complete.

It could be a good way to go.

PR #192, opened 2015-01-21
[enh] improve response time. close #100

### LINKED ISSUE #100: Improving response time : dns prefetch & tcp ssl pre-open
For each request, there is different steps :
- resolve dns
- open tcp connection
- negociate ssl
- send request
- receive response
- close connection

The first open can be optimize by prefetching dns. I guess it's a matter of 50 ms at most so... But it can be done easily by sending dns request with a timer (each hour for example).

The second and third one cost 100ms or more per request. If searx could start opening TCP connection when the user is one the search page, it could help. To implements this, a patch must done on grequests.

Of course, as chromium, this implies some useless TCP connections ( https://code.google.com/p/chromium/issues/detail?id=116982 )

dalf:
Package searx.poolrequests can replace the requests package.

It's the same API and implementation except that it's use two global HTTP connection pool (one for HTTP, one for HTTPS). In the standard implementation there is one connection pool for each request.

See issue #100 for the explaination. In few words : most of the time, the socket is already opened, and SSL negociation done (except if the server close the connection).

On my PC, the response time is decreased by 200ms most of the time (except for the first request).

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -17,13 +17,13 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 
 
 from lxml import etree
-from requests import get
 from json import loads
 from urllib import urlencode
 from searx.languages import language_codes
 from searx.engines import (
     categories, engines, engine_shortcuts
 )
+from searx.poolrequests import get
 
 
 def searx_bang(full_query):
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -1,6 +1,6 @@
 import json
-from requests import get
 from urllib import urlencode
+from searx.poolrequests import get
 from searx.utils import format_date_by_locale
 
 result_count = 1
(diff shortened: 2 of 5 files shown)


asciimoo:
Oh wow! Impressive speed improvement!
As far as i know, requests(lib) session object also persists cookies through the (HTTP) requests, which would be good to avoid. @dalf could you solve this issue?

dalf:
As on the requests lib implementation, a new Session is created for each request but the two HTTPAdapter remain the same : there is no cookie persistence.

The code I posted yesterday was reusing Session and clearing cookie, but I think it's too much complicated.

Several notes : 
- the 200ms is on my internet connection, may be on a server / datacenter it's different. 
- closed connection are detected by urllib3, but is there some exception, some special case ?

asciimoo:
You're right about the cookies, thanks!
Connection closing shouldn't be a problem, searx handles engine/request exceptions.

PR #195, opened 2015-01-22
[fix] google engine (close #193)

### LINKED ISSUE #193: Google engine : latency caused by redirection
The current implementation of the google engine has a little issue : there is a redirection for each request, but not for everyone.
It's depend of the geoip : https://support.google.com/websearch/answer/873?hl=en

If you look at the response time of the different searx instance, some time the google time is slow compare to the other engine, I guess this is this issue.

One way to avoid this is to set the PREF cookie (set but http://www.google.com/ncr ). Of course there is an unique ID. On solution is to initialize the cookie at start of the searx instance : 

``` python
pref_cookie = ''

def get_google_pref_cookie():
    global pref_cookie
    if pref_cookie == '':
        resp = get('https://www.google.com/ncr', allow_redirects=False)
        pref_cookie = resp.cookies["PREF"]
    return pref_cookie
```

and set the cookie in the request method : 

``` python
   params['cookies']['PREF'] = get_google_pref_cookie()
```

The cookie is global to the searx instance, so in my opinion there is no privacy issue.

Of course, if there searx instance is behind a router and share a public IP with other machine, the request are more "tagged", but isn't already the case since there is not all the normal header that a browser send ?

Another solution is to revert to the previous solution that api.google.com, but the results are different from the "browser" version (see #92)

dalf:
depending on the IP of the searx instance, each searx request where making two HTTP requests (see https://support.google.com/websearch/answer/873?hl=en )

see #193

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -11,6 +11,7 @@
 from urllib import urlencode
 from urlparse import urlparse, parse_qsl
 from lxml import html
+from searx.poolrequests import get
 from searx.engines.xpath import extract_text, extract_url
 
 # engine dependent config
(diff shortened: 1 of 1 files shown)


asciimoo:
Great, thanks!

PR #200, opened 2015-01-22
Change the cookie in bing_news to use the english interface

### LINKED ISSUE #198: bing_news can't parse other languages date
When searching for french article, the time is noted as "Il y a 5 minutes", and so, doesn't match for the regex `"^[0-9]+ minute(s|) ago$"`.

Do you see a way to internationalize this detection ?

Cqoicebordel:
But still uses the language to set the market, and so provide relevant results to the language.
Fix #198

Thanks @pointhi !


asciimoo:
Thanks!

PR #202, opened 2015-01-26
Unit tests

Cqoicebordel:
First pass at creating unit tests.

Of course it's a work in progress, but you can merge those if you want.

I have two issues : 500px and flickr-noapi. How can I import/use them ? Python doesn't seem to accept a number as first digit of an import, and a package with an hyphen isn't taken well too.

There is a bug in coverage : it tells that flickr is missing a statement, but that's not true. And if I add a print before the statement (a "continue", L66 of flickr engine), it sees it. So, if you have an anwser to that too, it would be great.


asciimoo:
Uh-oh, very-very cool! Also the tests and the engine tweaks!
Quick and dirty solution to the import is doing the same way as searx,  but we should rename these engines to fit to the python standards.
At first, I don't have any idea what can cause the coverage error.

Cqoicebordel:
I let you do the smart thing. There is lot to do before it's blocking ;)

Cqoicebordel:
Since I cherry picked the changes in 500px and flickr-noapi, Travis won't run.
You could merge the already commited changes, so the next ones would be verified by Travis.
And there are a lot of tiny fixes in the engines that could already benefit searx ;)

BTW, since the pirate bay came back, I reactivated it in the settings.

asciimoo:
Supercool!! Thank you sir!

PR #208, opened 2015-02-01
add 1x.com engine, improve yacy-engine

pointhi:
I have add a engine for www.1x.com, and improve the yacy-engine to work basically with !bangs.

1x.com is deactivated by default, because of the big amount of pictures provided by a request (all found images)

I think yacy requires much more refactoring as I have done yet, to enabling all possible features which are possible.

### DIFF
--- a/searx/engines/www1x.py
+++ b/searx/engines/www1x.py
@@ -36,7 +36,8 @@ def response(resp):
     results = []
 
     # get links from result-text
-    results_parts = re.split(r'(</a>|<a)', resp.text)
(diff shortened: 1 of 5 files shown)


Cqoicebordel:
Since you know how 1x works, it should be easier for you to do the unit test for it.

If not, I'll have to study the details of the engine and the 1x.com page before being able to do it...

asciimoo:
Great, thanks!

Cqoicebordel:
Awesome :)

PR #210, opened 2015-02-02
unit tests

Cqoicebordel:
Let's continue...

### DIFF
--- a/searx/engines/currency_convert.py
+++ b/searx/engines/currency_convert.py
@@ -13,12 +13,9 @@ def request(query, params):
     if not m:
         # wrong query
         return params
-    try:
-        ammount, from_currency, to_currency = m.groups()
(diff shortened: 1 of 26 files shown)


asciimoo:
=))
<3
Seems, the currency convert try/except block isn't necessary.

Cqoicebordel:
I thought so, but wasn't sure. The other float() had also a try/except so I thought it was mandatory.
Anyway, I will delete it tomorrow. Thanks !

Cqoicebordel:
I think I'm done here for now.

What is left is either engines that don't work anymore or is out of my depth, I think. Maybe I will try to tackle them, but not right now. If anyone want to do it, you are welcome to.

So, @asciimoo you can merge :)

asciimoo:
very-very cool, thanks again!

PR #212, opened 2015-02-03
Add Docker support.

glogiotatidis:
This is a simple dockerfile that creates a docker image with the current checked out files, installs packages needed and makes basic configuration to run searx. By default it runs a uwsgi server on port 5000.

I've used this successfully on [deis](deis.io/) and [dokku-alt](https://github.com/dokku-alt/dokku-alt) servers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM debian:stable
+
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


asciimoo:
cool, thanks!

PR #213, opened 2015-02-03
Stop running twice the tests on Travis

Cqoicebordel:
Tests are run twice. Once on make tests, and second in make coverage.
If we don't run make tests, we still need the make flake8. Hence the change.


asciimoo:
True, thanks!

PR #215, opened 2015-02-06
[enh] google engine:

dalf:
First point : For example google.hu gives different results from google.com even if the Accept-Language header is set to hu.

This pull request uses the search language setting to the defined the google site (more precisely the country of the search language).

It can be disabled by setting use_locale_domain to False in settings.yml 

Second point : the review extracted by google ("notes : 4.2/5" bellow the title, above the text) are added to the results.


dalf:
I close the pull request in the current state : it's work on my ADSL connection, but not on a server.
The requests are redirected to the captcha test.
I guess that google use the IP address.

PR #216, opened 2015-02-07
RTL handling in searx

Cqoicebordel:
First version, only in Oscar.
What is done :
- Taken care of every RTL script possible in webapp
- Page order
- Categories
- Footer of results
- Pagination
- Preferences

What bugs there are/what is still possible to do :
- Alignement of suggestions
- Alignement of cvs, json, xml
- Engines in prefs ( `google (go)` appears as `( google (go`. I think it will be solved by using a real rtl script but still)
- Alignement of images results
- Search field in index not the right size
- Others themes...

@pointhi you know bootstrap largely better than I, can you look at the styles needed for the buggy ones ?  
I know it should be easy for some. Like in preferences, I used nav-justified instead of putting the tabs to the right. Or the alignements stated above.

@GreenLunar if you know how to test it, please do. It's hard to tell if RTL is correct using latin script, and my hebrew is a bit rusty...


Cqoicebordel:
Yes I know. But I don't know how to handle it, given the fact that the changes are globally around order of elements. 
And as stated, it's a first pass... ;)

GreenLunar:
Start with little code. I will report on problems over time.

> Alignement of suggestions

Maybe.

> Alignement of cvs, json, xml

No.

> Engines in prefs `( google (go)` appears as `( google (go`. I think it will be solved by using a real rtl script but still)

Use `&lrm;` -> `(go)&lrm;`
http://www.fileformat.info/info/unicode/char/200e/index.htm

> Alignement of images results

Maybe.

> Search field in index not the right size

Where is that?

> @GreenLunar if you know how to test it, please do. It's hard to tell if RTL is correct using latin script, and my hebrew is a bit rusty...

I need changes to be applied in a public searx node in order for me to test this.

Cqoicebordel:
For the alignements, it's mainly to align the boxes to the right, not the text itself. The same for the images.

The search field is in the index page, the page you have when you land for the first time in searx.

Thanks for the `&lrm;`, it was that :)

I'm sorry, but I can't provide a public searx node. So we'll have to wait until the code is merged, and we will see from there.

My aim is to provide as soon as possible an RTL enabled interface here. Perfection will come later ;)

GreenLunar:
> For the alignements, it's mainly to align the boxes to the right, not the text itself. The same for the images.

Precisely.

> Thanks for the `&lrm;`, it was that :)

You are welcome.

> I'm sorry, but I can't provide a public searx node. So we'll have to wait until the code is merged, and we will see from there.

No problems.

Cqoicebordel:
I think I'm done here.

@asciimoo if you are happy with the code, you can merge.

@GreenLunar for further comments or changes, tell us in the issue you opened.
Once this code is merged and your translation in hebrew too, you will be able to test it at searx.me.

GreenLunar:
This lesser than two days period, in which I am waiting to test searx in RTL, feels like a week, to me.

asciimoo:
@Cqoicebordel thanks! The code still has redundant parts, but it's a good start.

Cqoicebordel:
Yes, thanks :)

Your approach is much better than mine. I tried to fit everything into a macro, but there were too many variables, so I went to the simplest form. Your way is the good way.

But it still weird to have to delete this branch without merges ;)

GreenLunar:
> But it still weird to have to delete this branch without merges ;)

Even weirder, implementing RTL support before having RTL locale added :P

Cqoicebordel:
It's your work now ;)
Go to transifex, there is an update of the traduction to do ;)

GreenLunar:
> Alignement of suggestions

Yes, set it to RTL. EDIT: **oscar theme**

GreenLunar:
> Alignement of suggestions

Yes, set it to RTL. EDIT: **oscar theme**

GreenLunar:
I forgot to mention: Great job!

PR #217, opened 2015-02-07
add Big map

pointhi:
I have improved map-handling, add a satellite layer and added a first version of a big map.

On the big map you can move like in the small maps, and search inside the map topic. I know that the current style is really messy, but I have not the experience to make it better.

Before I add new features to the map, I would look if I can find someone in the OSM-community who would create a better map :). I think there are some OSM-Programmer which would like a good open source metasearch-engine with good OSM-implementation. My frontend skills are much to low to create something really good.

My Ideas would be:
- much better style, including mobile support
- Better search (including autocompletion from photon)
- routing with OSRM
- geolayer? (georeferenced wikipedia photos)
- Overpass API implementation? (like the wizzard inside Overpass Turbo. First we need a Question Answering System before we can implement that in a proper way)


Cqoicebordel:
Really really nice !

I would like to add to your ideas list, to be able to go to the big map from a standard map result.

Also, you could ask for the position of the user to the browser to center the map (could be very helpful on a smartphone)

Beyond that, it's really really great already !

searx/webapp.py
```diff
@@ -401,6 +402,18 @@ def about():
     )
 
 
+@app.route('/map', methods=['GET', 'POST'])
+def map():
+    """Render about page"""
```
asciimoo, line 407: @pointhi could you fix this docstring?

asciimoo:
also interesting: https://blog.openstreetmap.org/2015/02/16/routing-on-openstreetmap-org/

pointhi:
A question: has anyone a problem if i'm rewriting the js code inside oscar to typescript? Advantage would be clearer code and include possibility, disatvantage: It has to be compiled, and it was created by microsoft :)(is opensource)

dalf:
I don't know Typescript, but why not ECMAScript 6 ? Since it will be the next version of (Java|ECMA)script.

The feature list : https://github.com/lukehoban/es6features#readme

An article about the classes : http://www.2ality.com/2015/02/es6-classes-final.html

And there is a way to compile ECMAScript 6 to version 5 which is supported by the current browsers : https://github.com/babel/babel

Cqoicebordel:
The big obstacle for this rewrite would be for me to find someone who know typescript. I don't personnally. And I don't see it as a future language, so I'm not seeing myself investing in it. But I could be wrong.  
But in the meantime, if you are the only one who can understand and modify the code I think it would be dangerous.

Another argument is that we have a lot of languages already. Another one would increase the barrier to participate.

Also, Oscar is already using Bootstrap and JQuery.

Finally, I'm a big defensor of Vanilla JS http://vanilla-js.com/

Anyway, I'm not opposing it. I just don't see the advantages and what is blocking you right now with regular JS. (Also, if TS when compiled give regular JS, why not write directly the regular JS part ?)

pointhi:
Typescript has many similarities to EmacScript6. Code style is similar to java, and readiblity is much higher as for vanilla js. Furthermore it has a grate typecheck sxstem.

I would only rewrite searx-specific code, we can still use parts written in vanilla js.

For me, I need a client language in which I can use OOP and includes in an easy way. Without function inside function inside prototype,...

dalf:
I say "ECMAScript 6" because it will be supported natively one day or another, and at this point more and more people will know it.

But yes type checking is not supported except with this kind of hack :  http://www.nczonline.net/blog/2014/04/29/creating-type-safe-properties-with-ecmascript-6-proxies/ (and I don't know how this transpile to ECMAScript 5)

BTW, I just found that there is a bunch of transpiler : http://kangax.github.io/compat-table/es6/

@asciimoo what do you think ?

pointhi:
Babel looks good to me, but there switching to ECMAScript 6 only makes sense if there is something like include support like in typescript. otherwise we couldn't handle bigger code-parts on client side.

Currently, we copy together all js-files and minimize them. But writing a good map with routing-capability,... requires much more code, and also creating dependencies between the files.

Here is a little example of which i have programmed a little time ago in typescript:

``` typescript
/// <reference path="../impl/iPoint.ts"/>

module engine {
    export module baseType {

        export class Point implements impl.iPoint {

            constructor (private x: number
                        ,private y: number) { }

            // get x-axis of Point
            getX() {
                return this.x;
            }

            // get y-axis of Point
            getY() {
                return this.y;
            }
        }
    }
}
```

and that is the created EMACScript5 to provide that functions:

``` javascript
/// <reference path="../impl/iPoint.ts"/>
var engine;
(function (engine) {
    (function (baseType) {
        var Point = (function () {
            function Point(x, y) {
                this.x = x;
                this.y = y;
            }
            // get x-axis of Point
            Point.prototype.getX = function () {
                return this.x;
            };

            // get y-axis of Point
            Point.prototype.getY = function () {
                return this.y;
            };
            return Point;
        })();
        baseType.Point = Point;
    })(engine.baseType || (engine.baseType = {}));
    var baseType = engine.baseType;
})(engine || (engine = {}));
```

TypeScript is very similar to EMACScript6, but as long as no one can provide a similar or better way to solve dependencies, typescript would be the better choise. In typscript you can use `/// <reference path="../impl/iPoint.ts"/>` to solve dependencies, and I haven't found anything similar for vanilla-js.

Cqoicebordel:
What is bothering me is that it seems to me you want to recreate the wheel with a technologie that is very very young, and client side.

Why not do the routing system in python, with ajax calls ?  
Routing can be CPU intensive, especially if you are using A*. And JS is not the language for it. At all.

pointhi:
I think you dosn't understand me right. I don't want to create a routing engine on js-basis. There are great projects out there like [OSRM](http://map.project-osrm.org/) which I want to use, but there is much more client-side programming required to get the routing-parameters, display it on map, display extra information,.... In addition there are some ideas which also require client-side programming, so the js-part in oscar would growing including more dependencies between the files.

I would not write that stuff in one file, because no one could maintenance it properly. Furthermore, Typescript or EMACScript6 are much more pleasant to programm as vanilla-js, especially for programmers which are not web-developers (like me).

There is the javascript-part which is used on OpenStreetMap, to get a feeling how big the client-side codbase could grow for a map: https://github.com/openstreetmap/openstreetmap-website/tree/master/app/assets/javascripts

pointhi:
Ok, I have started reimplementing the client-side script part in typescript. Mainly, because of the dependency-solver, typecheck and much better readable syntax.

The code is developed in a new branche:
[pointhi:big_map_typescript](https://github.com/pointhi/searx/tree/big_map_typescript)

I have not only copied old code, but also improved it to be more flexibel, and also added post-request-support to the autocompleter. Map isn't supported yet, because I have to write a nice object, which can handle the little map as well as the big map with routing,... in a clear way.

PR #220, opened 2015-02-08
Safesearch

pointhi:
I have add the safesearch-functionality, which is implemented in some search-engines. It is based on the pull-request #219, because blekko also support safesearch.


asciimoo:
Ah, good idea! Thanks!

PR #233, opened 2015-02-12
[fix] kickass engine

### LINKED ISSUE #197: Requests does not follow on an HTTP 301
We had already an issue when kickass redirected from https to http I seem to remember. Or it was on another engine. And the request wasn't followed so we had an empty response with status_code 301 in the engine.

Anyway, that is still the case in kickass if we try to perform a search when the query contains capitalized letters.

So why isn't 301 followed ? Is there a way to force it ?

dalf:
Change the host name to kickass.to (since kickass.so doesn't respond).

Close #197 perhaps not in clean way. 
Explanation :
In fact 301 responses are followed, except the hook is called for each HTTP response,
the first time for the HTTP 301 response then for HTTP 200 response.
Since the kickass engine excepts a real result, the engine crashes, AND the requests lib stops here.
Add a simple test at the beginning of the result function allows pass the first response and handle correctly the second response (the real one)

May be a proper way is to add this test in search.py ?

Code inside requests :
https://github.com/kennethreitz/requests/blob/53d02381e22436b6d0757eb305eb1a960f82d361/requests/sessions.py#L579
and line 591


asciimoo:
@dalf, it's a good idea, we should create a response callback wrapper which checks if it is a redirect, i'll play with it.
And thanks the fix, could you update the tests too?

dalf:
Done. Sorry for the messy commits.

asciimoo:
Great, thanks!

Cqoicebordel:
@dalf great catch !

@ascimoo indeed, a wrapper would be a good idea, because we had the same issue in another engine (but I don't remember wich. Vimeo ?)

PR #235, opened 2015-02-13
Some RTL fixes

Cqoicebordel:
- Pull the suggestions to the right in oscar
- Add `&lrm;` to a few places to print good url
- Always LTR code display
- Order pref label and info in pref for mobile view

And add `sort` for ui languages and languages to find them more easily.


asciimoo:
Thanks!

PR #237, opened 2015-02-13
[fix] close #197 for all engines

### LINKED ISSUE #197: Requests does not follow on an HTTP 301

dalf:
same as commit 9db0bbc30421cec9d42438ccab748439eea60105
but for all engines

### DIFF
--- a/searx/engines/kickass.py
+++ b/searx/engines/kickass.py
@@ -45,11 +45,6 @@ def request(query, params):
 def response(resp):
     results = []
 
-    # check if redirect comparing to the True value,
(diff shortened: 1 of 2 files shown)


asciimoo:
Clever, thanks!

PR #240, opened 2015-02-13
Experimental features : async parsing of results, thread pool

dalf:
## DO NOT MERGE

It's experimental, it's only to discuss if needed or to close and forget.

In the current implementation, the main thread does absolutely while the requests are resolved.

What is done : 
-  as soon as an first engine responds, the results are added, deduplicated and more important the https rewrite could be done here : even if the https rewrite is slow, there is an impact on the response time only on the last engine to respond.
- a thread pool, which avoid the cost of creating numerous thread for each searx request. Draw back : with uwsgi or whatever, there is a lot of thread.

Note : 
- There are bugs (I think there is an issue with the score for each result).
- There is new class Search_Result to consolidate the normal result. An similar class could be done for the info boxes (may be it's overkill)

### DIFF
--- a/searx/async_call.py
+++ b/searx/async_call.py
@@ -3,7 +3,7 @@ import Queue
 import time
 from searx import logger
 
-logger = logger.getChild('search')
+logger = logger.getChild('async_call')
 
 
 class AsyncCall(object):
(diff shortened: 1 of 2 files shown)


dalf:
About async requests : 
*http://tavendo.com/blog/post/going-asynchronous-from-flask-to-twisted-klein/ 

This is exactly the "problem" of searx

About uwsgi : 
- http://uwsgi-docs.readthedocs.org/en/latest/Async.html#running-uwsgi-in-async-mode

asciimoo:
Very cool ideas! Tests would be good to reveal bugs and measure response time/performance differences

pointhi:
If https://github.com/asciimoo/searx/commit/b6c3cb0bdd020a459d0ef5c21d1303ed0148cc0c has implemented all of this pull request it would be a good idea to close it.

asciimoo:
+1, dalf?

dalf:
+1

dalf:
Note : one thing that is missing is the current implement : reuse of threads (to avoid the cost of creating thread at each request which is time consuming). 
But the way it is implement in the PR is way too complicated.

PR #241, opened 2015-02-14
Remove whitespace and allow autocompleter to work in sub folders

Cqoicebordel:
Adding this conf change will remove every newline created by a Jinja command (`{% %}`). It can save a bit of bandwidth.


asciimoo:
Thanks - closes #244

PR #246, opened 2015-02-17
New theme !

Cqoicebordel:
Here is a new and beautiful theme !

Ok. It's not that beautiful, and to be honest, it's not that useful either. That wasn't the point of this theme.

The point was : because I wanted to :)

So this is a theme in pixel "art". Without any text result. Yes, no text at all.  
It displays the results in favicons only, and pixelated, at that.

The real point was to have fun, and show that we could do something differently :
- Different UI
- No categories
- No image proxy
- No completion
- Ajax loading of the next page results
- Display of results in canvas
- No bullshit

Anyway, it works best with Firefox, Opera 12, probably IE (can't test it) and maybe Safari (I have no idea), and will work with the next version of Blink (Chrome/Opera 15+).  
On the todo part, I want to integrate the images results in the same way (pixelated), but can't right now (Shapes in CSS are not implemented in enough browsers).


Cqoicebordel:
You don't want to merge this ?

I'm just curious about why, and why the other request isn't merged either...

asciimoo:
it's on my list, i have to test it

Cqoicebordel:
I hope you didn't take my question as pressure on you :)

Cqoicebordel:
Just so you know, it works now in Chrom(e|ium) (in 41+).

And this pull request is almost two months old. Just sayin'...

asciimoo:
Sorry for the big delay,
thanks!

PR #249, opened 2015-02-20
[fix] update yahoo engine according to the web site changes

dalf:
https://search.yahoo.com/ has changed. This pull request fixes the XPaths according to those changes. 

Note : https://*.search.yahoo.com/ use the old version (but those URL are not used by searx)


asciimoo:
thanks!

Issue #250, opened 2015-02-20
general-file.com


dalf:
According to [google safebrowsing](http://www.google.com.sg/safebrowsing/diagnostic?site=general-file.com/), general-file.com is a scam.

PR #251, opened 2015-02-20
Moar engines

Cqoicebordel:
Add Spotify, remove dead engines, and correct some TU to obtain 100%.

I think it does close the #250.


asciimoo:
\o/ thanks!

Issue #253, opened 2015-02-22
AttributeError: 'Response' object has no attribute 'is_redirect'


pointhi:
For a few repositories (my searx_oe5tpo and the metorology branche from @dalf), I get everytime the following error:

```
ERROR:searx.search:duckduckgo : engine exception : 'Response' object has no attribute 'is_redirect'
Traceback (most recent call last):
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/searx/search.py", line 67, in search_request_wrapper
    return fn(url, **kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/searx/poolrequests.py", line 35, in get
    return request('get', url, **kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/searx/poolrequests.py", line 28, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/eggs/requests-2.2.0-py2.7.egg/requests/sessions.py", line 383, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/eggs/requests-2.2.0-py2.7.egg/requests/sessions.py", line 491, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/eggs/requests-2.2.0-py2.7.egg/requests/hooks.py", line 41, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/home/thomas/Dokumente/Projekte/searx/better_stats/searx/searx/search.py", line 127, in process_callback
    if response.is_redirect is True:
AttributeError: 'Response' object has no attribute 'is_redirect'
WARNING:searx.search:startpage : engine timeout (waiting for threads)
```

dalf:
I can't reproduce the bug.

Does it solve with :

``` python
if hasattr(response, 'is_redirect') and response.is_redirect is True: 
```

?

pointhi:
yes, that line solve this issue.

asciimoo:
Same as @dalf, i cannot reproduce it.
@pointhi what is the version of the `requests` lib?

pointhi:
I think 2.2.0

edit: and there is also grequests 0.2.0

asciimoo:
hmm.. strange..

(note: grequests is not part of searx anymore)

[...]

PR #98, opened 2014-09-14
update versions.cfg to use current up-to-date packages

pointhi:
I have updated the versions.cfg, using the informations from https://requires.io/github/pointhi/searx/requirements/?branch=master

PR #257, opened 2015-02-25
[fix] #253

pointhi:
I found out why the bug was happen. version.cfg contain the version `2.2.0`, but `is_redirect`, was added `2.3.0`.

regarding to #98, we have to update our versions.cfg!!! Some packages are currently insecure. inside the versions.cfg. I don't know that much about python, but it would be the best if we could drop that file in some way, and fixing that issues soon. (or updating that file regulary)

https://requires.io/github/pointhi/searx/requirements/?branch=master


asciimoo:
Thanks! You're right, we have to do that. Searx works with the current latest dependencies imho.

Issue #60, opened 2014-03-27
adding Widget support


pointhi:
Currently I'm work to get and display weather-data like google. I have code it as engine with custom template, but I think, it would an nice idea, to add support for widgets.

As widget, I would define special functions, like displaying weather data, a calculator,... if required. So, It parse the querry, and when required it display special widgets on the result list (on the top).

pointhi:
@asciimoo it would be greate if offline engines support is working soon.

My implementation ideas:
- probably we can create an option `offline=true` and if this option is chosen only one function is calld
- or we define `url=None` and calling dectecting offline-functionality in this way

Furthermore, I think we should think about an master template for widegets. It would't be handable if we create a new .html-file for every wideget which is showing things a little bit different. Nothing which can handle all cases, but most.

asciimoo:
@pointhi, agree, the only thing that holds this issue is an other approach i'm thinking on. Kind of a plugin system, where plugins can attach themselves to specific events like search start/end and manipulate the actual state of the performed search (eg. reorder/add/filter results, rewrite engine/search settings, ..).

nogweii:
It'd be great to be able to integrate DDG's open source instant answers with searx more directly. Unfortunately, it's unlikely to directly embed them in the source code or server process -- the maintenance burden would be too high. I think something that'd be easier to maintain would be a secondary process written in perl with a lightweight API (HTTP on loopback, or something else) exposing the "goodies". (Unfortunately, not every instant answer is available on Github, some of them remain closed source due to licensing agreements with whomever.)

PR #270, opened 2015-03-08
Fix IP instant answer

Lamessen:
When you do a research on "ip", duckduckgo send an instant answer with the IP adress of the searx server, which is not useful.
This fix gives to user the ip adress that he uses to access searx.


pointhi:
I think we should add Widget/Plugin support, to support such features #60.

ldidry:
I don't think so (or at least, not now):
- as I can see on your issue, plugins are not ready
- right now, searx returns a false answer, which is _really_ bad for the user and searx reputation

Cqoicebordel:
+1 to both.

We should work on widget/plugin as soon as possible.

dalf:
+1 for widget / plugin system

EDIT : In my opinion, the fix should be in duckduckgo_definitions.py not in webapp.py

If I activate the zealot mode about privacy : searx shouldn't answer THIS question because, it's mean that the browser IP value will be spread over all the code.

If we talk about privacy we should even advice this in nginx configuration : 

```
uwsgi_param REMOTE_ADDR 127.0.0.1;
```

ldidry:
Don't you think it would be a good thing to merge this commit and revert it when the plugin system will be ready?

dalf:
All these queries give the IP address : 
- ip
- what is my ip
- what is my ip address

I suggest that
- duckduckgo_definitions.py detect the answser "Your IP address is ..."
- and either ignore this answer, or replace it with the right one.

ldidry:
There is now a plugin for that, introduced in commit 00cc4dcbf44d9ecea89befb08cae4ee5561c4247  
I think this PR can be closed, no?

Lamessen:
plugin now exists for that

PR #274, opened 2015-03-15
Fixes #271

### LINKED ISSUE #271: Use the method setting for opensearch.xml
In the preferences, there is a setting for setting the method (POST or GET) you want to use, but when you add searx to your web browser's search engines, you can only search with POST (except Chrome that can only handle GET method).

It seems logical to use the method the user choose for the web interface, don't you think? I could do a PR if you tell me it's OK.

ldidry:
The opensearch method is now the method set in the preferences.

As before, POST by default and GET for Chrome/Chromium which doesn't
handle POST


asciimoo:
Thanks!

PR #276, opened 2015-03-17
New theme: frama

ldidry:
This is a new theme based on Twitter bootstrap, which was first written from scratch. Then, the oscar theme has been released a few days before our PR.

The two themes were quite close to the eye (both bootstrapped) but the oscar theme was waaaay better at results handling (like the maps, the music, etc). We wanted to rewrite our theme but starting from oscar before make our PR. This is the result.

This theme changes a behavior of searx : you can't search through more than one category on the same time. Changing category will relaunch the search with the new category. This has been asked by users which were kind of lost with searx behavior. They were saying "I change category, but nothing happens. What's wrong? Is the search engine broken?".


asciimoo:
The theme is very similar to oscar and since category selection implemented as a plugin this theme doesn't provide significant extra functionality but bunch of duplicated code.
@ devs what do you think?

dalf:
Theme inheritance ? Say somewhere that frama theme is based on oscar, and frama theme could use the oscar resources.

ldidry:
Yes, some resources could be used directly from oscar theme but for now, I prefer to manually make the diff to prevent incompatibilities. Once I'll have all the differences between oscar and frama in mind, I'll try to reduce the number of files of frama theme to use oscar resources.

ldidry:
Hello, I just reworked on that theme in order to use as much as oscar resources as possible. Could you reconsider merging it ?

Thank you.

ldidry:
I forgot: you can test this theme on https://framabee.org. The default theme is another one (which is really close), so you'll need to change your preferences to use the `frama` theme.

return42:
Five years later, I can't say why this PR has was not further processed. I close the PR .. ask to reopne if there are any interest in.

Issue #236, opened 2015-02-13
Outgoing Proxy Support


nRaecheR:
I'm trying to get my local Searx instance to work behind an outgoing proxy server. I've fiddled around with all kinds of http(s)_proxy/HTTP(S)_PROXY settings in the environment, without success.

Environment is Ubuntu Server 14.10, amd64. As incoming proxy an Apache2 instance is used. 

Is there any way to configure an outgoing proxy in Searx ?

asciimoo:
Currently there is no HTTP(S) proxy support. As a workaround you can experiment with external transparent SOCKS proxy wrappers like `tsocks`.

dalf:
According to http://docs.python-requests.org/en/latest/user/advanced/#proxies
HTTP(S)_PROXY are supported. 
I miss something ?

nRaecheR:
I've tried this 

```
$ export HTTP_PROXY="http://myproxy:Port"
$ export HTTPS_PROXY="http://myproxy:Port"
$ python
>>> import requests
>>> requests.get("http://example.org")
```

and it works but Searx don't use the proxy setting from global environment. Does it use python::requests at all?

dalf:
How do you start Searx ?

Using a virtualenv and starting searx in debug mode with `python webapp.py` the HTTP(S)_PROXY variables are used.

I guess it's an issue with uswsgi / supervisord

nRaecheR:
Yes, it's with uswsgi and Apache as web server.

ghost:
It will be useful if I can set HTTP(S) proxy in searx's config file.

asciimoo:
implemented: #297

PR #297, opened 2015-04-25
[enh] Implement http proxies for outgoing requests. (see #236)

dalf:
`requests` package is used at different places : search.py, webapp.py, and in engines (see wikidata and google engines). Two solutions : either we are very vigilant are using the proxy everywhere, either 
the package `poolrequests` uses the defined proxies even it is not defined in the parameters.

This pull requests implements the second solution (exception : the image proxy which uses directly `requests` package)

The new setting `outgoing_proxies` can be set. It is directly pass are a parameter to request : http://docs.python-requests.org/en/latest/user/advanced/#proxies

SOCKS proxies are not supported : see https://github.com/kennethreitz/requests/pull/478

### DIFF
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -66,8 +66,10 @@ class SessionSinglePool(requests.Session):
 
 
 def request(method, url, **kwargs):
-    """same as requests/requests/api.py request(...) except it use SessionSinglePool"""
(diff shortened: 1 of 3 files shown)


asciimoo:
Very-very cool, thanks!
The new settings format proposal also (https://github.com/dalf/searx/wiki/settings.yml)

PR #298, opened 2015-04-26
[fix] SSL : SNI support

dalf:
In some engines (kickass, btdigg, photon), SSL check is disabled because without it, there is a SSLError.

This pull request fixes this, and improves SSL security using `pyopenssl` package, see below.

To what I understood, the searx error is the lack of [SNI](http://en.wikipedia.org/wiki/Server_Name_Indication) support. 

The package `pyopenssl` can bring that with others things  : https://urllib3.readthedocs.org/en/latest/security.html#openssl-pyopenssl

> By default, we use the standard library’s ssl module. Unfortunately, there are several limitations which are addressed by PyOpenSSL:  (Python 2.x) SNI support, (Python 2.x-3.2) Disabling compression to mitigate CRIME attack.

`requests` supports this fix if `ndg-httpsclient` and `pyopenssl` are available : https://github.com/kennethreitz/requests/commit/18857a0eedcebbfc40bb1ae431daed935cd51d56

So this pull request : 
- adds the required packages
- includes the [certifi](http://certifi.io/en/latest/) package
- removes `params['verify'] = False`

Note : 
- SNI seems supported in [python 2.7.7](https://www.python.org/dev/peps/pep-0466/#new-security-related-features-in-python-2-7-maintenance-releases) and python 3
- pyopenssl can be slow to start : https://github.com/pyca/pyopenssl/issues/137

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,8 @@ lxml
 pyyaml
 pygments
 python-dateutil
+ndg-httpsclient
+pyopenssl
+pyasn1
+pyasn1-modules
+certifi
--- a/searx/engines/btdigg.py
+++ b/searx/engines/btdigg.py
@@ -29,10 +29,6 @@ def request(query, params):
     params['url'] = search_url.format(search_term=quote(query),
                                       pageno=params['pageno']-1)
 
-    # FIX: SSLError: hostname 'btdigg.org'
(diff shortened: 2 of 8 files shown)


dalf:
I don't know how people upgrade theirs searx instance. One `git pull` will be not enough, they have to run `pip install -r requirements.txt` (or equivalent), if they don't they will see new "bugs".

asciimoo:
Great, thanks!

Issue #264, opened 2015-03-03
rel="noreferrer"


dalf:
Adding rel="noreferrer" to the result links would help to "hide" searx : 
https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer

Firefox support : ok since version 33 : https://bugzilla.mozilla.org/show_bug.cgi?id=530396
Chrome support : ok : https://code.google.com/p/chromium/issues/detail?id=345783
IE : I don't know

ghost:
searx/searx/templates/default/result_templates/code.html

<a href="{{ result.url }}">{{ result.title|safe }}</a>
--->
<a href="{{ result.url }}" rel="noreferrer">{{ result.title|safe }}</a>

dalf:
implemented: #299

PR #299, opened 2015-04-26
Noreferrer

dalf:
Implements #264 using two methods : 
- https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer
- http://w3c.github.io/webappsec/specs/referrer-policy/ (draft, 3 March 2015)

Why : 
- with GET requests of searx, the search text won't be pass to the target web site.
- the target web site won't even know the existance of the searx instance (with either a GET or a POST request on searx)

Note : 
- without this pull request, the referrer is hidden when the searx instance uses the https protocol, and the target web site uses the http protocol.
- not all browser support this feature. A solution for the user : check the referrer in the searx results page. If it is set, then the browser doesn't support this feature. 

How to test : search for "what is my referrer" on the searx instance to test. There is link to www.whatismyreferer.com (other solution search for ipconfig.me )


asciimoo:
\o/ thanks!

PR #300, opened 2015-04-26
settings.yml clean up : remove locale

dalf:
locale was declared for bing\* engines and vimeo.
- bing\* engines use the language settings.
- vimeo uses neither locale nor language settings.

This avoid confusion


asciimoo:
Thanks!

PR #303, opened 2015-05-01
Revert of #195 when the search language is not english

dalf:
Sometimes there is two requests to google (depending of the source IP) : the first to google.com which redirect to the second : google.fr (for instance).

Going to https://www.google.com/ncr and saving the PREF cookie for future use prevent this (there is no redirection).

But, recently (or not ?), by doing this the search returns English results even if the Accept-Language is specified.

There is still a way to prevent this : going to preference, set the search language. I don't know if this can be done by searx.

For now, a quick fix is to disable the use of the PREF cookie when the search language is not English (google engine will slower but returns excepted results).

Note : sending the Accept-Language header when getting the .../ncr URL doesn't help.

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -76,7 +76,8 @@ def request(query, params):
                                       query=urlencode({'q': query}))
 
     params['headers']['Accept-Language'] = language
-    params['cookies']['PREF'] = get_google_pref_cookie()
+    if language.startswith('en'):
+        params['cookies']['PREF'] = get_google_pref_cookie()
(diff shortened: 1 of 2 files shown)


dalf:
One complex solution : 
- merge #215
- but detect if there is a redirection to /sorry..., in this case the engine uses google.com eventually with the PREF cookie (it can be done one time at the start)

As I said, a complex solution...

asciimoo:
Thanks!
As you said, we still need a proper solution.

PR #305, opened 2015-05-02
[enh] reduce the number of http outgoing connections.

dalf:
Engines modified by this pull request : 
- google autocompleter
- deviantart (including thumbnail). Without this PR, the https request redirects to the http website. 
- digg for thumbnails
- google image for thumbnails

Engines that still use http : 
- gigablast
- bing image for thumbnails
- 1x
- dbpedia autocompleter


searx/engines/www1x.py
```diff
@@ -19,8 +19,8 @@
 categories = ['images']
 paging = False
 
-# search-url
-base_url = 'http://1x.com'
+# search-url, no HTTPS
```
pointhi: I think the comment is wrong, because this engine would use HTTPS after merging that pull request.

dalf: Oups sorry... 
that's not the comment but the code which is wrong : 1x.com doesn't respond on https request (there is a https://secure.1x.com with an invalid certificate)

searx/engines/www1x.py
```diff
@@ -27,6 +27,7 @@
 # do search-request
 def request(query, params):
     params['url'] = search_url.format(query=urlencode({'q': query}))
+    print params['url']
```
Cqoicebordel: You should delete this print

asciimoo:
Thanks!

PR #308, opened 2015-05-02
update versions.cfg to use the current up-to-date packages

dalf:
Update of the pull request #98 using the same tool : https://requires.io/github/dalf/searx/requirements/?branch=versions_upgrade

pep8 is updated so there is new rules.
I don't know if this is the right way to fix.

`make robot` returns with the exit code 2 even if all tests has been passed.

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -28,7 +28,7 @@ from searx.poolrequests import get as http_get
 
 
 def get(*args, **kwargs):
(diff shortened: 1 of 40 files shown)


asciimoo:
Thanks, also the pep8 fixes!
The problem is in the tear down function of robot tests, i have a fix for it

dalf:
:-)

PR #310, opened 2015-05-02
[fix] yahoo_news engine

dalf:
The language doesn't seem to be taken into account. But the https://{lang}.news.yahoo.com have a completely different skin.


asciimoo:
Thanks!

PR #320, opened 2015-05-12
[enh] currency_convert engine : "1 dollars in euros"

dalf:
The currency names are fetched wikidata and stored into a static file : searx/data/currencies.json (not at runtime, only one time)

This file is loaded when the currency_converter is loaded.
A database is perhaps more appropriated.

### DIFF
--- a/searx/data/currencies.json
+++ b/searx/data/currencies.json
@@ -1,0 +1,7655 @@
+{
+    "names": {
+        "francos franceses": [
+            "FRF"
+        ], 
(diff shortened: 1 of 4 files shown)


Cqoicebordel:
If it's read/written only once, it's not important if it's a file or a DB.

If and when the DB we already talked about come to life, we can move it to the DB, but meanwhile, the file is good.

asciimoo:
handy feature, thanks!

PR #327, opened 2015-05-25
Meta keywords are obsolete = removal

HLFH:
Hi,

Meta keywords are obsolete since 2009 according to [Google](http://googlewebmastercentral.blogspot.fr/2009/09/google-does-not-use-keywords-meta-tag.html) and the [Moz Marketing Scientist](http://moz.com/community/q/are-meta-keywords-coming-back#reply_136215), so I removed them.

Thanks in advance,

### DIFF
--- a/searx/templates/courgette/base.html
+++ b/searx/templates/courgette/base.html
@@ -3,7 +3,6 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="description" content="Searx - a privacy-respecting, hackable metasearch engine" />
-        <meta name="keywords" content="searx, search, search engine, metasearch, meta search" />
         <meta name="generator" content="searx/{{ searx_version }}">
         <meta name="referrer" content="no-referrer">
         <meta name="viewport" content="width=device-width, maximum-scale=1.0, user-scalable=1" />
--- a/searx/templates/default/base.html
+++ b/searx/templates/default/base.html
@@ -3,7 +3,6 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="description" content="Searx - a privacy-respecting, hackable metasearch engine" />
-        <meta name="keywords" content="searx, search, search engine, metasearch, meta search" />
         <meta name="generator" content="searx/{{ searx_version }}">
         <meta name="referrer" content="no-referrer">
         <meta name="viewport" content="width=device-width, maximum-scale=1.0, user-scalable=1" />
--- a/searx/templates/oscar/base.html
+++ b/searx/templates/oscar/base.html
@@ -3,7 +3,6 @@
 <head>
     <meta charset="UTF-8" />
     <meta name="description" content="Searx - a privacy-respecting, hackable metasearch engine" />
-    <meta name="keywords" content="searx, search, search engine, metasearch, meta search" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="generator" content="searx/{{ searx_version }}">
     <meta name="referrer" content="no-referrer">
--- a/searx/templates/pix-art/base.html
+++ b/searx/templates/pix-art/base.html
@@ -3,7 +3,6 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="description" content="Searx - a privacy-respecting, hackable metasearch engine" />
-        <meta name="keywords" content="searx, search, search engine, metasearch, meta search" />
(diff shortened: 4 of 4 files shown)


pointhi:
I think we should keep the keywords. Probably google and co. didn't use it, but there is a chance that other search-engines are using them.

ghost:
I doubt you'd look up a search engine on another search engine, though.

HLFH:
@pointhi If some crazy search engines still use it, it's to help you to unrank your searx website... The best SEO practice is not to use meta keywords because it's now considered as an old spam/black hat technique. A lot of SEO bloggers agree with that. **Instead of using the obsolete meta keywords, you have to optimize your content and set up smart keywords in your posts/page to drive action.** [Matt Cutts from Google agreed with that](https://www.youtube.com/watch?v=jK7IPbnmvVU). The meta keywords were useful when AltaVista was leading the search engine market, and it's closed since 2013... Even the well-known SEO French community agreed that you should not use meta keywords. http://www.webrankinfo.com/dossiers/debutants/meta-keywords ; and it's an article for beginners... the ABC of SEO, so. The triumvirate Google, Bing and Yahoo ignore the meta keywords.

@asciimoo Hi, what do you think of this small PR? Could it be merged?

dalf:
[noise]
The "new" meta are : 
- [microdata](https://en.wikipedia.org/wiki/Microdata_%28HTML%29)
- [open graph](https://en.wikipedia.org/wiki/Facebook_Platform#Open_Graph_protocol) (used by FB and Google)

[/noise]

Cqoicebordel:
@HLFH I think you are misquoting the article you linked.
1. The article said that **most** search engines don't look at them, but not all. Yandex, for example, may look at them. Yandex represent 42% of searches in Russia. That's not small.
2. The article say you must delete them only so your competitors don't know what keywords you are targeting.
3. The article say also that keeping them won't hurt **at all**

So why deleting them ? Even if they are not used by most search engines crawlers, some might, and it doesn't hurt to keep them.

@plsng Why not ? "Privacy respecting search engine". Bam. :)

HLFH:
@Cqoicebordel Well... Let's take old figures (2012) from the product qSearch of ComScore. 
search-marketshare
In fact, the opponents to the US supremacy Baidu and Yandex still use meta keywords. At the end of 2012, they represented 11% of global market share, according to ComScore. The US triumvirate Google, Bing, and Yahoo represented 72.6% of market share. Should we keep the meta keywords because we should care of these 11% that will probably stop using the meta keywords or should we use the best practices used by those 72.6% (and probably more...). If you disagree, we should vote.

Cqoicebordel:
A vote is not the way of searx. It's the consensus, based on arguments.

You have shown that the keywords are not used by at least 73% of the market. Fine. But why oh why should we remove them ?
They are not hurting, they are not harming searx. In fact, they are useful for 11% of the market, that reprensent ~330,000,000 potential users (of the 3 billions Internet users).

Let's make a pros and cons list :
Keeping them : 
- Pros : Might be helpful for the 11% search market share
- Cons : We are not in the advance guard of technology

Removing them :
- Pros : We are in the advance guard of technology
- Cons : Might hurt on 11% search market

For me, the decision is clear. But if you have other arguments, I'm ready to listen :)

HLFH:
Ok. Incroyable ce désaccord entre parisiens :'( 
I believe the main arguments have been said. Thanks for the disagreement, it always helps to find new relevant informations. I removed searx. I like bleeding-edge (here = Microdata, Open Graph, Twitter Cards), so - sadly or not - I'll keep Google as my main search tool. Have a nice day. Maybe I'll come back when [searx 0.9 will be released](https://github.com/asciimoo/searx/milestones/version%200.9.0).

Cqoicebordel:
C'est dommage de partir juste parce que tu n'es pas d'accord sur un point :)
The point of this tool is indeed I think, not being bleeding edge, but stable enough to accomodate the most people possible (including the long tail), while respecting privacy. So we have to think of everybody, in Russia, in Africa using cellphones, in the silicon valley.
Anyway, you are welcome to come back any time. And maybe not wait for the 0.9, but instead the 0.8 ;)

(oh and btw, just so you know, Yahoo still uses keywords in its own homepage (granted, not Google, Bing, Yandex or Baidu do the same ;) ))

HLFH:
I'm back :0 https://searx.hlfh.space/

PR #333, opened 2015-05-27
[fix] Piratebay engine

### LINKED ISSUE #330: Piratebay engine : update the URL
The current URL is https://thepiratebay.se which doesn't respond.
This one seems to work https://thepiratebay.mn  (see https://en.wikipedia.org/wiki/The_Pirate_Bay )

Another solution : disable / remove the engine in settings.yml

dalf:
Fix #330

Fix the URL, and disable the engine by default by the history of this website seems "strange" now.
May be it should be completely deleted.

### DIFF
--- a/searx/engines/piratebay.py
+++ b/searx/engines/piratebay.py
@@ -20,7 +20,7 @@ categories = ['videos', 'music', 'files']
 paging = True
 
 # search-url
-url = 'https://thepiratebay.se/'
+url = 'https://thepiratebay.am/'
 search_url = url + 'search/{search_term}/{pageno}/99/{search_type}'
 
 # piratebay specific type-definitions
@@ -41,10 +41,6 @@ def request(query, params):
     params['url'] = search_url.format(search_term=quote(query),
                                       search_type=search_type,
                                       pageno=params['pageno'] - 1)
-
-    # FIX: SSLError: hostname 'kthepiratebay.se'
-    # doesn't match either of 'ssl2000.cloudflare.com', 'cloudflare.com', '*.cloudflare.com'
-    params['verify'] = False
 
     return params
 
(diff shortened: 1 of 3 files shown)


Cqoicebordel:
If this engine is deleted, so must be kickass, for the same reasons.
But I don't think it's the solution.

Maybe, in the case of those search engines, a fallback possibility should be implemented -> Rotating domain name based on history of the local searx instance (already tried and failed or succeded this domain name)

dalf:
Good point.

Two solutions :  
- make the request, in case of failure, switch to another domain. The user doesn't have result for one request.
- It requires to test the different domains, and choose the right one in background.

The second solution can be done at the load time, but if the searx instance has a big uptime, the test has to done again, and there is nothing if the searx core to do that easily ( yes a timer inside the engine can be a solution, but if each engine starts to do the same it will be a mess ? )

asciimoo:
Do you have any information about these "new" piratebay sites/domains?
Who runs it and where?
Are there any differences in the content comparing to `thepiratebay.se`?

dalf:
The thepiratebay.am is registered to Fredrik Neij (co-founder of TPB).
The registar is http://www.1abcdomain.com/
The resolved IP is  104.18.48.8 which belong to cloudflare.

Not so much information...

Cqoicebordel:
The new pirate bay is set to change frequently of domain name, following seizure by authorities.
They have changed their logo to become an hydra (don't know the right word -> mythical creature that sprouts several heads when one is cut). It's the official TPB.

So their extension are .mn, .gd, .la, .am, .gs and .vg. And I think .am was already seized, but I didn't follow that that closely.

The solution could be a mix of both your solutions @dalf :
- Each request, test the next one on the list that the last test was positive and don't display any result if negative
- In background retest regularly every negatives' ones.

But what should be done also, is a way to configure that easily for the admins : this should go into the conf, with a rotator between it and the engines, so this solution isn't only applicable to TPB, but also to others in the future.

asciimoo:
.am is officially listed on https://en.wikipedia.org/wiki/Thepiratebay so merging as a tmp fix.

PR #336, opened 2015-05-27
startpage and ixquick enignes : disabled by default, and timeout set to 6 seconds.

### LINKED ISSUE #329: Startpage engine broken
There is no result when I'm using the bang "!startpage"

dalf:
Fix #329

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -203,12 +203,15 @@ engines:
   - name : startpage
     engine : startpage
     shortcut : sp
+    timeout : 6.0
+    disabled : True
 
-# +30% page load time
-#  - name : ixquick
-#    engine : startpage
-#    base_url : 'https://www.ixquick.com/'
-#    search_url : 'https://www.ixquick.com/do/search'
+  - name : ixquick
+    engine : startpage
+    base_url : 'https://www.ixquick.com/'
+    search_url : 'https://www.ixquick.com/do/search'
+    timeout : 6.0
+    disabled : True
 
   - name : twitter
     engine : twitter


asciimoo:
Totally agree with disabling both by default, they are quite slow and unstable, but the 6sec default timeout isn't too much? It can cause three times slower responses in general category if one of these enabled. Personally i don't really like longer responses than 3sec.
Another alternative is a preferences option to set the timeout parameter individually (for each engine?).

dalf:
6 sec is based on the average response time I saw on searx.me. It's average so there is slower time!  

If the engine is disabled by default, I think the use case is to enable it using a bang, and in this case, 6 sec is long but it is better than no result. I forgot the other use case : enable the engine.

Some option about the time out are perhaps a solution, but this PR is quick & dirty fix that resolved the global response time, and try to make those two engines usable. Purpose : release a 0.8.0 version without referenced bugs.

asciimoo:
You can enable them from the preferences too and there is no sign or warning about the significant response time increase which can be confusing - and i know, timeout and no results are confusing too.. =)

dalf:
Ok, so let's switch back to 3 seconds ?

asciimoo:
Overall i think it would be better at first.
devs?

Cqoicebordel:
I think the 6sec + disabled was a good quick and dirty fix.

But yes, we need more :
- Being able to set the timeout in the prefs (with a bit of info about latency)
- Having a tiny note indicating what search engines timeouted on the result page

And yes @asciimoo you are right too. We can't distribute a version which has this kind of issue (-> no warning of the big latency if enabled)

asciimoo:
@Cqoicebordel yes, a small notification or a box on the right side with the engine timeouts would be useful too!

dalf:
Another idea : why set a time out when there is only one engine ? If the engine take time to respond, then searx also, but since there is nothing else to show, it's not issue.

So why not change the time out behaviour : 
- if there is only one engine : set the time out to 60 seconds (yes, because there is a connection pool, so if the user send a lot of requests they will have an end )
- if there is more than one engine : normal time out behaviour.

Cqoicebordel:
That could work too, but I'm afraid that the user will refresh the page sooner that the timeout, and so, resending the request, and reseting the timeout to 0 each time, keeping himself away from the results.

dalf:
see #343

asciimoo:
Nice solution @dalf! Thanks!

asciimoo:
ah, sry, i merged my local branch, but it's in the master: aac8d3a7bfdd77a5369e52a4ece99b20669a4625
closing

PR #343, opened 2015-05-30
[enh] oscar theme: display engine response times in the preferences

dalf:
Oscar theme :   
- display avg response time (display in red if it is more than the default time out)
- display time out (display in red if it is more than the default time out)
- display safe search support

[Screenshot](http://postimg.org/image/nu2pp52g3/)
### WARNING :
- the rtl doesn't work
- translations not updated

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -71,6 +71,9 @@ def load_engine(engine_data):
     if not hasattr(engine, 'language_support'):
         engine.language_support = True
 
+    if not hasattr(engine, 'safesearch'):
+        engine.safesearch = False
+
     if not hasattr(engine, 'timeout'):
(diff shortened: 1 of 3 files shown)


dalf:
More notes : 
- the "language support" is not shown : it is always True
- the result count is not shown : wikidata and duckduckgo_definition has zero results but add infoboxes : it can be misleading for the users.
- the "Allow / Deny" button is on left part (on ltr languages) for small screen : there is a horizontal scrolling, but the most important informations are show : allow / deny and engine names. A better solution could be : https://css-tricks.com/responsive-data-tables/

asciimoo:
Very-very cool! +1 reason for the stats refactor, i'm not sure, that the numbers are always correct - because of timeouts/errors/etc

PR #344, opened 2015-05-30
google engine

dalf:
Avoid some "sorry google" by adding another cookie : NID. This cookie is specific by google hostname (it is set once).

This allow to send request directly to google.\* (according to the search language).
Before this commit, request in other languages than English was sent to www.google.com which was redirected to www.google.* (www.google.hu, www.google.fr, etc...)

The PREF is still use on the www.google.com domain.

Moreover, when the sorry page is given an exception is thrown, which can be seen in the statistics.

WARNING : need some testing on different IPs / location. May be google has a different behaviour.

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -14,18 +14,76 @@ from lxml import html
 from searx.poolrequests import get
 from searx.engines.xpath import extract_text, extract_url
 
+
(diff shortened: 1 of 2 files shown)


asciimoo:
I've tested on some IPs, seems it works, thanks!

Issue #331, opened 2015-05-26
Youtube engine is broken


dalf:
The only result result returns by the youtube engine is to https://youtube.com/devicesupport

dalf:
I'm not sure if the API v2 is still working (it is deprecated since last year).

And the v3 API requires a key.

I suggest the same implementation than the flickr engine : 
- an engine based on scraping ( see https://www.youtube.com/results?search_query=test without javascript)
- an engine based on the API v3 if we have time to write it.

Cqoicebordel:
I'm waaaaaaayyyyyyy behind on searx dev, but I like writting engines, so I'm OK to do it, probably within the next week.

asciimoo:
@Cqoicebordel that would be awesome, thanks!

Issue #326, opened 2015-05-25
Add engine Qwant 


beudbeud:
Qwant is a french search engine respectful of privacy

dalf:
Seems doable : 
https://api.qwant.com/api/search/web?count=10&offset=0&f=&locale=fr_fr&q=qwant
the response is in JSON format.

Replace web by knowledge, social, news, suggest, smart_top (no count or offset for this one).

dalf:
Done

beudbeud:
thanks

PR #346, opened 2015-05-30
Adds engines : Youtube with or without API and multiple Qwant

Cqoicebordel:
The API needs an API_KEY.
The NOAPI doesn't have the published dates.

I haven't deleted the old youtube engine just in case, it could be useful in the future if Google changes its mind, but I doubt it.
For the published dates, the page I had was in French, and I was a bit lazy to tweak the cookies to have it as uniform as possible so the date detection would work. I think it's good enough for now, but tell me if I'm wrong. Youtube doesn't provide precise date on its search page anyway, so it would be hard to display something useful.

EDIT : solve #331


Cqoicebordel:
Added a few Qwant engines : 
- Web
- Images
- News
- Social (seems to provide only results from Twitter, but search efficiently)

A few more possibilities are available from the Qwant 'API' : 
- Knowledge graph (with keyword 'knowledge')
- smart_top
- shopping
- videos (but this one provide only videos from youtube, so I skipped it)

Solve #326 

(and sorry for putting it in the same branch as youtube, it was an error on my part)

asciimoo:
Great! What about implementing all the Qwant engines in a single file to handle different categories like the piratebay engine or swisscows (#348)?

Cqoicebordel:
The difference is that for each type, in qwant, it's a different source of data, like bing or google.
It allow a bigger granularity than TPB for ex. by allowing disabling the images or news sources.
I don't think it would be a good idea to put them in a single file.

dalf:
Perhaps what we miss here it is inheritance using Class and Object.
It could help to have common code at one place.

For now, we have two options :
- With one python module per category it is easy to understand the query and the schema of the result.
- With one python module for all categories, the copy / paste is avoided, but we have to carefully read the code to understand the schema of the result for each category.

Cqoicebordel:
Yes , there is a bit of code redundancy here. But if we look at all the engines, there is _a lot_ a code redundancy.

In this case, I think I used the right approach, because if there was only one python module, it must have done four (4) requests by itself (because it's 4 different urls), and that's not the pattern we used so far : 
Until now, we did a module per request (url).

So, I'm not sure why suddendly it's an issue blocking this merge ?

dalf:
I forgot : my comment was not about why we should block the merge of this PR. Only a brainstorming that should be another "issue". @asciimoo : merge ?

asciimoo:
What's the problem with defining more than one engines/categories in one file if the code is nearly identical? Don't repeat yourself. If there are redundant parts in the code, reduce them instead of writing new ones.
With the current solution an API change requires to modify the code in 4 different places, imho it's not the right approach.

Cqoicebordel:
Ok, I understand.
But in this case, since it's using 4 different URLs, each API can change independantly.
If you want, I'll put them all in a single file, but I strongly think that's the wrong approach here.

Cqoicebordel:
Nevermind, I understood, and I'll propose another way tonight.

Cqoicebordel:
Sorry it took me so long to understand. I tought for a moment you wanted me to perform four requests simultaneously.

Tell me if this way is ok.

asciimoo:
ah, yes, so pretty and much simpler. Just a minor note: maybe a dict which maps categories to keywords would be better than setting `search_url_keyword` in the config.
The actual search category is available from the `request()` function as `params['category']` and as `resp.search_params['category']` from the `response()` function in each engine.

Cqoicebordel:
Agreed, it's much better. With a dict too.

And thanks for the pointers. They weren't the right ones, but thanks anyway ;)

asciimoo:
=]
Thanks!

pointhi:
youtube_api is not working for me:

```
ERROR:searx.search:engine crash: youtube
  File "/usr/local/searx/searx/search.py", line 41, in search_request_wrapper
    return fn(url, **kwargs)
  File "/usr/local/searx/searx/poolrequests.py", line 80, in get
    return request('get', url, **kwargs)
  File "/usr/local/searx/searx/poolrequests.py", line 73, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/sessions.py", line 465, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/sessions.py", line 579, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/hooks.py", line 41, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/usr/local/searx/searx/search.py", line 107, in process_callback
    search_results = callback(response)
  File "/usr/local/searx/searx/engines/youtube_api.py", line 56, in response
    videoid = result['id']['videoId']
KeyError: 'videoId'
```

Furthermore, the normal youtube api is also not returning results.

Issue #347, opened 2015-05-31
Add swisscows search engine


bitwave:
Swisscows (https://swisscows.ch/) is an interesting search engine, which could be integrated into searx...

pointhi:
I have added it to the wiki: https://github.com/asciimoo/searx/wiki/possible-search-engines. Probably I'm implementing it soon, because the search-quality is very good.

thx

bitwave:
+1 cool thx

PR #348, opened 2015-06-01
implement swisscows engine

pointhi:
I have implemented general and image search using swisscows.ch, regarding to #347


asciimoo:
cool! Thanks!

Issue #319, opened 2015-05-11
autocomplete : add startpage


dalf:
As suggest by @fr330n1y, startpage can be added as an autocomplete backend.

Example : 
https://startpage.com/do/suggest?query=aaaa

PR #349, opened 2015-06-01
Add startpage as an autocompleter engine

Cqoicebordel:
As the title say.
Solve #319


dalf:
Thank's !
Line 161 there is a `print`

Cqoicebordel:
Ah fu\* yes... It happens to everyone ;)

Issue #351, opened 2015-06-03
qwant news: unable to set language


cy8aer:
Maybe my fault: I am not able to set another language as french on qwant news/social. My global language setting is "Automatic" - but for qwant news I would like to have german or english. Tried
- locale : de-DE
- language : de-DE
- language : German

but everything fails. Setting global to german it works fine. For me this is the first situation for setting a special language for a service.

cy8aer:
Works. Thank you.
(language: de_DE)

PR #353, opened 2015-06-03
New feature : language per search engine !

Cqoicebordel:
If an user wants to define a specific language for a unique search engine, now he can.

In settings, only add the code for the language (en_US, fr_FR, de_DE...), and if the engine can take it into account, it will.

Solve #351


asciimoo:
Clever solution, thanks!
A note would be good about this feature in the engine overview wikipage - or maybe a standalone configuration page

Cqoicebordel:
I'll do a new page describing the settings.

PR #354, opened 2015-06-04
Bing news based on RSS output

### LINKED ISSUE #352: Bing_news based on the RSS output ?
The bing_news engine crash again on searx.me, searx.oe5tpo.com.

It doesn't crash when I use searx from my DSL connection. I guess there is different version of the website.

Should we rewrite the engine using the RSS format ?

Example: 
https://www.bing.com/news/search?q=searx&setmkt=en-US&first=1&format=RSS

dalf:
Fix  #352

### DIFF
--- a/searx/engines/bing_news.py
+++ b/searx/engines/bing_news.py
@@ -6,18 +6,17 @@
               max. 5000 query/month
 
  @using-api   no (because of query limit)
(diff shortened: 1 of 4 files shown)


asciimoo:
Thanks!

PR #356, opened 2015-06-05
fix little bug inside settings, improve oscar template

pointhi:
**fix**: When someone deactivated all engines from a defined categorie, the categorie wasn't displayed anymore inside the settings.

**enh**: use of table instead of container-fluid in the cookie tab, to provide a continuous design and to show more contrast between the single rows.


asciimoo:
Thanks!

PR #357, opened 2015-06-05
[enh] google engine : parse map links and more

dalf:
When there is a map in the google results, extract :
- a link if there is one (to google plus or somewhere else, most probably the official website)
- geoloc
- address & phone number

The map can show one precise location, or a set of location (when searching "town + restaurant"). In the later case, there is only a set  of links and phone numbers (there is no precise geoloc of the different links).

In case of a precise location, the "map.html" template is used to show the precise location directly inside searx. There is some "issues" about that : 
- the official website can into the result of other engines, but since the "map.html" template is used, the two results are not merged.
- the map.html is "grouped" at the beginning (can be ok) or the middle of the other results (not easy to find).

A infobox is also added to make thing more clear (google this information in the same way). If the location could be shown with an embed map, the "map.html" result could be removed.

I feel like the default template for result need some extension.

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -282,7 +282,8 @@ def response(resp):
                 results.append({'url': url,
                                 'title': title,
                                 'content': content})
-        except:
+        except Exception, e:
+            print e
             continue
 
     # parse suggestion
@@ -305,7 +306,8 @@ def parse_images(result, google_hostname):
                         'title': '',
                         'content': '',
                         'img_src': img_src,
-                        'template': 'images.html'})
+                        'template': 'images.html'
+                        })
 
     return results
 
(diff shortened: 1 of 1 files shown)


pointhi:
There is a Licence issue, because it is not allowed to display google-map-results on other maps than from google. Furthermore, merging of results is very hard.

dalf:
There is a trick : the engine gets the geoloc from the google link nothing more. Then OSM is used to display the location. Is there a licence issue in this case ?

pointhi:
I'm not sure, but probably (what I know from google maps).

What I know sure is, that coordinates which are comming from google's geolocation are not allowed to be imported into the osm-database. Anyway, the google geolocator is not very accurate.

dalf:
Would it be ok without the map links ?
In this case, there is no merging of results issue.

pointhi:
I think this shouldn't be a problem.

dalf:
Question (already implemented, but easy to remove) : use the [geo URI](http://en.wikipedia.org/wiki/Geo_URI), nothing is display, it's only a link.

asciimoo:
Thanks!

PR #360, opened 2015-06-06
[enh] implement simple spell checker as plugin, FIX #189

### LINKED ISSUE #189: detect spelling error using provided suggestions

pointhi:
it's the same spellchecker like in the pull-request #190, but implemented as plugin and without usage of a defined wordlist.

Fixing issue: #189


asciimoo:
Could you add some tests, to demonstrate the usage/efficiency/performance?

asciimoo:
Sorry for the long-long delay, I reviewed the code at last.
The functionality is ok, but it doesn't fit to the current plugin system. Implements custom variables/functionality in templates and webapp. Two possible solutions:
- Improve the plugin system and templating
- Rewrite it as a standard preferences option which is really easy since the user settings refactor

pointhi:
hm, the thing why I changed some implementation in the webapp and template was because I added a new field in the template specific for such data. I'm not sure how to achive this in another way.

At least, we could extend the current template system to be able to add custom variables easily, but nevertheless it's somehow required to change the template to parse this new datasets.

asciimoo:
I understand the purpose of the modifications, it is just out of a plugins scope. Yes, we can extend the template system to handle these cases, but i think it's easier and less effort to rewrite the pr as a standard preferences option like the autocompleter.

ibrokemypie:
personally i think extending the scope of plugins to useful functionality like this would be much better than just adding them as features. independently developed plugins could add to the growth without adding bloat, plugins being independant to each instances choice.

return42:
@pointhi is this PR any longer relevant?

return42:
no feedback, I guess we can close this PR.

return42:
@unixfox lets close this PR / thanks!

Issue #359, opened 2015-06-05
currency engine : regex


dalf:
https://github.com/asciimoo/searx/blob/master/searx/engines/currency_convert.py#L12
the regex should be something like 

```
^\W*(\d+(?:\.\d+)?)\W+([^.0-9].+)\W+in?\W*([^\.]+)\W*$
```

some `\W*` are replaced by `\W+`, may be a better one could be found.

fro now,  the currency engine catches `3d dinosaures` : 
http://searx.me/?q=3d+dinosaures&category_general=on

Cqoicebordel:
Lol
Maybe, an easier way would be to include the spaces around 'in' in the regex.

asciimoo:
fixed: #362

Issue #355, opened 2015-06-04
Error 500 when converting currencies with non US-ASCII symbols


David-5-1:
Trying to convert currencies using non US-ASCII symbols (with € and £ or seemingly any character : ø, à, …) results in Internal Server Errors.

```
curl 'https://searx.me/?q=10 $ in €'      -> Error (500)
curl 'https://searx.me/?q=10 $ in EUR'    -> OK (200)
curl 'https://searx.me/?q=€ø…'            -> OK (200)
```

Cqoicebordel:
Confirmed
I'll look into it tomorrow.

Cqoicebordel:
Ok, I give up.
It's clearly an unicode issue. But python 2.7 is a real pain to handle this.
And so I've tested lots of solutions, but found nothing usable.

If anybody know how to have a string with non-ascii chars encoded as \uXXXX, that could be a good approach to a solution.

asciimoo:
i'll take a look on it

pointhi:
probaly using encode like inside the image_proxy? (https://github.com/asciimoo/searx/blob/master/searx/webapp.py#L260)

Cqoicebordel:
Hmmm, didn't try that, but could help a lot.
But there would still be an issue, because not all money chars are defined by their htmlentities. I mean, it would take a bit of work to check every data in currencies.json, and add the correctly encoded money symbols.
Or it could be easy, by replacing every \u by something like #xx;

Anyway, I can't work on it this we, and my brain is fucked up by all this encode/decode/unicode/unidecode/str stuff to do it properly, so I passed the torch on this one ;)

Cqoicebordel:
Finally found it, but lost a bit of hair in the process.

asciimoo:
=))
Thanks again!

PR #362, opened 2015-06-07
Allow use of chars as € in currency converter

Cqoicebordel:
Encoded the regex and the query, so that the regex doesn't eliminate spaces chars part of the UTF-8 chars.

Retooled a bit the regex, so `3d dinosaures` doesn't match it anymore as suggered in #359.

Solve #355 too.


asciimoo:
nice fix, thanks!

Issue #316, opened 2015-05-09
tracking/affiliate link remover plugin


pointhi:
I would propose a plugin idea, to remove known tracking and affiliate links from the result url.

For example, amazon urls sometimes containing the url argument `tag` which is an affiliate-id. We could improve that list with other known url arguments of sites, which are only used for tracking and affiliat, and create a plugin which is deleting those parts of the url.

dalf:
+1

Cqoicebordel:
Awesome idea !

privacytoolsIO:
I can help with that. I'm a big affiliate marketer. A website of mine is currently participating in over 3000 affiliate programs.

dalf:
@privacytoolsIO : great :-) 

If you can give a list of website and affiliate parameters / patterns, it would be awesome.

The implementation will be something that look like the [https_rewrite plugin, line 225](https://github.com/asciimoo/searx/blob/master/searx/plugins/https_rewrite.py#L225)

Cqoicebordel:
@privacytoolsIO If you can provide me a list of trackers arguments in URL, or anything similar, it would be welcome to enhance what I already commited #365 :o)

privacytoolsIO:
@Cqoicebordel Sorry, I'm late. Just found this:
https://s3.amazonaws.com/lists.disconnect.me/simple_ad.txt

This is used by the plugin "Disconnect" and should cover most of ads / affiliate links.

Cqoicebordel:
Nice !
This list could be very useful to remove unwanted links !

But I was talking more about the arguments in the URLs, like `?utm_origin=XXX` etc.
If you know about others than 'utm' it would be very useful. If you don't, don't worry :)

PR #365, opened 2015-06-09
Add a plugin to remove trackers from results URLs

Cqoicebordel:
Add a tiny plugin that remove the trackers arguments (like 'utm_*') in URLs returned after a search.

I was not able to test it with a real life test, because I couldn't find an example where a search engine would return such URLs.

First draft of #316

### DIFF
--- a/searx/plugins/tracker_url_remover.py
+++ b/searx/plugins/tracker_url_remover.py
@@ -1,0 +1,40 @@
+'''
+searx is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+searx is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with searx. If not, see < http://www.gnu.org/licenses/ >.
+
+(C) 2015 by Adam Tauber, <asciimoo@gmail.com>
+'''
+
+from flask.ext.babel import gettext
+import re
+
+re1 = re.compile(r'utm_[^&]+&?')
+re2 = re.compile(r'(wkey|wemail)[^&]+&?')
+re3 = re.compile(r'&$')
+re4 = re.compile(r'^\?$')
+
+name = gettext('Tracker URL remover')
+description = gettext('Remove trackers arguments from the returned URL')
+default_on = True
+
+
+def on_result(request, ctx):
+    url = ctx['result']['url']
+
+    url = re1.sub('', url)
+    url = re2.sub('', url)
+    url = re3.sub('', url)
+    url = re4.sub('', url)
+
+    ctx['result']['url'] = url
+    return True
--- a/searx/plugins/tracker_url_remover.py
+++ b/searx/plugins/tracker_url_remover.py
@@ -18,10 +18,9 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 from flask.ext.babel import gettext
 import re
 
-re1 = re.compile(r'utm_[^&]+&?')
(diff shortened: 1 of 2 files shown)


searx/plugins/tracker_url_remover.py
```diff
@@ -0,0 +1,40 @@
...
+You should have received a copy of the GNU Affero General Public License
+along with searx. If not, see < http://www.gnu.org/licenses/ >.
+
+(C) 2015 by Adam Tauber, <asciimoo@gmail.com>
+'''
+
+from flask.ext.babel import gettext
+import re
+
+re1 = re.compile(r'utm_[^&]+&?')
```
asciimoo: this could be a list/tuple, which contains the regexps.

searx/plugins/tracker_url_remover.py
```diff
@@ -0,0 +1,40 @@
...
+re2 = re.compile(r'(wkey|wemail)[^&]+&?')
```
asciimoo: ^^^ replaces e.g. `http://wkey.com/` to `http://`
maybe checking the query params from `result['parsed_url']` would be a safer solution

searx/plugins/tracker_url_remover.py
```diff
@@ -0,0 +1,40 @@
...
+re3 = re.compile(r'&$')
+re4 = re.compile(r'^\?$')
```
asciimoo: is the above regex necessary? it matches only the `?` string

Cqoicebordel:
Hum... I agree, it needs retooling.

But :
- result['parsed_url'] has always empty arguments
- I found those regex in a JS to remove tracker, I forgot they only applied to the arguments parts. I'll retool them.
- That explain the '?' string : after removing all arguments, if there is only a '?', we can also remove it. The order of the subs is important.

Agreed also on the list for the regexes.

(It pisses me off when I produce bad code like that)

Cqoicebordel:
@asciimoo ok, tell me if this is better.

searx/plugins/tracker_url_remover.py
```diff
@@ -0,0 +1,44 @@
...
+           re.compile(r'(wkey|wemail)[^&]+&?'),
+           re.compile(r'&$')}
+
+name = gettext('Tracker URL remover')
+description = gettext('Remove trackers arguments from the returned URL')
+default_on = True
+
+
+def on_result(request, ctx):
+    splited_url = ctx['result']['url'].split('?')
```
dalf: Perhaps, just in case `.split('?', 1)` or use`result['result']['parsed_url'].query`

Cqoicebordel:
For the `split('?', 1)`, I thought about it, but it's more useful to test the lenght after, and doing nothing if the length is not 2. It eliminates the "no query" and the "hmmm there is something bad in this URL" type or results.
As for `result['result']['parsed_url'].query`, when I tested it, it always had empty arguments.

So I think I used the most usable way of doing things. But, as always, I may be wrong.

asciimoo:
that's strange, `result['result']['parsed_url'].query` shouldn't be empty. I've tested with `https://aa.bb/cc?dd=ee#ff` and it contained the `dd=ee` string.

Cqoicebordel:
How can you try with a link you provide ?

I'll test again. Maybe I messed up, but I don't think I did.

Anyway, and off topic, isn't it redundant to have the url and the parsed_url in the same var ? It could be tricky too, if you change one but not the other.

Cqoicebordel:
Yeah. I messed up.

I looked at params, and not at query...

Is it good this time ?

asciimoo:
yap, perfect, thanks!
Storing the full and the parsed url is a bit redundant, i agree, but can be handy sometimes (e.g. result merge)

Cqoicebordel:
Ah finally ! :)

For the redundancy, for me, the issue is not so much that it's redundant, it's that inevitably, in the future, a bug will arise from that, because one value would have been changed but not the other.
And yes, it's handy, but it's pretty easy to infer one from the other, at any time : it's just a urlunparse/urlparse away...

dalf:
One way, it's to have an Class like in this project :  https://github.com/gruns/furl/

asciimoo:
ah, yes, furl would be an elegant solution!

PR #369, opened 2015-06-12
Adds a FAQ to explain the settings page

Cqoicebordel:
I tried to explain every settings and panels, but I may have not done a great job.
A link is added in every available description in the settings page.

I choose not to include a link in the header, it would have been too weird to have on every page a link to the FAQ for the settings.
Maybe, when the FAQ will be enhanced to include **every** function of searx, we could add a link to it in the header.


dalf:
Hé cool, but is it more a help / documentation than a FAQ ?
More over, there is only a version for oscar theme, not for the other themes. How does it behave with the others themes ?
Anyway, thank's !

Cqoicebordel:
Yeah, you are right, it's not formally a FAQ, but I used the user point of view, and tried to answer the questions "What is X ?" each time.
I had some return from non-geek trying searx saying that they didn't understand what autocomplete meant, for example, and that's what I tried to answer.
Also, I think common users are more used to the form of a FAQ that give quick answers than documentation.

As for only a version for Oscar, I looked to include this in other themes, but Oscar is the only one providing a beginning of info on items, where I could add a link "More info...".
Beyond that, what I think is more important, is providing info to the default theme (not the theme named "default", but oscar, in our case), so the first time an user uses searx, he has access to it. If he changes something in the settings, I think that this particular user is not a new comer anymore.

Anyway, that was my point of view, but if you want me to change anything, don't hesitate to tell me :)

asciimoo:
Very cool! Useful info about the preferences, wouldn't be better as small tooltips or descriptions near the features?

Cqoicebordel:
Hum... Yes !
But I don't know how to do it without JS.
I'll look into it.

dalf:
With the title attribute ? 
- http://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_global_title
- http://www.w3schools.com/tags/att_global_title.asp

Cqoicebordel:
Ok, I (finally) changed it to icons+hover text (title indeed), so, no JS.

**BUT** I made the icon myself, and it's... I don't have a word for it. Abysmal ? Horrendous ? Despicable ? You get the point.
So, I need your help on that one to build an icon that fits gracefully in the UI.

And it might also be useful to do icons for the others themes, so I can integrate this FAQ to them.

return42:
It looks like no one is working on this solution anymore. I'll close it for now.

Issue #368, opened 2015-06-12
Vimeo engine : website has changed


dalf:
The XPath have to be adapted.

asciimoo:
fixed: #370

PR #370, opened 2015-06-12
Fix vimeo

Cqoicebordel:
Fix XPaths for Vimeo, and add a bit of testing in swisscow, yahoo_news and youtube_noapi to improve coverage.

Solve #368


asciimoo:
Thanks!

PR #371, opened 2015-06-13
Add a self user agent plugin

ldidry:
Just like with the "ip" query, duckduckgo gives the server's information with the "user agent" query.
This corrects this behavior by adding a plugin based on self_ip.py plugin.


dalf:
Thank's. I don't know if the code souldn't be merged into self_ip.py (and renamed self_ip.py to self.py or something like that) ?
From the user point of view, these features are about displaying information about the browser.

@asciimoo what do you think ?

Cqoicebordel:
I wasn't asked, but I think it's indeed a good idea to merge the two.

Also, @ldidry, the regex should be compiled outside the function, in the class : in the function, at each request the regex is compiled, in the class, it's compiled only once.

ldidry:
I wait for the decision about merging the two plugins in one and will do it as soon as I have a "go for it"

asciimoo:
Good idea, +1 to the merge

ldidry:
Merged :smiley:

asciimoo:
great, thanks, two small notes:
- The name `self` can be misleading in python it could be changed to `self_info` or `client_info` or something similar.
- Description lacks of info about the usage/keywords - and the reference to DDG can be confusing

ldidry:
Changed.

asciimoo:
cool, thanks!

PR #374, opened 2015-06-21
[fix] reduce server traffic by optimizing images

haasn:
Simply running ‘optipng’ on every PNG image reduces the overall size of
them by 10% on average.


asciimoo:
Thanks!

PR #384, opened 2015-07-09
implement sympy as plugin (CAS-System)

pointhi:
I implemented a Computer Algebra System as plugin, using [SymPy](http://www.sympy.org).

current Features:
- calculation of simple equations like: `1+2*3/5`, `sin(1)`, `sqrt(2)`,...
- solving of equations with one unknow variable: `15*a+3=30`, `sin(3)=x*2`,...
- solving of simple integrals: `integral of sin(x) dx`, `integrate e^2*x dx`,...

TODO:
- graphical output of functions
- add possibility to insert most SymPy Commands and output them with correct formating

Because I have to use MathML for formating, I changed the answere system to work more similare linke the results (to allow different key-value pairs). Now Answers could contain html-code, or simple text. I think we should improve the answere system in the future, to allow things like thumbnails,...

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -20,6 +20,7 @@ from searx import logger
 logger = logger.getChild('plugins')
 
 from searx.plugins import (https_rewrite,
+                           math_solver,
                            self_info,
                            search_on_category_select,
                            tracker_url_remover)
@@ -72,6 +73,7 @@ class PluginStore():
 
 plugins = PluginStore()
 plugins.register(https_rewrite)
+plugins.register(math_solver)
 plugins.register(self_info)
 plugins.register(search_on_category_select)
 plugins.register(tracker_url_remover)
--- a/searx/plugins/math_solver.py
+++ b/searx/plugins/math_solver.py
@@ -183,9 +183,8 @@ def post_search(request, ctx):
                                               'label': 'sympy'}]
                 return True
 
-    except Exception,e:
+    except Exception, e:
         print str(e)
-
 
     if regex_integral.match(ctx['search'].query):
         ''' Solve integral equations
--- a/searx/plugins/math_solver.py
+++ b/searx/plugins/math_solver.py
@@ -15,6 +15,7 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 (C) 2015 by Thomas Pointhuber
 '''
 from flask.ext.babel import gettext
+from searx.plugins import logger
 from sympy import simplify, solve, Symbol, count_ops, S
 from sympy.printing.mathml import mathml
 from sympy.utilities.mathml import c2p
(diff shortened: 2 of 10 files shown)


dalf:
Whaou, nice !

Some notes : 
- perhaps add sympy in requirements.txt
- chrome has removed the mathml support, one workaround : https://github.com/mathjax/MathJax
- `(10*1.1) eur in usd` works in google (I know that's suppose to include in sympy in an engine) 

We need a way to have engine without HTTP request (perhaps sympy parsing is done after all requests and slow down the global response time).

dalf:
Another thing : `cos(4*x)+sin(2*x)=1` has more than one results, and the global width is too large (the results are hidden by the suggestions).

Perhaps add a line break between the results ?

dalf:
One more thing : there is no timeout

a equation like `cos(x)^2+sin(2*x+pi)=tan(1)` can easily create a DOS.

pointhi:
yes, but I don't know how to do it properly (From what I found out, there is no clear way in python to handle timeouts of functions): http://eli.thegreenplace.net/2011/08/22/how-not-to-set-a-timeout-on-a-computation-in-python/

pointhi:
ok, I found out using signals would be working for sympy, but we cannot use them in threads: `ValueError: signal only works in main thread`.

I tried to implement https://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread-in-python, but I didn't got it working. Any other idea?

Cqoicebordel:
I would recommand that you try to do it the hard way : 
Create a new thread containing the solve(), and if it's still alive 10s later (or whatever the time is set in pref), you kill it (if it's still alive, it didn't return the solution and killed itself).
It's not classy, but should work.
Look at https://docs.python.org/release/2.7/library/multiprocessing.html and the answer with 25 votes in your link.

ibrokemypie:
any updaye? looks like an awesome idea

dimqua:
Any chance it will be merged any time soon?

schwukas:
I'll give this one a bump. I think this would be a nice extra especially for users switching from popular search engines to searx. Is it only the review that is missing or does it actually need extra code? I would certainly try to help, but I'm not very proficient when it comes to threads in python,

tbergeron:
I really wish this could be integrated. Whenever I have to do quick maths I used to type it in the browser bar and any search engine would return the results but now I have to start the good old calculator app (or worse! use my BRAIN! hehehe)

Is there any reason it's taking so long to get integrated? Is there anything we could do about it? Having complex calculation support would be great but I don't believe it's needed for a first step towards integrating arithmetic into the engine.

return42:
> I really wish this could be integrated.

In order not to end up in a dead end, we should first consider how best to embed such tools into the search syntax. See my [comment at PR1146](https://github.com/asciimoo/searx/issues/1146#issuecomment-611003952).

PR #385, opened 2015-07-09
Add Image on opensearch

ldidry:
When you add Searx as search engine in your browser, you don't have the favicon in your search engines list.

This fixes this behavior by specifying Image in `opensearch.xml` (http://www.opensearch.org/Specifications/OpenSearch/1.1#The_.22Image.22_element).

I had to fix the URL with a `replace` filter, otherwise the URL looks like `http://example.org//static/themes/oscar/img/favicon.png` which is not correct. If someone has a better idea…


asciimoo:
Thanks!

PR #395, opened 2015-07-19
btdigg engine: fix text mojibake.

gugod:
Seeing on the official instance, searching with non-ascii character
usually renders mojibake non-sense:

```
http://searx.me/?q=%E9%A3%9F%E5%A0%82&categories=files
```

Screenshot with broken: 

Screenshot of the fixed rendering:


dalf:
Nice.
The `searx/tests/engines/test_btdigg.py` has to change.

Example, line 26 (same line 104 and 370) : 
`response = mock.Mock(text='<html></html>')` 
should be changed to 
`response = mock.Mock(content=bytearray('<html></html>'))`

(not tested)

Also, according to travis, there is an issue with the selenium tests, I really don't know why.

gugod:
@dalf Thanks for the pointer.

I modified the tests and made them pass. I only ran `make tests`.

asciimoo:
@gugod thanks!

PR #409, opened 2015-09-01
Duckduckgo url encoding

beniz:
Revealed while working on something else. Basically, pickle was failing on ddg results, and the URL was the culprit. Hope it helps.


beniz:
oops, wrong rebase, will PR it again.

Issue #255, opened 2015-02-22
result cache / temporary save


xavierle:
a search engine find results, store them and then serve prepared results in a very fast way.

actual searx work as a very great metasearch engine.

but to make it server restuls a faster way
would there any practical way to cache pages results for some time

so that during that period of time searx only serve the results very fast without searching again for them ?

as you know better searx code than me
at what level would you recommend me do try that ?

any searx function ?
uWSGI cache function ?
nginx cache function ?

thank you

Cqoicebordel:
I think there would be some issues there : 
- First, it would be a privacy issue : if you do a search and searx response time is low, you know somebody else on the server have already made the same request.
- Secondly, if it should work in a cross users way, the chance that two persons would make the exact same search in a short time period is almost nil.

So for me the only use case would be when you do a search, click on a link, and then go back to the result page (with back button). In some cases, the search is launched again, and yes, this shouldn't happen.

As searx doesn't save anything, except stats in a macro way, and doesn't plan to save anything else, I'll recommend you to dig in uWSGI and/or nginx and/or Apache, and probably more on the two latter ones, because they are those serving the static (rendered) pages.

And please bring us anything you will find, we are very interested :)

xavierle:
thank you for fast awser

actually the group of users testing my install are focus on privacy but not that strong
and they would expect answers to go quite fast like for example with duckduckgo

i am checking if i find something with uWSGI / nginx

apache is usually not the way to make thing faster

my idea actually would be to cache any served pages for some hours/days : 

http://searx.instance/?q=keyword&pageno=1&category_general

and to find a way that the cached page is served preferably

dalf:
One idea : use the browser cache with the help of the ETag HTTP header.

Idea of implementation : 
- Create a hash of all parameters (query, browser language, selected engine based on query and cookie values) and add a ETag HTTP header.
- Store this ETag and an expiration date
- When there is a new request, check If-None-Match HTTP header : it's the previous ETag value stored by the browser. If the two ETag match and are not expired, then searx can response with the HTTP status [304](http://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection)
- Important part : disable POST, and only use GET

Improved idea : 
- the ETag is a hash of all parameters, an expiration date can be added. For example next day at midnight. When the result expired, the ETag changes.
- In this case
  -  there is no need to store a map (ETag -> expiration date).
[...]

PR #411, opened 2015-09-01
Result storage with access from the API in order to retrieve past results without calls to the underlying search engines

beniz:
This PR adds:
- Storage of search results when 'Storage Results in preferences tab is set and/or storage_results is 1 in settings.yml
- Retrieval of stored results when they exist, via the `c=1` parameter in URL from the command line

Technicals:
- storage uses Python [shelve](https://docs.python.org/2/library/shelve.html)
- when storage is activated, all queries and results are stored, without filter
- storage key is the query, so issuing query with different search engines will overwrite existing results in storage (i.e. results from different engines via multiple queries will _not_ be merged)

Usage:
- query 'test' from Web interface
- get stored results:

```
curl -XGET "http://localhost:8888/search?q=test&c=1&format=json"
```

Result should appear instantaneously.

Considerations
- This PR is primarily for my own use, aiming at sharing search results in a p2p net, there'll be no hard sentiment if it is rejected :)
- The storage of results obviously must be considered as a privacy leak (i.e. unless you're willing to share, which is not the issue here). In case this PR is of interest to some, there are some ways to mitigate the privacy issue, and they can be discussed here or elsewhere in subsequent additions.

Thanks for searx, this is a tremendously well executed piece of software, simple and easy to hack, which is why I doubly enjoy it :)

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -72,7 +72,7 @@ def response(resp):
         # append result
         results.append({'title': title,
                         'content': content,
-                        'url': res_url})
+                        'url': res_url.encode('utf8')})
 
     # return results
     return results
--- a/searx/search.py
+++ b/searx/search.py
@@ -359,7 +359,8 @@ class Search(object):
 
         # set caching
         if self.request_data.get('c'):
-            self.cached = True
+            if self.request_data.get('c') == '1':
+                self.cached = True
         
         # set pagenumber
         pageno_param = self.request_data.get('pageno', '1')
--- a/searx/search.py
+++ b/searx/search.py
@@ -17,6 +17,7 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 
 import threading
 import re
+import shelve
 import searx.poolrequests as requests_lib
 from itertools import izip_longest, chain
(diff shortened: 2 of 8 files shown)


beniz:
Oops, forgot to mention one of the issues: the results are stored into `results.db` which is written from where the webapp.py is started. Obviously this cannot left as is in case of a merge. I've notice that the setup.py arranges for a .searx file or repository but I haven't tested this type of installation yet.

Given proper instructions on a good location for `results.db`, I am willing to make the required changes.

beniz:
Weird, `make robot` passes just fine for me with the fix, difficult to tell what's up with Travis right here...

EDIT: started from scratch with same cloning as Travis, and `make robot`, so... <grunt>.

return42:
@beniz what do you think, is this PR any longer relevant?

unixfox:
It's not a bad idea, but it is heavily coded only for a very old searx version. I'm closing this PR but the issue related to this PR is still opened here: https://github.com/searx/searx/issues/255

beniz:
@return42 @unixfox hello there! This smells like code I don't remember having worked on :) Deserves a kill I guess, thanks for the ping!

PR #416, opened 2015-09-04
Migrating Travis CI from legacy to container-based infrastructure

pointhi:
Documentation about this changes:
- http://docs.travis-ci.com/user/migrating-from-legacy/
- http://docs.travis-ci.com/user/caching/

If all is working correctly, searx should be build faster on the container-based infrastructure. Probably we also fixed the partial build errors while installing the packages, because they are now cached (and not downloaded every time)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 cache:
   - pip
   - npm
-  directories:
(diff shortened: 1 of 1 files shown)


asciimoo:
great, thanks!

Issue #265, opened 2015-03-04
split core and ui


dalf:
For now, the core and ui are not clearly separated.

Example : search.py use directly request.cookies from flask framework.

A clear split between the two parts would help : 
- to switch to another UI framework is needed (to solve the the database "issue", or some other reason)  
- to create a nice command line tool easily

A suggest to create two python packages : `searx.core` and `searx.flask`.
`webapp.py` could stay in the same location to avoid trouble with existing installations (uwsgi to reconfigure etc...).

There is no hurry, and may be not relevant.

pointhi:
I have not so much experience with pip and python, but could we implement a plugin-system in that way easily?

Because if #60 would be implemented, and we add functionalities like calculator or qr-code creator, much new dependencies are required. Using plugins which can be installed later and maintaining there own dependencies would help to let the searx-core slim, and easier to maintain.

PR #427, opened 2015-09-10
Refactored user setting management

kantord:


pointhi:
+1

I have done a rough review of the code, and the basic code/structure looks good to me. But it need some rework and bugfixes:
- Please add some comments to the most important functions/objects, add copyright header.
- Loading default values doesn't work (for example default-theme)
- some Tests doesn't work
- it would be better to use `Accept-Language` or another language source by default instead of `"locale": "en"`.
- `blocked_engines` not initialized using settings.yml, the same with `disabled_plugins`.
- disabling plugins is not working
- probably other things I haven't spotted.

kantord:
@pointhi thanks for the review. I'll get to it soon. About comments: what parts of the code, especially, which methods are most obscure or hard to understand? Beyond adding docstrings, I'd consider renaming them, too.

pointhi:
I would add at least class descriptions.

Furthermore, methodes which are using functions like `hasattr(self, "verify_%s" % setting)` are not easy to understand, especially for python beginners (I required some time to understand the behaviour of this code parts). Same with methodes like `override_from_form` which are doing much stuff.

Most of the code is self explanatory and doesn't need comments (which is greate :+1:).

searx/user_settings.py
```diff
@@ -0,0 +1,239 @@
...
+        if hasattr(self, "deserialize_%s" % setting):
+            return getattr(self, "deserialize_%s" % setting)(value)
+        else:
+            return value
+
+    def restore_from_cookies(self, cookies):
+        for setting, value in cookies.iteritems():
+            self.set(setting, self._deserialize(setting, value))
+
+    def save_to_cookies(self, cookies):
```
asciimoo: Should simply return with the dict object (line 97). With this minor modification we can remove the `cookies` parameter to avoid side effects. The function name could be `get_as_cookies` or something like this.

searx/user_settings.py
```diff
@@ -0,0 +1,239 @@
...
+    pass
+
+
+class UserSettingsBase(object):
+    defaults = None
+    _collection_fields = {}
+
+    def __init__(self):
+        self._settings = dict()
+        self._is_empty = True
```
asciimoo, line 35: This can be retrieved from `self._settings`

asciimoo:
Really-really cool, especially the test coverage! =)
I have some code specific note, check the annotated diff.
Another small issue: instead of the `%s` style formatting please use the `.format()` function.

kantord:
@pointhi I concur with your stand on `hasattr()` and friends. Documenting this with docstrings, I find a bit challenging, though. (I mean, is it something that should be in a docstring?) I added comments at least, and also made `verify` and `validate` protected, since they aren't and should not be ever called from outside settings management classes. I'd also like to add that, though it currently seems to be an overkill, it'd be better to use objects for each setting field so that the class could call their methods instead of putting all these methods into a single messy class. This could also remove a lot of code duplication, but I'm not sure that it'd worth implementing it currently.

searx/user_settings.py
kantord: This was originally designed like this so that the class would not remove cookies not handled by it, which was handy when doing TDD. (This way I could chunk up the refactoring process into tiny bits, so only one test failed at a time). However, now that no cookies unknown to the class are allowed, this change can indeed be made.

searx/user_settings.py
kantord, line 35: `as_cookie_dict` now basically handles the serialization process. I don't think it would be a good idea to store values serialized. So I'm afraid we do need at least some algorithm here :( Kinda just a `map`, though, but it may be too ugly to make it a one-liner. What do you guys think?

Cqoicebordel:
I didn't have the time to review the code, so my comment won't be that useful. But please, when you commit, _always_ write a comment. It's easier for everyone if you do.

kantord:
@Cqoicebordel good point. I plan to rearrange this into a single commit, with a more helpful commit message.

pointhi:
how is the current status of this pull request? It would be cool to finish this in some time and merge it.

dalf:
one step for #265 (split core and ui)

asciimoo:
closing - see #534

PR #435, opened 2015-09-17
Correct the loop through search engines in Pix-Art

### LINKED ISSUE #432: List of search engines is not displayed in pix-art theme
It appears list of search engines is intended to appear in pix-art theme, but no list appears.

Cqoicebordel:
Fix #432


asciimoo:
Thanks!

PR #436, opened 2015-09-17
Add JSON api support.

moritan:
This hack let you start the searx server in Rest mode (with the rest-server.py ).
 This allow to use searx in an app with json. 
I use it to create a dinamic app with ember.js. ( code incomming).

This hack remove the usage of cookie for settings(better for privacy).
Delegating templating and translation to the client will give more resources on the server.

Removing IHM from the server engine, allow to have a better integration in an existing site.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 flask
 flask-babel
-flask-cors
 requests
 lxml
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 flask
 flask-babel
+flask-cors
 requests
 lxml
 pyyaml
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -69,6 +69,21 @@ class PluginStore():
 
         return ret
 
+    def callAPI(self, plugin_type, user_plugins, *args, **kwargs):
+        ret = True
+        for plugin in user_plugins:
+            if hasattr(plugin, plugin_type+'API'):
+                ret = getattr(plugin, plugin_type)(*args, **kwargs)
+                if not ret:
+                    break
+            elif hasattr(plugin, plugin_type):
+                try:
+                    ret = getattr(plugin, plugin_type)('', *args, **kwargs)
+                    if not ret:
+                        break
+                except:
+                    ret = True
+        return ret
 
 plugins = PluginStore()
 plugins.register(https_rewrite)
--- a/searx/plugins/self_info.py
+++ b/searx/plugins/self_info.py
@@ -43,6 +43,7 @@ def post_search(request, ctx):
         ctx['search'].answers.add(ua)
     return True
 
+
 # attach callback to the post search hook
 #  request: flask request object
 #  ctx: the whole local context of the pre search hook
--- a/searx/plugins/self_info.py
+++ b/searx/plugins/self_info.py
@@ -42,3 +42,17 @@ def post_search(request, ctx):
         ctx['search'].answers.clear()
         ctx['search'].answers.add(ua)
     return True
+
+# attach callback to the post search hook
+#  request: flask request object
+#  ctx: the whole local context of the pre search hook
+def post_searchAPI(ctx):
+    if ctx['search'].query == 'ip':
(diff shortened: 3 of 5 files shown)


Cqoicebordel:
Ok, but what are the advantage of your solution against using searx with `format=json` option ?
Because this option will indeed bypass the template engine, not using as much ressources as the classic IHM way.

Just asking to better understand what you did :)

moritan:
First, it's return more info than "format=json" option ( format=json return just search result, not answers ,suggestions...)

My hack don't use cookies, but a json structure for settings, i think it's a more clean way to manage settings.

I thinks it's more easy to create an site base on searx this way, because you could update the engine without to have to merge each time your ihm change.

If you use javascript framework like Angular, ember.js, or even just javascript, your client side could be serve by nginx/apache/etc, . 
You could even do the client side with PHP , but it's not so hype ;)

Cqoicebordel:
I agree on what you want to do, but it seems to me that it's not the right way to do it :
From what I can gather from your code, you are duplicating the central engine of searx to provide the right JSON response.

From my POV, the right way to do it would have been to :
1.  Extend the current `format=json` to add in it what is missing
2.  Add the JSON settings directly into the main engine, so that if present, the JSON settings will replace the defaults and/or the cookies' ones.
Which would have limited the duplication of code, while giving so much more possibilities.
With this solution, you would also still be able to update the engine easily, since your UI is elsewhere, and it would be also possible to serve it using Apache/Nginx (or even none of this as the engine can be started with `python searx/webapp.py`)

But that's only my opinion. Anyone else ?

pointhi:
I agree with @Cqoicebordel, extending the current json format would be much better.

moritan:
Oh yes, i should have precise something,  I have duplicate 2 files webapp.py and search.py.
Of course, this should be re-merge in one file.
To be clear, it's the first time i'm coding in python ( i'm a Java dev), so i create new file to avoid break existing code. 
This PR main goal is to exchange with you ;)

I read somewhere, there is a refactor of settings in progress, so i propose a posibility with searchAPI.py, but i'm agree that "The can be Only one". (Duncan McLeod get out of this code !)

i don't have clear idea for your point 1. My first version of code update the /search entrypoint to add missing element, but it was not pretty.
Error return HTML page, by example and the resust was a big method not readable with "if format == 'json' " everywhere.

So i decide to split entrypoint in 2, one (the current) for the Human interaction with html templating  and the " /api/\* " entypoint for machine interaction.

I'm still not sur of something about the starting class. 
If i want to start in rest-only mode ( to serve an android app for example), the actual api only is good because, it won't expose entrypoints that you won't use. The same way if you don't want to expose api.
But if you want both ?

It's really depend  on how you want to use searx engine and i don't know if flask allow to enabled/disabled entrypoint by settings.

I will try to modify webapp.py to be able to use my version of search.py and merge rest-server.py in it.

Cqoicebordel:
You'll see, python isn't that hard ;) (I'm from Java and C++ myself)

Ok, so just to be clear, this PR won't/shouldn't be merged.

What I'll recommand is that you try just to do the point 1 and do a PR _just on that_. No new entry point, just using what is already in place, enhancing it by adding what is missing.
We will better be able to guide/help you on smaller code than on the whole thing you already commited.

I don't think we need multiple entrypoint to separate api or not. `format=json` should (must ?) suffice.
And why care about enabled entrypoint if they are not visible ? If you wish to embed searx in an app, you still can, even if the UI entry points are enabled, they won't be used, and won't use ressources.

As for the HTML error page, I'll look into it.

In any case, start small, and build on that :)

asciimoo:
@Cqoicebordel <3
@moritan is it ok if i close this?

roscopecoltran:
Is the JSON API a feature officially added to searx ? it would be awesome and flexible

virtadpt:
As far as I know, the answer is "Yes."  Here's the documentation:

https://asciimoo.github.io/searx/dev/search_api.html

roscopecoltran:
Awesome, my bad, I missed this part.

Thanks a lot !

return42:
I think we can close this PR / no answer on https://github.com/searx/searx/pull/436#issuecomment-168447341

PR #452, opened 2015-10-24
[enh] fix content fetching, parse published date from description for startpage and ixquick

pointhi:
I fixed the broken description parsing for startpage and ixquick, and furthermore added parsing of publishedDate (which is embedded in the description).

### DIFF
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -12,6 +12,8 @@
 
 from lxml import html
 from cgi import escape
+from dateutil import parser
+from datetime import datetime, timedelta
 import re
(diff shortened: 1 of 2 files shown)


Cqoicebordel:
(there is no commit for ixquick. Is that wanted ?)

pointhi:
ixquick and startpage using the same engine, and also the same layout (I think it's the same company)

Cqoicebordel:
Sorry, jumped the gun a bit too quickly ;)

asciimoo:
@pointhi, great, thanks! =)

Issue #199, opened 2015-01-22
Inaccurate statistics


dalf:
the /stats URL shows different numbers when the page is refreshed.

Explanation (I guess) : the statistics are calculated per process (with uwsgi at least).

It's not a big issue except (perhaps ?) for the error counts.

pointhi:
I think we can fix that if we are using a database instead local variables to store the stats.

Furthermore, a restarted searx-engine can build on the previous data like `score`.

asciimoo:
Collect stats related improvements/ideas to #162

PR #477, opened 2015-12-25
[enh] measure response time with more details.

### LINKED ISSUE #162: enhanced stats-functions
I'm not sure if this feature would be good for searx, but it would be interresting adding an enhanced stats-site.

I'm thinking of things like `searchquerys/day or hour`. Things which are interesting, but nothing which is decreasing the privacy of the users. (General stats over all users)

This is only an idea, no "must have" :-).

dalf:
First step to fix #162
The shown statistics are the same, except in memory there is a lot more information.
The second steps would be to re-implements UI. See https://github.com/dalf/searx/tree/metrology

Notes : 
- there is no tests
- the stats  are per unix process, so if there more than one as uwsgi does, it won't be accurate ( #199 ). The current implementation has already this issue.
- the idea of implementation.py is to add a redis / sql / whatever backend later. This would fix the previous note, except we have to take into account the speed. One idea : recording a measure only records an action to do later. A dedicate thread reads all these actions and executes them. Note : two unix processes equals two dedicated threads to record measures.

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -19,8 +19,6 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 from os.path import realpath, dirname, splitext, join
 import sys
 from imp import load_source
-from flask.ext.babel import gettext
-from operator import itemgetter
 from searx import settings
 from searx import logger
 
@@ -92,14 +90,6 @@ def load_engine(engine_data):
                          .format(engine.name, engine_attr))
             sys.exit(1)
 
-    engine.stats = {
-        'result_count': 0,
-        'search_count': 0,
-        'page_load_time': 0,
-        'score_count': 0,
-        'errors': 0
-    }
-
     if hasattr(engine, 'categories'):
         for category_name in engine.categories:
             categories.setdefault(category_name, []).append(engine)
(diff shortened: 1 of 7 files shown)


pointhi:
To share data between uwsgi threads we could propably using [SharedArea](https://uwsgi-docs.readthedocs.org/en/latest/SharedArea.html)?

dalf:
I'm not sure about the score stats.

Also I saw a small issue with request time measure per engine : it's include the call to the request fonction (since it's based on params['started'] ). 

Perhaps the per engine measure should also include the time out / error requests, after all if one engine has 20% of timeout each time the user has too wait for that.
But here there is an issue : how to get the params['started'] in the function wrapper (the one with the try / catch ).

After bring back the graph is not too difficult, nearly done actually except I would like to double check the consistency.

@pointhi : I don't know this projects why not. The only : measure response time mustn't be slow.

dalf:
In hurry, I mess up : I pushed the metrology2 branch to asciimoo/searx instead of dalf/searx and added my settings.yml...

By the way, there are the graphs in /stats2
Warning : all measures have to be checked, double checked, there is always tricky parts...

asciimoo:
no problem, force push allowed on non-master branches =)
..and I'll check it, thanks

return42:
@dalf what do you think about this PR? .. if it has a value, it might be better tagged as RFC .. otherwise close?

dalf:
I've started to rewrite this PR because many things have changed since the time I wrote that PR.
I close this one and will open another PR,.


Also, I would like to keep the https://github.com/searx/searx/tree/metrology2 branch until the new one is pushed.

PR #478, opened 2015-12-28
Wolfram Alpha API

MarcAbonce:
Added Wolfram Alpha as an engine.

It's disabled by default because it's very slow and it requires an API key which (when free) allows only 2000 queries per month (enough for personal use though).

Right now it only works when the answer is a line of plain text because my response returns an 'answer' rather than a 'result'.


a01200356:
To test it you can use this key: '5952JX-X52L3VKWT8'

asciimoo:
wow! Very cool, thanks! =)

PR #481, opened 2015-12-30
Better Docker embedded support

Wonderfall:
Hi, I'm a Docker size freak (and beyond that, a Docker lover)! I've been maintaining wonderfall/searx (Docker Hub) for a while, and I decided to give something back to searx : it's a minimalist Dockerfile (less than 100MB once builded) based on Alpine Linux, and it works flawlessly.
- Completely rewritten from Alpine Linux 3.3 (base).
- It installs essential packages, no more, no less.
- It has less layers (always preferred).
- You can easily set a base_url, or enable/disable image_proxy before building.
- More readable (easy to maintain).
- ... it just works.

This Dockerfile was made to be build manually so it's slightly different from mine which is on Docker Hub : for example, the secret key is generated during the build time, not the first time the container is started (I don't recommended to make a public automated build). And of course I do my best to follow best practices for writing Dockerfiles.

Please feel free to discuss about it if you want!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,14 @@
 FROM alpine:3.3
-MAINTAINER Wonderfall <wonderfall@opmbx.org>
 
 ENV BASE_URL=False IMAGE_PROXY=False
 
-COPY . /usr/local/searx
+EXPOSE 8888
+
+WORKDIR /usr/local/searx
+
+CMD ["./run.sh"]
+
+COPY requirements.txt /usr/local/searx/requirements.txt
 
 RUN apk -U add \
     build-base \
@@ -18,13 +23,7 @@ RUN apk -U add \
     openssl \
     openssl-dev \
     ca-certificates \
- && adduser -D -h /usr/local/searx -s /bin/sh searx searx \
- && cd /usr/local/searx \
- && pip install --no-cache -r requirements.txt \
- && sed -i "s/127.0.0.1/0.0.0.0/g" searx/settings.yml \
- && sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml \
- && sed -i "s|base_url : False|base_url : $BASE_URL|g" searx/settings.yml \
- && sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx/settings.yml \
+ && pip install --no-cache -r /usr/local/searx/requirements.txt \
  && apk del \
     build-base \
     python-dev \
@@ -35,10 +34,21 @@ RUN apk -U add \
     libxml2-dev \
     openssl-dev \
     ca-certificates \
- && chown -R searx:searx /usr/local/searx \
  && rm -f /var/cache/apk/*
 
-EXPOSE 8888
+COPY . /usr/local/searx
+
+RUN adduser -D -h /usr/local/searx -s /bin/sh searx searx \
+ && chown -R searx:searx /usr/local/searx
+
 USER searx
-WORKDIR /usr/local/searx
-CMD ["python", "searx/webapp.py"]
+
+RUN cd /usr/local/searx \
+ && sed -i "s/127.0.0.1/0.0.0.0/g" searx/settings.yml \
+ && sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml
+
+RUN echo '#!/bin/sh' >> run.sh \
+ && echo 'sed -i "s|base_url : False|base_url : $BASE_URL|g" searx/settings.yml' >> run.sh \
+ && echo 'sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx.setting.yml' >> run.sh \
+ && echo 'python searx/webapp.py' >> run.sh \
+ && chmod +x run.sh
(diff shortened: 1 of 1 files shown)


asciimoo:
@Wonderfall seems good to me, i like alpine, but i'm not a big docker expert.
@glogiotatidis, you wrote the current docker file, what do you think?

Dockerfile
```diff
@@ -1,22 +1,44 @@
...
-
-COPY requirements.txt /app/requirements.txt
-RUN pip install --no-cache -r requirements.txt
-
-COPY . /app
-RUN sed -i -e "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml
+EXPOSE 8888
+USER searx
+WORKDIR /usr/local/searx
+CMD ["python", "searx/webapp.py"]
```
glogiotatidis: Since we're going for optimized dockerfile lines 41, 43 and 44 can be moved at the top, e.g. after line 4, to take full advantage of docker caching even when packages change.

Dockerfile
```diff
@@ -1,22 +1,44 @@
-FROM python:2.7-slim
+FROM alpine:3.3
+MAINTAINER Wonderfall <wonderfall@opmbx.org>
 
-WORKDIR /app
+ENV BASE_URL=False IMAGE_PROXY=False
```
glogiotatidis, line 3: :+1:  for extra settings

Dockerfile
```diff
@@ -1,22 +1,44 @@
...
+    python-dev \
+    py-pip\
+    libffi-dev \
+    openssl-dev \
+    libxslt-dev \
+    libxml2-dev \
+    openssl-dev \
+    ca-certificates \
+ && chown -R searx:searx /usr/local/searx \
+ && rm -f /var/cache/apk/*
```
glogiotatidis, line 42: I cannot really see why having less layers is "always preferred". IMO this is less readable and it loses when it comes to image rebuilding. 

adduser, pip, sed and chown commands have nothing to do with the  apk command, are small and fast commands and can be separated into multiple layers.

Also if you break them into multiple cmds you can move COPY after the package install and therefore reuse the package install layers on new searx versions

Dockerfile
```diff
@@ -1,22 +1,44 @@
...
+    openssl \
+    openssl-dev \
+    ca-certificates \
+ && adduser -D -h /usr/local/searx -s /bin/sh searx searx \
+ && cd /usr/local/searx \
+ && pip install --no-cache -r requirements.txt \
+ && sed -i "s/127.0.0.1/0.0.0.0/g" searx/settings.yml \
+ && sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml \
+ && sed -i "s|base_url : False|base_url : $BASE_URL|g" searx/settings.yml \
+ && sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx/settings.yml \
```
glogiotatidis: Since BASE_URl and IMAGE_PROXY are env variables, these sed commands should be moved in a start up script executed on CMD so users can override them with `docker run -e` on run time.

Dockerfile
```diff
@@ -1,22 +1,44 @@
-FROM python:2.7-slim
+FROM alpine:3.3
```
glogiotatidis, line 1: :+1: nice! I generally go with debian:stable these days instead of python b/c I usually need packages not available in alpine. But in the case it's a perfect match.

glogiotatidis:
Good job @Wonderfall. I like alpine and more flexibility with env variables. Also I prefer more docker cmds for readability and cachability. 

Thanks for pinging me @asciimoo

Wonderfall:
Thanks for your clever comments, @glogiotatidis, I'll take a look.

Wonderfall:
I made some changes, thanks to @glogiotatidis's suggestions.
Now it should take more advantage from Docker caching, resulting in faster builds, but I wanted to keep the same size (I mean it could be even more optimised for caching, I know). 

What do you think? I'm still open to suggestions!

EDIT : Sorry for multiplying commits, I'm not confortable with Git yet! Hope it's fine now.
By saying "it could be even more optimised", let me explain myself : `pip` and `apk` could be separated (may be better for caching), but if I do so, I can't delete unnecessary packages (~100MB wasted).
https://pix.schrodinger.io/#nVrA0l9CILRmqGqt-m5o8w

Dockerfile
```diff
@@ -1,22 +1,54 @@
...
+
+COPY . /usr/local/searx
+
+RUN chown -R searx:searx /usr/local/searx
+
+USER searx
+
+RUN cd /usr/local/searx \
+ && sed -i "s/127.0.0.1/0.0.0.0/g" searx/settings.yml \
+ && sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml
```
dalf: Why not move this line to run.sh ? This would create a secret key  each time the docker image is runed.

Wonderfall: This is totally right, this is what I do for my automated build : the key is generated the first time the container is started. It's not a problem here since the Dockerfile should be manually built : a unique secret key is generated during the build time. But I think like you, that moving this generation to run.sh brings more flexibility.

asciimoo:
You are awesome, thanks =)

glogiotatidis:
Nice! Thanks for updating :)

PR #486, opened 2016-01-06
[enh] WolframAlpha no API engine (and tests for both)

MarcAbonce:
What's new:
    - WolframAlpha no API engine
    - Unit tests for both versions
    - Fixes in API engine (untested code was indeed broken code)

The two versions have some minor differences though:
    - api answers in mathematical notation ("∫1/x dx=log(x)+constant") while noapi answers in plain english ("integral 1/x dx = log(x)+constant"). Both are readable though.
    - api can display multiple answers ("solve x^2+x = 0"). noapi doesn't.

The engine disabled by default, but can be called with !wolframalpha or just !wa


asciimoo:
@a01200356 thanks. There were some pep8 checking issues, could you rebase your pr to the current master?

asciimoo:
@a01200356 thank you!

PR #490, opened 2016-01-10
update & fix install guide after install refactor

kvch:
I updated the installation and added the Docker guide from the wiki.

Also, headings are fixed and TOC is added to the top of the page.


asciimoo:
Thanks =)

PR #493, opened 2016-01-15
Some changes in Dockerfile

cy8aer:
- searx/settings.yml is the right name in second sed line
- COPY requirements.txt . does not work: "Not a directory"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,12 +8,12 @@ CMD ["./run.sh"]
 RUN adduser -D -h /usr/local/searx -s /bin/sh searx searx \
  && echo '#!/bin/sh' >> run.sh \
  && echo 'sed -i "s|base_url : False|base_url : $BASE_URL|g" searx/settings.yml' >> run.sh \
- && echo 'sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx.setting.yml' >> run.sh \
+ && echo 'sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx/settings.yml' >> run.sh \
  && echo 'sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml' >> run.sh \
  && echo 'python searx/webapp.py' >> run.sh \
  && chmod +x run.sh
 
-COPY requirements.txt .
+COPY requirements.txt ./requirements.txt
 
 RUN apk -U add \
     build-base \


Wonderfall:
Thanks for correcting. :+1: 

But `COPY requirements.txt .` works for me (I just builded). I see no reason why it shouldn't.
Where did you get this? What's your docker version?

Otherwise, I want to suggest something. What do you think about making a separate run.sh file ?
Dockerfile : https://pix.schrodinger.io/#V1LdhPcpQS8V7Zetieuhwg
run.sh : https://pix.schrodinger.io/#yPBW4Wr8PwDOpkYg3HJkJw
(Just a suggestion.)

cy8aer:
for the requirements name: I am using docker 1.8.3~ds1-2 (debian testing). But I have problems with simply copying files into a directory instead of naming it for a long time. Looks like this for me:

```
Step 5 : RUN adduser -D -h /usr/local/searx -s /bin/sh searx searx  && echo '#!/bin/sh' >> run.sh  &&  echo 'sed -i "s|base_url : False|base_url : $BASE_URL|g" searx/settings.yml' >> run.sh  && echo 'sed -i "s/image_proxy : False/image_proxy : $IMAGE_PROXY/g" searx/settings.yml' >> run.sh  && echo 'sed -i "s/ultrasecretkey/`openssl rand -hex 16`/g" searx/settings.yml' >> run.sh  && echo 'python searx/webapp.py' >> run.sh  && chmod +x run.sh
 ---> Running in 37501430e394
 ---> 4a53733158d2
Removing intermediate container 37501430e394 
Step 6 : COPY requirements.txt .
stat /var/lib/docker/btrfs/subvolumes/f1325da84c75d4e52b00ab9c263718a6a456f51587769096f498b72b11b3e280/usr/local/searx/requirements.txt: not a directory
```

When I for myself build containers I create separate files every time. This is better to maintain.

cy8aer:
requirements.txt: I am using a btrfs based docker system (running with -s btrfs). And this may be the point why I need to explicitely need to name the target file in copy sections. This may be interpreted in another way like, say, aufs.

Wonderfall:
Okay, this is interesting to know. I'm using Docker 1.10-dev and overlay as the filesystem used by docker (Debian Jessie + Linux 4.4). I'll try to reproduce your issue.

cy8aer:
I'd like to know what happens there: https://forums.docker.com/t/copy-different-with-btrfs/5659

dalf:
`COPY requirements.txt ./requirements.txt` is working whatever the used FS, so why not.

@asciimoo  what do you think : merge ?

Wonderfall:
A friend of mine tried with Docker 1.10.0 + btrfs on Jessie, it worked.
I still don't know why he gets this message... but if it solves his problem, we can consider this.

PR #504, opened 2016-02-04
Simpson screenshot search engine

kvch:
Frinkiac is a Simpson screenshot search engine. More information can be found [here](https://langui.sh/2016/02/02/frinkiac-the-simpsons-screenshot-search-engine/) about the engine.

It is in the images category disabled by default.


asciimoo:
thanks =)

PR #505, opened 2016-02-09
Dockerfile enhancements

Wonderfall:
I made some improvements : 
- Zombie process problem is fixed, more info [here](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/).
- I added some labels and informations about the maintainer which is the project here.

A searx container now stops immediately and securely after using `docker stop`.


asciimoo:
great, thanks

Issue #508, opened 2016-02-12
Wolfram Alpha Integration?


nixoye:
Would it be possible to use the Wolfram Alpha api  so we could type in questions like " what is 20 degrees Fahrenheit in Celsius" and have the searx engine list it like it does for addition or subtraction questions.
http://products.wolframalpha.com/developers/

pointhi:
please look at #486

MarcAbonce:
I just realised that the no API version doesn't work anymore.

Recently, WolframAlpha's front-end was entirely rewritten in Angular. I don't know Angular, but it looks like now it gets all the data asynchronously instead of it being "hardcoded" in the page like before.
So, unless there is a way to get the results without simulating an entire browser, for now you'd have to get an API key, which for free is limited to 2000 queries per month.

For what it's worth, though, ddg_definitions does answer OP's question.

PR #509, opened 2016-02-13
Remove wolframalpha_noapi engine.

MarcAbonce:
Recently, WolframAlpha's front-end was entirely rewritten in Angular, so now all this code is useless.

Related issue: #508


MarcAbonce:
Wow. I hadn't thought of that.
This is great. Thanks.

Issue #428, opened 2015-09-11
Problem with Gigablast results


SomeDudeHere:
Greetings, Searx is capable of getting results from gigablast but when you click on the result, the search engine does not send you there, instead you get a 404 not found for some reason

Thank you

PR #512, opened 2016-02-26
Fix results with no scheme

ldidry:
Related to #428, which was fixed in e3df22b but broken in a refactoring
(b6c3cb0)


asciimoo:
thanks

PR #513, opened 2016-02-28
WolframAlpha infobox

MarcAbonce:
This places wolframalpha results inside an infobox.

It also makes some changes to the infobox styles and templates that were necessary for wolfram's infobox to look nicer. (See 4267b11a45b7427bcba91259fcda202bd049f004)

If you don't like how the table width looks now, I can change the style again so that it only applies when using wolframalpha.


asciimoo:
neat, thanks

PR #516, opened 2016-03-02
Translation fix

pointhi:
I updated the translation file some time ago on transifex, and now I think we can merge the big amount of changes and additions which were made.

There are also some new languages: Bulgarian, Greek (Greece), Esperanto and Português (Brazil)

I also extended the language strings. Now every language has it's writing in it's own language, and the english name of the language in brackets. Also I sorted the language by local (except english) to be more logical.


pointhi:
Also, we have to fix the plugin translation handling. Because we generate translations, but don't display them in the template.

asciimoo:
Cool, nice job!
About the plugin translations: it's easy to fix, just wrap the plugin string variables with `_()` in the template. Could you do it in this pr as well?

asciimoo:
great, thanks =)

PR #517, opened 2016-03-02
[fix] improve argument parsing of manage.sh

pointhi:
`manage.sh` doesn't worked correctly with invalid arguments, so I added some improvements.


asciimoo:
thanks

PR #518, opened 2016-03-03
[mod] simplify searx/search.py

dalf:
[edit] 
It's more to open the discussion than to merge.
I think search.py is more easier to read with the commit.

The call to 
`engine.request(self.query.encode('utf-8'), request_params)`
has been moved in the `search_request_wrapper` function.

See https://github.com/asciimoo/searx/blob/master/searx/engines/wolframalpha_noapi.py#L67  : all requests are suspended waiting for the WA token. With this PR, it's no more a problem.

With a global function, it is easy to create an offline engine.

Notes : 
- engine time out is now about the whole thing (prepare the request, send the request, parse the response). It was more or less the case before since the `threaded_requests` function ignores the thread after the timeout even the HTTP request is ended.
- the response times in the stats are about the HTTP request only. So the avg time can't be compared to the time out in the preferences.

Off course, the draw back is the pycurl branch, but I think a solution can be found.

As I said, it's to open the discussion.

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -88,6 +88,9 @@ def load_engine(engine_data):
         'result_count': 0,
         'search_count': 0,
         'page_load_time': 0,
+        'page_load_count': 0,
+        'engine_time': 0,
+        'engine_time_count': 0,
         'score_count': 0,
         'errors': 0
     }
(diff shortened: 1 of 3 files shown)


asciimoo:
I like it, +1

dalf:
Closed. See PR #751

Issue #506, opened 2016-02-10
Engine > Science > BASE


jibe-b:
BASE (Bielefield Academic Search Engine) is a proxy to search in 4000 open access science publications databases (including ArXiV, BiomedCentral). It harvests available databases through the "Open Archives Initiative Protocol for Metadata Harvesting" (OAI-PMH) and stores its own index. 

Search webpage:
https://base-search.net/

About:
https://www.base-search.net/about/en/help.php

API:
http://api.base-search.net/

(OAI: http://oai.base-search.net/ )

In PDF
http://www.base-search.net/about/download/base_interface.pdf

List of other OAI-PMH harvesters:
https://www.openarchives.org/service/listproviders.html

People searching scientific literature would benefit from having both engines "general" and "science" on the same search webpage.

I will suggest this idea to the people behind BASE.

jibe-b:
@asciimoo, I try to aggregate as much information as possible to make it "let's read the spec, let's implement it" in case there is conscensus in adding this engine (and it could be me who implements it, indeed). What information lack?

By the way, I  sent a message to the BASE team.

asciimoo:
@jibe-b thanks the suggestion. It would be a useful engine and seems, it isn't hard to integrate it to searx. Could you work on it? If you can't, i'll do it =)

jibe-b:
Unfortunately, BASE requires authentication. I will discuss with them this point, even though I already guess possible reasons for locking. But I think there should not be neasons for keeping it locked.

Anyhow, there are other search engines that index Open Archives:
- OpenAire.eu [](openaire.eu), a European Union funded harvester and working group
  [limited API doc](http://api.openaire.eu/#cha_http-api)
- OAISTER by worldcat.org [](oaister.worldcat.org)
  [API doc](www.oclc.org/developer/develop/web-services/worldcat-search-api.en.html)

but BASE seems the most mature.

pietsch:
BASE member of staff here. I like your idea, @jibe-b! We will discuss it in our team meeting tomorrow. There might be a way to authenticate searx instances automagically.

pietsch:
Good news: On behalf of the BASE team, I can confirm that BASE Search is interested in becoming integrated with Searx. We will implement an authentication mechanism that is compatible with Searx's distributed, easy-to-install approach, e.g. by matching the User-Agent HTTP header. Please state your preferences and the next steps to take here.

asciimoo:
[...]

PR #522, opened 2016-03-17
add BASE engine in category "Science" 

jibe-b:
I proposed to add BASE in #506.

This engine performs basic search using the BASE API.

It needs the user-agent in order to match the authorization policy implemented by BASE.

### DIFF
--- a/searx/engines/base.py
+++ b/searx/engines/base.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python
+
+"""
+ BASE (Scholar publications)
+
+ @website     https://base-search.net
+ @provide-api yes with authorization (https://api.base-search.net/)
+
+ @using-api   yes
+ @results     XML
+ @stable      ?
+ @parse       url, title, publishedDate, content
+ More info on api: http://base-search.net/about/download/base_interface.pdf
+"""
+
+from lxml import etree
+from urllib import urlencode
+from searx.utils import searx_useragent
+from cgi import escape
+from datetime import datetime
+import re
+
+
+categories = ['science']
+
+base_url = 'https://api.base-search.net/cgi-bin/BaseHttpSearchInterface.fcgi'\
+           + '?func=PerformSearch&{query}&boost=oa&hits={hits}&offset={offset}'
+
+# engine dependent config
+paging = True
+number_of_results = 10
+
+# shortcuts for advanced search
+shorcut_dict = {
+    # user-friendly keywords
+    'format:': 'dcformat:',
+    'author:': 'dccreator:',
+    'collection:': 'dccollection:',
+    'hdate:': 'dchdate:',
+    'contributor:': 'dccontributor:',
+    'coverage:': 'dccoverage:',
+    'date:': 'dcdate:',
+    'abstract:': 'dcdescription:',
+    'urls:': 'dcidentifier:',
+    'language:': 'dclanguage:',
+    'publisher:': 'dcpublisher:',
+    'relation:': 'dcrelation:',
+    'rights:': 'dcrights:',
+    'source:': 'dcsource:',
+    'subject:': 'dcsubject:',
+    'title:': 'dctitle:',
+    'type:': 'dcdctype:'
+}
+
+
+def request(query, params):
+    # replace shortcuts with API advanced search keywords
+    for key in shorcut_dict.keys():
+        query = re.sub(str(key), str(shorcut_dict[key]), query)
+
+    # basic search
+    offset = (params['pageno'] - 1) * number_of_results
+
+    string_args = dict(query=urlencode({'query': query}),
+                       offset=offset,
+                       hits=number_of_results)
+
+    params['url'] = base_url.format(**string_args)
+
+    params['headers']['User-Agent'] = searx_useragent()
+    return params
+
+
+def response(resp):
+    results = []
+
+    search_results = etree.XML(resp.content)
+
+    for entry in search_results.xpath('./result/doc'):
+        content = "No description available"
+
+        date = datetime.now()  # needed in case no dcdate is available for an item
+        for item in entry:
+            if item.attrib["name"] == "dchdate":
+                harvestDate = item.text
+
+            elif item.attrib["name"] == "dcdate":
+                date = item.text
+
+            elif item.attrib["name"] == "dctitle":
+                title = item.text
+
+            elif item.attrib["name"] == "dclink":
+                url = item.text
+
+            elif item.attrib["name"] == "dcdescription":
+                content = escape(item.text[:300])
+                if len(item.text) > 300:
+                    content += "..."
+
+# dates returned by the BASE API are not several formats
+        publishedDate = None
+        for date_format in ['%Y-%m-%dT%H:%M:%SZ', '%Y-%m-%d', '%Y-%m', '%Y']:
+            try:
+                publishedDate = datetime.strptime(date, date_format)
+                break
+            except:
+                pass
+
+        if publishedDate is not None:
+            res_dict = {'url': url,
+                        'title': title,
+                        'publishedDate': publishedDate,
+                        'content': content}
+        else:
+            res_dict = {'url': url,
+                        'title': title,
+                        'content': content}
+
+        results.append(res_dict)
+
+    return results
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -408,17 +408,21 @@ def index():
 
         # TODO, check if timezone is calculated right
         if 'publishedDate' in result:
-            result['pubdate'] = result['publishedDate'].strftime('%Y-%m-%d %H:%M:%S%z')
-            if result['publishedDate'].replace(tzinfo=None) >= datetime.now() - timedelta(days=1):
-                timedifference = datetime.now() - result['publishedDate'].replace(tzinfo=None)
-                minutes = int((timedifference.seconds / 60) % 60)
-                hours = int(timedifference.seconds / 60 / 60)
-                if hours == 0:
-                    result['publishedDate'] = gettext(u'{minutes} minute(s) ago').format(minutes=minutes)
-                else:
-                    result['publishedDate'] = gettext(u'{hours} hour(s), {minutes} minute(s) ago').format(hours=hours, minutes=minutes)  # noqa
+            try:  # test if publishedDate >= 1900 (datetime module bug)
+                result['pubdate'] = result['publishedDate'].strftime('%Y-%m-%d %H:%M:%S%z')
+            except ValueError:
+                result['publishedDate'] = None
             else:
-                result['publishedDate'] = format_date(result['publishedDate'])
+                if result['publishedDate'].replace(tzinfo=None) >= datetime.now() - timedelta(days=1):
+                    timedifference = datetime.now() - result['publishedDate'].replace(tzinfo=None)
+                    minutes = int((timedifference.seconds / 60) % 60)
+                    hours = int(timedifference.seconds / 60 / 60)
+                    if hours == 0:
+                        result['publishedDate'] = gettext(u'{minutes} minute(s) ago').format(minutes=minutes)
+                    else:
+                        result['publishedDate'] = gettext(u'{hours} hour(s), {minutes} minute(s) ago').format(hours=hours, minutes=minutes)  # noqa
+                else:
+                    result['publishedDate'] = format_date(result['publishedDate'])
 
     if search.request_data.get('format') == 'json':
         return Response(json.dumps({'query': search.query,
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -34,9 +34,15 @@ outgoing: # communication with search engines
 #        - 1.1.1.2
 
 engines:
+<<<<<<< HEAD
   - name : arch linux wiki
     engine : archlinux
     shortcut : al
+=======
+  - name : base
+    engine : base
+    shortcut : bs
+>>>>>>> Add BASE engine in category Science. Basic and advanced request.
 
   - name : wikipedia
     engine : mediawiki
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -34,15 +34,13 @@ outgoing: # communication with search engines
 #        - 1.1.1.2
 
 engines:
-<<<<<<< HEAD
   - name : arch linux wiki
     engine : archlinux
     shortcut : al
-=======
+
   - name : base
     engine : base
     shortcut : bs
->>>>>>> Add BASE engine in category Science. Basic and advanced request.
 
   - name : wikipedia
     engine : mediawiki


searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+
+# engine dependent config
+paging = True
+number_of_results = 10
+
+
+def request(query, params):
+
+    offset = (params['pageno'] - 1  ) * number_of_results
+    print (params['pageno'] -1)* number_of_results
```
asciimoo: please delete this debug print

searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+
+            elif item.attrib["name"] == "dclink":
+                url = item.text
+
+            elif item.attrib["name"] == "dcdescription":
+                content = escape(item.text[:300]) + "..."
+
+        #tmp: dates returned by the BASE API are not in iso format
+        #so the three main date formats are tried one after the other
+        try:
```
asciimoo: this nested exception handling should be rewritten to something like this:

``` python
for date_format in range ["formatstr1", "formatstr2", ..]:
    try:
        publishedDate = datetime.strptime(date, date_format)
        break
    except:
        pass
```

searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+                try:
+                    publishedDate = datetime.strptime(date, '%Y-%m' )
+                except:
+                    try:
+                        publishedDate = datetime.strptime(date, '%Y' )
+                    except:
+                        try:
+                            publishedDate = datetime.strptime(harvestDate,  '%Y-%m-%dT%H:%M:%SZ' )
+                        except:
+                            publishedDate = datetime.now()
```
asciimoo: Do we really want to set the current date if we cannot parse the original one?

searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+                date = item.text
+
+            if item.attrib["name"] == "dctitle":
+                title = item.text        
+
+            elif item.attrib["name"] == "dclink":
+                url = item.text
+
+            elif item.attrib["name"] == "dcdescription":
+                content = escape(item.text[:300]) + "..."
```
asciimoo: it appends `"..."` even if the content is shorter than 300 chars

searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+def request(query, params):
+
+    offset = (params['pageno'] - 1  ) * number_of_results
+    print (params['pageno'] -1)* number_of_results 
+
+    string_args = dict(query=urlencode({'query': query}),
+                       offset=offset,
+                       hits=number_of_results)
+
+    format_strings = list(Formatter().parse(base_url))
```
asciimoo: what is the purpose of the above line?

searx/engines/base.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+ BASE (Scholar publications)
+
+ @website     https://base-search.net
+ @provide-api yes with authorization (https://api.base-search.net/)
+
+ @using-api   yes
+ @results     XML
+ @stable      ?
+ @parse       ?
```
kvch: Values of parse must be the keys of the results dict. In this case it is url, title, publishedDate, content.

searx/engines/base.py
jibe-b: done in next commit

searx/engines/base.py
jibe-b: You're right, as Searx does not need a date.

still work to be done.

searx/engines/base.py
jibe-b: done in next commit

searx/engines/base.py
jibe-b: copy-pasting side effect ;)
deleted in next commit

searx/engines/base.py
jibe-b: ok.
publishedDate is optional in next commit. How to declare in parse?

searx/engines/base.py
```diff
@@ -0,0 +1,121 @@
...
+        #tmp: dates returned by the BASE API are not in iso format
+        #so the three main date formats are tried one after the other
+        for date_format in ['%Y-%m-%dT%H:%M:%SZ', '%Y-%m-%d', '%Y-%m', '%Y']:
+            try:
+        	publishedDate = datetime.strptime(date, date_format)
+        	break
+    	    except:
+                pass
+
+	if publishedDate != None:
```
kvch: This format of None-checking does not comply with PEP8. It should be `if publishedDate is not None:`. For further information on that, please check [programming recommendations section of PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

jibe-b: done in next commit.

searx/engines/base.py
```diff
@@ -0,0 +1,122 @@
...
+            elif item.attrib["name"] == "dcdescription":
+                content = escape(item.text[:300])
+                if len(item.text) > 300:
+                    content += "..." 
+
+        #tmp: dates returned by the BASE API are not in iso format
+        #so the three main date formats are tried one after the other
+        for date_format in ['%Y-%m-%dT%H:%M:%SZ', '%Y-%m-%d', '%Y-%m', '%Y']:
+            try:
+        	publishedDate = datetime.strptime(date, date_format)
```
kvch: publishedDate must be initialized before checking if it is none. If an exception is thrown from the try block, it remains uninitialized, so the None-checking fails and the engine craches.

jibe-b: Ok. done in next commit.

Thanks for explaining in detail.

asciimoo:
@jibe-b there was an error in the travis configuration, but from now it shows coding style issues properly. Please check the output (https://travis-ci.org/asciimoo/searx/builds/116728852#L683) and fix the issues.

jibe-b:
@asciimoo I cleaned my code.

[Last build](https://travis-ci.org/asciimoo/searx/builds/116762265) failed with

> The command "./manage.sh pep8_check" exited with 1.
> 
> 1.90s$ ./manage.sh styles
> 
> [!] Building styles
> 
> The compress option has been deprecated. We recommend you use a dedicated css minifier, for instance see less-plugin-clean-css.

which is related to older code.

kvch:
@jibe-b The failure is not caused by older code. The output you pasted here belongs to the manage.sh styles command. It is just a warning, so that command does not fail.

The PEP8 check still fails because of your code. There is only one remaining error to fix. It can be found in the build log above the first line you pasted. Please, check again and fix it.

jibe-b:
Thanks for explaining in detail, you're right I didn't read the build log carefully enough. I am learning using travis while doing.

I fixed this line.

searx/engines/base.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for entry in search_results.xpath('./result/doc'):
+        content = "No description available"
+
+        date = datetime.now()  # needed in case no dcdate is available for an item
+        for item in entry:
+            if item.attrib["name"] == "dchdate":
+                harvestDate = item.text
+
+            if item.attrib["name"] == "dcdate":
```
kvch: In this loop you are checking the value of only the `name` key in all of the `if`s. Using `elif` would be much better because this code would be skipped if one of the conditions were true. The same applies to the next `if` below this one. In the last two cases using 'elif` is the appropiate choice. :)

searx/engines/base.py
```diff
@@ -0,0 +1,121 @@
...
+        'language': 'dclanguage',
+        'publisher': 'dcpublisher',
+        'relation': 'dcrelation',
+        'rights': 'dcrights',
+        'source': 'dcsource',
+        'subject': 'dcsubject',
+        'title': 'dctitle',
+        'type': 'dcdctype'
+    }
+# replace shortcuts with API advanced search keywords
```
kvch: Please, indent the comments just like the code.

searx/engines/base.py
```diff
@@ -0,0 +1,121 @@
...
+           + '?func=PerformSearch&{query}&boost=oa&hits={hits}&offset={offset}'
+
+# engine dependent config
+paging = True
+number_of_results = 10
+
+
+def request(query, params):
+    # shortcuts for advanced search
+    shorcut_dict = {
```
kvch: Could you write new shortcuts for the advanced search? These words could cause problems. For example if I want to searcch for the term `cat title` the query would be transformet to `cat dctitle`. That is not what I wanted to search for.

I checked the shortcuts at BASE. It seems to me that it has completelydifferent advanced search syntax. For instance if you want to search for something that is titled French, you would enter `tit:French`. To not to confuse experienced BASE users, I suppose the shortcut keys should be the same as at BASE.

kvch: Furthermore, you should move this definition out of this fuction. This dict is defined everytime someone makes a request. If you move it outside where the `base_url` is found, it would be only initialized at startup. This change would spare us time. :)

searx/engines/base.py
jibe-b: elif done in next commit.

searx/engines/base.py
jibe-b: done in next commit.

searx/engines/base.py
jibe-b: @kvch This goes further than what I propose to add to Searx, so I will continue in a new branch [base.py](https://github.com/jibe-b/searx/blob/advanced-search/searx/engines/base.py) in this [issue](https://github.com/jibe-b/searx/issues/1).

kvch: @jibe-b what do you mean by this going further than you proposed to add to searx? I think it is enough to change the keys in the dict to `tit:`, `aut:` and so on like described in the Limit search area section of BASE help. Extra API params could be parameterized using keywords you/we come up with. Or maybe we do not need all of the params provided by the BASE API.

What do you think @asciimoo?

asciimoo: It would be good to behave similar to the original sites query syntax, even if it's just a subset of the available options.

jibe-b: @kvch Just didn't want to be messy doing several things at a time.

So when we agree on [base.py](https://github.com/jibe-b/searx/blob/advanced-search/searx/engines/base.py), I will merge it into this pull request

searx/engines/base.py
```diff
@@ -0,0 +1,128 @@
...
+
+            elif item.attrib["name"] == "dclink":
+                url = item.text
+
+            elif item.attrib["name"] == "dcdescription":
+                content = escape(item.text[:300])
+                if len(item.text) > 300:
+                    content += "..."
+
+# dates returned by the BASE API are not several formats
```
kvch, line 101: This indentation is still wrong.

jibe-b, line 101: done in next commit.

kvch:
I've found a bug while testing the engine locally. When I search for author:Kate, the engine crashes. The reason is quite funny. The error message is the following:

`ValueError: year=1847 is before 1900; the datetime strftime() methods require year >= 1900`

So it means that something has to be done about dates before 1900. I would say that for now, when ValueError is raised publishDate must remain None and the loop should be exited.

jibe-b:
It seems to be an old bug in the datetime module that can be solved either in altering datetime () 

http://stackoverflow.com/questions/10263956/use-datetime-strftime-on-years-before-1900-require-year-1900/32206673#32206673

strftime() is used in searx/webapp.py as well so I agree the exception handling. When webapp.py will support dates earlier than 1900, be it by using another module or by patching datetime.strftime, I will do the same in the base.py engine.

jibe-b:
@kvch is it better to make the base.py return publisedDate only when greater than 1900 or to add the exception handling in webapp.py?

The first option limits changes but the second prevents this from happening with other engines.

kvch:
@jibe-b It is a great idea to add the exception handling to webapp.py. You should put it there. :) It could save us from a few engine crashes. :)

jibe-b:
ok, I'll do it. Just let me time to understand how webapp.py works.

jibe-b:
@kvch what should publishedDate take as a value in case the exception is raised?

kvch:
It should take None.

searx/webapp.py
```diff
@@ -406,17 +406,20 @@ def index():
...
         if 'publishedDate' in result:
-            result['pubdate'] = result['publishedDate'].strftime('%Y-%m-%d %H:%M:%S%z')
-            if result['publishedDate'].replace(tzinfo=None) >= datetime.now() - timedelta(days=1):
-                timedifference = datetime.now() - result['publishedDate'].replace(tzinfo=None)
-                minutes = int((timedifference.seconds / 60) % 60)
-                hours = int(timedifference.seconds / 60 / 60)
-                if hours == 0:
-                    result['publishedDate'] = gettext(u'{minutes} minute(s) ago').format(minutes=minutes)
+            try:  # test if publishedDate >= 1900 (datetime module bug)
+                result['pubdate'] = result['publishedDate'].strftime('%Y-%m-%d %H:%M:%S%z')
```
kvch, line 412: It seems to me that only this line could throw the ValueError that should be caught. The lines above this should be put inside an else clause of the try statement.

jibe-b:
@kvch I have been using the engine during the last week for basic search, but I know the queries I made were in the bubble of my current work. Are there manual tests I should consider doing?

kvch:
@jibe-b That should suffice. I also tested the engine. I had no problem with it apart from that ValueError.

I think @asciimoo going to review your code soon, too. In the mean time could you fixup your patches into 1-2 commits? 42 commits is a lot for an engine. :)

jibe-b:
@kvch that's right, this time it was 42 steps, probably less next time ;)

I guess I make a new pull request?

kvch:
@jibe-b  No, you should not make a new pull request. You should rebase your branch and force push it to github.

kvch:
@jibe-b And of course it is going to be less next time. :) You are welcome to add engines in the future. :)

asciimoo:
@jibe-b i have a small note on the code, please fix it and if you squash your commits it's ready to pull =)
thanks

jibe-b:
sorry @asciimoo, which note?

searx/engines/base.py
```diff
@@ -0,0 +1,128 @@
...
+            if item.attrib["name"] == "dchdate":
+                harvestDate = item.text
+
+            elif item.attrib["name"] == "dcdate":
+                date = item.text
+
+            elif item.attrib["name"] == "dctitle":
+                title = item.text
+
+            elif item.attrib["name"] == "dcdate":
```
asciimoo: seems these two lines are the same as 87-88

asciimoo:
@jibe-b sry, i forgot to send m(, see above

jibe-b:
Dear @asciimoo and @kvch it's ready!

it's rebased, commits squashed into two commits (though there is too much in the commit secondary text as I don't master the rebase command yet).

asciimoo:
@jibe-b, @kvch, thanks your hard work =)

jibe-b:
Thanks @asciimoo! I learned while doing this :)

jibe-b:
@asciimoo How do you proceed in the diffusion of information about Searx?

The BASE engine may interest people in academia and I want to spread the news in the circles I know (young researchers in France).

Maybe is it better to wait until a new release so that existing instances integrate BASE.

@pietsch, would you participate in advertising to academics "a (meta)-search engine with both normal search tab and scholar search tab"?

pietsch:
I am traveling this week. After that, I would love to chime in. I'll also do my best to persuade BASE, but they usually rely on word of mouth instead of PR.

jibe-b:
@pietsch, I consider communication within a month. And I will discuss with others about it on the HackYourPhD fb group: https://www.facebook.com/groups/499463776745202/ (and this PR can be closed)

Issue #527, opened 2016-03-29
crash on specific search [BUG]


prolibre:
Online exemple : https://www.heraut.eu/search/
go to **news** and search "_yahoo lent_"

Result : Internal Server Error

Logs:

> ERROR:searx.search:engine crash: google news
> Traceback (most recent call last):
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/searx-0.8.1-py2.7.egg/searx/search.py", line 40, in search_request_wrapper
>     ret = fn(url, *_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/searx-0.8.1-py2.7.egg/searx/poolrequests.py", line 82, in get
>     return request('get', url, *_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/searx-0.8.1-py2.7.egg/searx/poolrequests.py", line 75, in request
>     response = session.request(method=method, url=url, *_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
>     resp = self.send(prep, *_send_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/sessions.py", line 582, in send
>     r = dispatch_hook('response', hooks, r, *_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/requests/hooks.py", line 31, in dispatch_hook
>     _hook_data = hook(hook_data, *_kwargs)
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/searx-0.8.1-py2.7.egg/searx/search.py", line 117, in process_callback
>     search_results = callback(response)
>   File "/usr/local/searx/searx-ve/lib/python2.7/site-packages/searx-0.8.1-py2.7.egg/searx/engines/google_news.py", line 50, in response
>     if not search_res.get('responseData', {}).get('results'):
> AttributeError: 'NoneType' object has no attribute 'get'
> ERROR:searx.webapp:Exception on / [POST]
> Traceback (most recent call last):
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
>     response = self.full_dispatch_request()
>   File "/usr/local/searx/searx-ve/local/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
[...]

PR #530, opened 2016-03-30
Check for 'too many requests per second' error in Google News engine

ghost:
There were two issues reported in the original message in #527, but only one of them has been fixed. One small issue remains: Google News has a limit on how many requests can be made in a second and fails if this limit is exceeded, which leads to this nasty error message: http://i.imgur.com/OCF5AIc.png

When this happens, the `responseData` variable is equal to `None`. This patch adds an additional check on the data returned in the server's response and returns an empty result if we're being temporarily blocked by the service.

### DIFF
--- a/searx/engines/google_news.py
+++ b/searx/engines/google_news.py
@@ -45,9 +45,10 @@ def response(resp):
     results = []
 
     search_res = loads(resp.text)
+    response_data = search_res.get('responseData', {})
 
     # return empty array if there are no results
-    if not search_res.get('responseData', {}).get('results'):
+    if not response_data or not response_data.get('results'):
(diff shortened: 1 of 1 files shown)


asciimoo:
I agree that the error message is ugly, but it's a good practice to raise an exception inside an engine if something went wrong. Searx has an engine suspend mechanism which checks engine errors and suspends misbehaving engines for a while, what can be useful in this case.
A possible solution can be a dedicated exception type where no traceback displayed.

Also, i'm not sure if google news engine is working at all, but it needs more investigation

ghost:
Thanks for the information, this is not the best solution then.

PR #531, opened 2016-04-03
Add doku engine

guyou:
Dokuwiki is a widely used wiki engine: it could be used publicly (Ubuntu-fr) or in corporate environment.

So, here is an (initial) implementation for adding dokuwiki support in searx.


asciimoo:
@gbonnefille nice
Please remove ubuntu-fr from the settings, because it's not a general service - or comment it out and it can still demonstrate the usage of the engine.

asciimoo:
great, thanks

PR #534, opened 2016-04-08
new preferences handling

kvch:
Preferences class was introduced in order to handle user preferences. Right now
it parses cookies and the form in preferences. Also it can retrieve settings
based on the name of the setting.

**ATTENTION**
**Please note that engine preferences are handled differently from now on. So it
introduces incompatible changes. Every user who has saved preferences should reset and
save his/her settings again.**

This change was needed, because everytime a default disabled engine was
added saved user preferences would broke. Now engine setting tracking is
fixed.

Changes:
- new Preferences class
- followed up changes in `webapp.py` & `search.py`
- new unittests
- new mocked function in robot tests
- fixed `settings_robot.yml`
- new robot tests from prevous PR #427

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -1,0 +1,269 @@
+from searx import settings, autocomplete
+from searx.languages import language_codes as languages
+
+
+COOKIE_MAX_AGE = 60 * 60 * 24 * 365 * 5  # 5 years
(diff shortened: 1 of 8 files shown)


asciimoo:
great, thanks =)

PR #536, opened 2016-04-09
General minor refactors

kvch:
Most of the changes are minor, invisible stuff. The last two commits are in connection with the preferences refactor. `blocked_engines` variables were renamed to `disabled_engines` in order to be consistent with the cookie nomenclature. Unfortunately, I forgot to use the new preferences was instead of accessing cookies in the last PR. But now I added a commit which contains the missing follow ups.


asciimoo:
thanks

Issue #447, opened 2015-10-13
Infobox is always in English


GreenLunar:
A search for [שלום](https://searx.me/?q=%D7%A9%D7%9C%D7%95%D7%9D) (hello/peace) or  [שחיתות](https://searx.me/?q=%D7%A9%D7%97%D7%99%D7%AA%D7%95%D7%AA) (corruption), returns information in English in Searx Infobox.  Sometimes, the InfoBox returns nothing; [הגשש החיוור](https://searx.me/?q=%D7%94%D7%92%D7%A9%D7%A9%20%D7%94%D7%97%D7%99%D7%95%D7%95%D7%A8) (HaGashash HaHiver).

_Users who use Searx in languages that are not English, would expect or prefer to see text of Infobox with their respective language._

GreenLunar:
The above search term is not a good example.  Try this https://searx.me/?q=%D7%94%D7%92%D7%A9%D7%A9%20%D7%94%D7%97%D7%99%D7%95%D7%95%D7%A8 (no InfoBox)

GreenLunar:
Correction, it is mostly and only in English https://searx.me/?q=%D7%A9%D7%97%D7%99%D7%AA%D7%95%D7%AA

שחיתות = corruption
IfoBox shows whistleblower (in English)

MarcAbonce:
I'm currently working on this one here: https://github.com/a01200356/searx/commits/wikipedia_infobox

It cuts out the middleman by scraping wikipedia directly because I couldn't find in duckduckgo's api's documentation how to set the language (I don't even know if they support it at all).

It still needs testing though.

asciimoo:
fixed by @a01200356 , thanks

GreenLunar:
I don't see any difference.

asciimoo:
@GreenLunar it works for me: https://searx.me/?q=%21wp%20%3Ahu%20k%C3%A1poszta&categories=none - it is hungarian

GreenLunar:
It appears not to work when `!wp :hu` is not used https://searx.me/?q=k%C3%A1poszta&categories=none

dalf:
The data seems to come from the wikidata engine.

Issue #446, opened 2015-10-13
[BiDi] Text direction of Infobox


GreenLunar:
Text direction inside the Infobox of searx side panel, should explicitly be set by its language.  If the language from which sub-domain or directory the information is taken from, is RTL, set text direction to RTL, otherwise set text direction to LTR.

Open this link https://searx.me/?q=hello with searx in RTL (e.g. Hebrew locale)
Also try this link https://searx.me/?q=%D7%A9%D7%9C%D7%95%D7%9D

`class="panel panel-default infobox"`

GreenLunar:
It appears to be working.  Shall we close this issue?

PR #540, opened 2016-04-15
[enh] wikipedia infobox

MarcAbonce:
creates simple multilingual infobox using wikipedia's api

extracting the opening paragraph is the only tricky part because the api doesn't provide that directly.

also modified:
- duckduckgo definitions also accpets more languages
- disabled duckduckgo definitions engine (for supported languages it's still better though)
- wikidata: return local link as infobox id (to make it mergeable with wikipedia's)
- results: fixes in merge_two_infoboxes
- template: print bidirectional text


MarcAbonce:
related to #447 and maybe #446

dalf:
Sorry I wake up late: the duckduckgo definition engine can produce results in other languages, at least French (using the `kl` argument if I remember correctly).

But there is an problem : the [current implementation](https://github.com/asciimoo/searx/blob/master/searx/results.py#L126) uses an `id` which is the [the English wikipedia link](https://github.com/asciimoo/searx/blob/master/searx/engines/duckduckgo_definitions.py#L94) (see also https://github.com/asciimoo/searx/blob/master/searx/engines/wikidata.py#L191 ). So if the duckduckgo definition engine returns no English results, how to merge the infoboxes ?

MarcAbonce:
Actually, it looks like Duckduckgo Definitions takes the language from the Accept-Language http header. The kl argument did nothing to me.
But it still doesn't work for all languages that Wikipedia supports. Chinese, Arabic and Hebrew don't return anything.

To merge the infoboxes, Wikidata now returns the id in the user's language.

MarcAbonce:
Ok, now I also modified duckduckgo_definitions to accept more languages.
The infoboxes merge correctly as far as I've seen.

searx/engines/duckduckgo_definitions.py
```diff
@@ -103,6 +104,10 @@ def response(resp):
         urls.append({'title': search_res.get('DefinitionSource'),
                     'url': definitionURL})
 
+    # to merge with wikidata's infobox
+    if infobox_id:
+        infobox_id = sub(r'^http:', r'https:', infobox_id)
```
asciimoo: Compiled regex would be more efficient

searx/engines/duckduckgo_definitions.py
```diff
@@ -103,6 +106,10 @@ def response(resp):
         urls.append({'title': search_res.get('DefinitionSource'),
                     'url': definitionURL})
 
+    # to merge with wikidata's infobox
+    if infobox_id:
+        infobox_id = sub(http_regex, 'https:', infobox_id)
```
asciimoo: thanks the fix,
this could be also `http_regex.sub('https:', infobox_id)`

asciimoo:
great, thanks

PR #542, opened 2016-04-16
[fix] incorrect URLs in Reddit image search results - closes #538

### LINKED ISSUE #538: reddit: cannot view image
- search: !re wallpaper dark blue
- select some image
- click "View Image" button

Then the image should be shown but the searx start page is loaded. "View Page" shows the image. This is because image==page but this is an inconsistency (sort of)

ghost:
Thanks to @cy8aer for reporting the issue #538. This should fix it. A few examples:

№1:
Image link: http://i.imgur.com/071gi6k.png
Post link: http://i.imgur.com/M3FZFqO.png

№2:
Image link: http://i.imgur.com/CowKV7o.png
Post link: http://i.imgur.com/4N2k5vP.png

Note: it's not always possible to get a direct link to the image (because Reddit returns the link just as user has posted it, without any modifications), and there is no uniform method to extract one, which would work on any site, so sometimes by clicking the "Get image" button you'll be redirected to a page like this: https://imgur.com/CowKV7o , or this: https://www.flickr.com/photos/jurw/7296760374/

### DIFF
--- a/searx/engines/reddit.py
+++ b/searx/engines/reddit.py
@@ -13,7 +13,7 @@
 import json
 from cgi import escape
 from urllib import urlencode
-from urlparse import urlparse
+from urlparse import urlparse, urljoin
 from datetime import datetime
 
 # engine dependent config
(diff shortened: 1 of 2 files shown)


asciimoo:
thanks

Issue #539, opened 2016-04-14
problems redirecting from / to /searx using reverse proxy


IguanaMeow:
**Hello,**

**I have searx and installed via docker, but I'm having problems setting up searx with https as https://domain.tld/searx via reverse proxy on nginx.**
**Searx works as for insance as http://domain.tld:9999 and as https://domain.tld**

**I've tested and searx works as http://domain.tld:9999**
docker run -d --name searx -p 9999:9999 -e IMAGE_PROXY=True searx

**Also it works as https://domain.tld using proxy**
docker run -d --name searx -p 9999:9999 -e IMAGE_PROXY=True searx
**Nginx reverse proxy config:**
location / {
        proxy_pass             http://localhost:9999/;
        proxy_pass_header      Server;
        proxy_set_header       Host                 $host;
        proxy_set_header       X-Real-IP            $remote_addr;
        proxy_set_header       X-Forwarded-For      $proxy_add_x_forwarded_for;
        proxy_set_header       X-Remote-Port        $remote_port;
        proxy_set_header       X-Forwarded-Proto    $scheme;
        proxy_buffering        off;
    }

**Now I just have a problem with correctly getting searx to work as https://domin.tld/searx
I have tried adding these extra configs to the proxy, but no luck in displaying it correctly as the last two methods above**
location /searx {
        rewrite                /searx/(.*) /$1 break;
        rewrite                ^/searx$ /searx/ permanent;
        proxy_redirect         / /searx;
    }

**Also I have done this with docker:**
docker run -d --name searx -e IMAGE_PROXY=True -e BASE_URL=https://localhost/searx searx
docker run -d --name searx -p 9999:9999 -e IMAGE_PROXY=True -e BASE_URL=https://localhost/searx searx

**But it seems it doesnt make a difference.**

**Any ideas?**

IguanaMeow:
I have solved it with the help of asciimoo on IRC using this method:
https://github.com/asciimoo/searx/blob/master/searx/webapp.py#L697

IguanaMeow:
location /searx {
        proxy_pass             http://127.0.0.1:9999;
        proxy_set_header       Host $host;
        proxy_set_header       X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header       X-Scheme $scheme;
        proxy_set_header       X-Script-Name /searx;
        proxy_buffering        off;
}

asciimoo:
The documentation should be updated

asciimoo:
[...]

PR #544, opened 2016-04-21
install docs update && dev tips

kvch:
- install docs updated with configuration based on #539, please, check if the lines are correct. :)
- add small tips to dev quickstart


asciimoo:
thanks

PR #545, opened 2016-04-21
Oscar improvement

pointhi:
Add support for `img_src` in default results of the oscar theme. (the fdroid engine for example require it)


asciimoo:
thanks

PR #547, opened 2016-04-24
Vim-inspired hotkeys plugin

ghost:
Hi,

Would you be interested in a plugin which adds Vim-like navigation to searx? Some very basic Vim-like shortcuts are available on many sites (for example, j/k work here: https://github.com/asciimoo/searx/issues), so I thought, why not.

Basically it's a lightweight analogue to browser plugins like Vimperator for those who aren't comfortable with using them on all sites all the time (like myself).

Here is a short screencast (3 MiB) I made which shows most of the supported functions:

https://dl.dropboxusercontent.com/u/18321926/cast.mkv

I plan to make the shortcuts configurable when I have more free time.

Any suggestions?

Cheers!

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -23,7 +23,8 @@ from searx.plugins import (https_rewrite,
                            open_results_on_new_tab,
                            self_info,
                            search_on_category_select,
-                           tracker_url_remover)
+                           tracker_url_remover,
+                           vim_hotkeys)
 
 required_attrs = (('name', str),
                   ('description', str),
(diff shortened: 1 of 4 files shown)


asciimoo:
neat, thanks

GreenLunar:
> https://dl.dropboxusercontent.com/u/18321926/cast.mkv

What is the screen recording software you used to compose this screencast with indication of given pressed keys?

By the way, for being so productive, I think @ukwt should be joined as a member of searx organization.

ghost:
@GreenLunar, it's a combination of two things: `vokoscreen` (https://github.com/vkohaupt/vokoscreen) and `key-mon` (https://code.google.com/archive/p/key-mon/).

`vokoscreen` can show keys too, but in a form not suitable to demonstrate vim-like navigation: https://i.imgur.com/2qFmTaq.png

GreenLunar:
Thank you for the information.

@danielgtaylor, why did not you upload key-mon, key-train and the rest of your projects there to your domain here? << I will delete this line later because it is absolutely off topic.

PR #549, opened 2016-04-28
Use new settings to fine tune the connection pool

guyou:
In some circumstances, it is necessary to increase size of pool.

For example, in my corporate environment, I have to increase the pool's maxsize to 150 in order to avoid errors.


kvch:
Nice. :)
Please add the new options and its short description to `settings.yml`. Also please fix ypur code, because the PEP8 check fails.

kvch:
Oh, sorry I haven't noticed you pushed the fixes. :) I think it can be merged. @asciimoo what do you think?

asciimoo:
sorry for the delay and thanks, it's a useful feature to finetue searx

PR #551, opened 2016-05-02
use ngettext for plurals

mikhirev:
Hi!
This adds proper plurals handling with ngettext().


mikhirev:
I also found that Plural-Forms header in Russian .po file was wrong.

mikhirev:
Hi! Are there any problems with this PR? The fix is trivial, but it requires changes in translations. Sooner it will be merged and TX project synced, more time translators will have to fix translations.

asciimoo:
sorry for the delayed answer, it's a nice improvement, thanks

pointhi:
splitting `'{minutes} minute(s) ago'` was a very bad idea, because in German (and likely also other languages, you are not able to create meaningfull translations). For example (translated to english), It would be something like `ago 5 hours, 10 minutes` or `5 hours, ago 10 minutes`.

asciimoo:
@pointhi is there any better solution than reverting?

mikhirev:
Oops, I'm sorry. There are two possible solutions: use [contexts](https://www.gnu.org/software/gettext/manual/gettext.html#Contexts) or simply split the line in another place, e.g. before the comma. The first is better IMHO.

asciimoo:
to not introduce new bugs in the release, reverting and leaving this issue to the next release

mikhirev:
Please re-apply commit 242c9ba, it fixes another issue and does not break anything else. I'll fix ngettext/pngettext related stuff later.

BTW Russian translation looks like a crap now. An I guess that not only Russian, but all languages that have more than 2 plural forms.

PR #552, opened 2016-05-02
Plugin to generate direct links to media files (closes #537)

### LINKED ISSUE #537: [plugin] download video
Offer a plugin that, in addition to show/hide video, would provide a selection of available qualities of a given video to be downloaded.

ghost:
Hi,

This adds a new feature allowing user to download videos in various formats from all websites supported by youtube-dl. Links are generated by youtube-dl on the server side and passed to the client, which builds a table like this: http://i.imgur.com/oc6KDK7.png. Click on any link you are interested in and the download starts.

Thanks to @GreenLunar for the idea and to @pointhi & @GreenLunar for valuable feedback and suggestions.

This service is protected against abuse by the same mechanism as the image proxy (thanks to @pointhi for pointing that out). You can sort the table by any column and search for stuff in it.

Here is how it works: https://dl.dropboxusercontent.com/u/18321926/cast3.mkv

If you have any suggestions or concerns I'll address them and fix the code asap.

Cheers.

---

This closes #537.


asciimoo:
This is a great feature. It would be good not to limit this to users who use javascript.
IMHO the basic functionality (popup the info) could be javascript free.

ghost:
@asciimoo, this is a good idea and I'll certainly add this functionality, but I had to abandon my old account and no longer have access to this pull request (nor to its code), so it'll have to wait until the code is merged into `master`. Sorry.

dalf:
@asciimoo : How do you want to make it javascript free ? It's require to download the metainformation for each videos which will slow down the response time a lot each, even with parallel downloads.

Perhaps you ask for `/video_links` to returns HTML instead of JSON ? But that's Javascript that will load the content anyway (except if IFrame are used ?)

Or perhaps I misunderstood your comment.

Note : the PR can't be merged because of oscar.min.css, searx.min.js files.
Another PR can be created from https://patch-diff.githubusercontent.com/raw/asciimoo/searx/pull/552.diff

asciimoo:
@dalf it could work similar as the image proxy. If we don't want to stream the whole video through searx by default we can create an endpoint which accepts the video page url (+hash) and queries the metadata and redirects the client to the direct video url.

dalf:
In the current implementation, the video is not downloaded from searx, youtube-dl gets the URLs nothing more  ( https://github.com/asciimoo/searx/pull/552/files#diff-36c9ae084b973041809e31fa6cc693d1R82 ).

Proxy the download could be great, but is flask the best framework to do that ? One download will block the one process for a very long time. Twisted ? Go ? I think it's related to the resize image proxy.

asciimoo:
@dalf so my proposed solution is partially done =] just the webapp needed to be modified to respond with a HTTP redirect to the video download url: https://github.com/asciimoo/searx/pull/552/files#diff-02503eba222466e11a3f7ae36debdc5cR710 - line 710

dalf:
I think it's return a list of URLs / videos : one URL for each different format.

asciimoo:
Yap but it can be rewritten to return with a default format. If it isn't enough, a preferences option could be added where everybody can set their own preferred formats.

dalf:
I think the format can be different each time / search. Sometimes to get the HD version, sometimes to get only the music, sometimes to get the low quality version. 
Wouldn't be easier to have a global option in advanced settings, with a default setting from the cookies / preferences? But even in this case how to deal with videos that are not encoded in the preferred format?

kvch:
I  don't see why preferred format would change. I mean I would always download specific quality of videos and audios. It would only change if I changed the computer I use. Could anyone provide a reasonable usecase when someone would want to download a different format as set in his/her preferences? If there is, I agree with @dalf on the advanced setting option.

Maybe a warning should be shown when the video is not in the preferred format? And/or users should be asked if the available format should be downloaded?

return42:
It looks like no one is working on this solution anymore. I'll close it for now .

PR #559, opened 2016-05-12
Oscar theme redesign

janoamaral:
Full redesing of the logo and theme Oscar.

**Home page**
searx-home

**Default results**
searx-result-default

**Code results** 
searx-result-code

**Image results**
searx-result-images

**File results**
searx-result-files

**Maps results**
searx-result-maps

**Social media results**
searx-result-media

**Music results**
searx-result-music

**News results**
searx-result-news

**Video results**
searx-result-video


MarcAbonce:
When I run it, it automatically enables all the engines. I tried resetting my cookies to see if it was an error on my settings, but the preferences changed again.

dalf:
Very nice, should it be another theme or update the current one ? @asciimoo , @pointhi  ?

Some comments : 
- there are many modified mod bits from 100644 to 100755
- the use of `fonts.googleapis.com` and `maxcdn.bootstrapcdn.com` create a privacy issue. Note : `fonts.googleapis.com` serves different content according to the user-agent, see http://stackoverflow.com/questions/25011533/google-font-api-uses-browser-detection-how-to-get-all-font-variations-for-font
- the PR seems to be based on an old version of the code. For instance, the commit 8d335dbdaedd6113242e785e8fabac86128d069a modifies searx/templates/oscar/infobox.html to add the `bdi` HTML element, but this PR removes that element. Same the commit d27f7a1b071b4796e4716951a8e8e62a11dc99c3 seems to be cancelled by this PR. The error described by @a01200356 may be related to this.

PwnArt1st:
using safari, the logo that is meant to be at the top left had expanded to cover about a quarter of the page, sitting in the background behind results and the search bar. i merged your theme into my fork but had the revert the custom logo to the original

ibrokemypie:
seems to be overly shortening the result previews under general, makes it impossible to use it effectively

Cqoicebordel:
From my POV, it has enough differences to be another theme altogether. Could be the default one though.

janoamaral:
@PwnArt1st Can't replicate the bug, version of safari?
screen shot 2016-05-15 at 10 08 31 pm

MarcAbonce:
ok, i realised where the bug comes from.
apparently, blocked_engines was changed to disabled_engines in the cookies (8c4db08443c7a55344fa0011b89303467fa62efe).
in your preferences template, the engines' checkbox_toggles check for the old variable.

ibrokemypie:
@logico-dev Version 9.1 (11601.5.17.1)

asciimoo:
Compelling!

My 2 cents: the results are too loose to me. I prefer compact result pages where i can see 8-10 results without scrolling. Currently only the "default" (we should rename it) theme compact enough.

As i see it doesn't modify the structure of the theme and i don't prefer a new theme with lots of duplicated template code. Maybe a better solution would be a selectable CSS styling for oscar - like in courgette theme.

Cqoicebordel:
> My 2 cents: the results are too loose to me. I prefer compact result pages where i can see 8-10 results without scrolling. Currently only the "default" (we should rename it) theme compact enough.

Pix-art is really compact too ;)
Agreed on the renaming of default, since it's no longer the default.

> As i see it doesn't modify the structure of the theme and i don't prefer a new theme with lots of duplicated template code. Maybe a better solution would be a selectable CSS styling for oscar - like in courgette theme.

Agreed. Seems to be the best option.

PwnArt1st:
can this theme be separated from oscar for now at least?

ibrokemypie:
in my fork i was able to fix all the issues i had, and then split it off as oscar-neue, leaving the original oscar as is. https://github.com/PwnArt1st/searx/compare/master@%7B2hours%7D...master

could be a stopgap to add the theme to instances before the other issues mentioned have been dealt with

jmaris:
This is really **REALLY** nice

ibrokemypie:
close now?

GreenLunar:
Beautiful theme.

The dropdown menu appears to be buggy on small screens.

asciimoo:
closing - f496dc353d336ffe11dde161963d72366f0eee16

Issue #519, opened 2016-03-05
Proxying via Tor


ghost:
Is it possible to proxy to the search engines via Tor using socks5 to port 9050? I've tried a few things and so far none of them have worked.

GreenLunar:
Also consider to forge/switch timezone once IP changes or every X minutes.

**Beware of changing time itself**, because Tor would not let you in to the Tor network once your time has an unreasonable difference (e.g. a difference of an hour).

Notice that only after I setup system clock to a reasonable time, Tor would connect.

```
Apr 13 16:55:29.000 [notice] Application request when we haven't used client functionality lately. Optimistically trying directory fetches again.
Apr 13 17:56:09.000 [notice] Your system clock just jumped 3595 seconds forward; assuming established circuits no longer work.
```

```
$ tailf /var/log/tor/tor.log
Apr 13 16:54:48.000 [notice] Bootstrapped 5%: Connecting to directory server
Apr 13 16:54:48.000 [notice] Bootstrapped 10%: Finishing handshake with directory server
Apr 13 16:54:49.000 [notice] Bootstrapped 15%: Establishing an encrypted directory connection
Apr 13 16:54:49.000 [notice] Bootstrapped 20%: Asking for networkstatus consensus
Apr 13 16:54:50.000 [notice] I learned some more directory information, but not enough to build a circuit: We have no recent usable consensus.
Apr 13 16:54:50.000 [notice] Bootstrapped 25%: Loading networkstatus consensus
Apr 13 16:54:51.000 [notice] Application request when we haven't used client functionality lately. Optimistically trying directory fetches again.
Apr 13 16:54:52.000 [notice] I learned some more directory information, but not enough to build a circuit: We have no recent usable consensus.
Apr 13 16:55:05.000 [notice] Application request when we haven't used client functionality lately. Optimistically trying directory fetches again.
Apr 13 16:55:29.000 [notice] Application request when we haven't used client functionality lately. Optimistically trying directory fetches again.
Apr 13 17:56:09.000 [notice] Your system clock just jumped 3595 seconds forward; assuming established circuits no longer work.
Apr 13 17:56:09.000 [notice] Tried for 3680 seconds to get a connection to [scrubbed]:443. Giving up. (waiting for circuit)
Apr 13 17:56:09.000 [notice] Tried for 3681 seconds to get a connection to [scrubbed]:443. Giving up. (waiting for circuit)
[...]

Issue #413, opened 2015-09-02
request for .onion searches


ghost:
I might take a look at this later and try to write some plugins for tor search engines.

It would be nice to be able to directly address those searches.
Although I am not sure if this would have to be filtered or displayed in some way like ahmia.fi does it, where they include things like "SCAM!" or "MALICIOUS" or similar in their results.

PR #565, opened 2016-05-20
New category: Onions

MarcAbonce:
With engines Ahmia, Torch and Not Evil.

Disabled by default. Can be enabled by setting the SOCKS proxies to wherever Tor is listening and setting using_tor_proxy as True.

To avoid manually adding the timeout on each engine, you can set extra_proxy_timeout to account for Tor's (or whatever proxy used) extra time.

Should solve #413 and probably #519


GreenLunar:
There is also **not Evil** at http://hss3uro2hsxfogfq.onion/ (was **TorSearch** at http://kbhpodhnfxl3clb4.onion/)

MarcAbonce:
By the way, ~~htk2rvtgj6vv3ppj.onion~~ is currently running this branch.
Edit: ~~mze6ter6fchwojcr.onion~~ ~~http://jiwfbtg2kfkadjku.onion~~ http://7ea7vegvcqgdzxjz.onion

asciimoo:
@a01200356 great, the only problem is that it isn't compatible with the existing non-tor configuration. It would be cool to provide both to the users. Could be a separated `settings_tor.yml` for the tor config.

MarcAbonce:
Sorry for the delay.
I just updated the PR. But instead of having two settings files, I added the using_tor_proxy setting that should automatically set up all the other Tor specific configurations.
When set to False or commented out, as default, everything should behave exactly the same as current version on upstream.

gsantner:
@a01200356 I'm currently setting up searx and gonna try the version of your branch.
I can give some feedback If I run into problems.

MarcAbonce:
I just fixed the issue I had with not evil's encoding.

About the last commit, I just assume that using Tor Browser's user agent would make our traffic seem more "normal" when using Tor than using a random user agent. However, this is merely an assumption on my part. I can't really tell if it actually helps or not.

ghost:
Looks interesting.

searx/utils.py
```diff
@@ -33,14 +33,18 @@
...
 
 blocked_tags = ('script',
                 'style')
 
 
 def gen_useragent():
     # TODO
-    return ua.format(os=choice(ua_os), version=choice(ua_versions))
+    if settings['outgoing'].get('using_tor_proxy'):
+        return tor_browser_ua
```
dalf: Why a specific behavior for Tor ?

MarcAbonce:
~~With the fix in #1121, Google doesn't block Tor exit nodes any more.~~
I also reverted the Tor Browser user-agent in Tor requests, I don't think it makes any difference anyway.
Also, DuckDuckGo currently uses the normal website because the hidden service is extremely slow, which makes it reach timeout most of the time.
An updated instance running this branch is in ~~http://jiwfbtg2kfkadjku.onion~~ http://7ea7vegvcqgdzxjz.onion

asciimoo:
@MarcAbonce sorry for the delayed answer. I still think that this is a useful functionality, could you please rebase it to the latest master? I'm happy to do it if you don't have capacity.

MarcAbonce:
I rebased to master and I also added another commit.

asciimoo (approved):
Awesome, thanks!

asciimoo:
This is a great addition, we should mention it in the documentation.

dalf:
Note about the plugin, it can't be disabled, I guess on purpose :
https://github.com/searx/searx/blob/db703a0283ee169381aeea97c678e666ae508348/searx/plugins/__init__.py#L179-L188

The plugin is loaded after `enabled_plugins` parsing and in the preferences (in the oscar theme), it is hidden:
https://github.com/searx/searx/blob/db703a0283ee169381aeea97c678e666ae508348/searx/templates/oscar/preferences.html#L261

MarcAbonce:
Yeah, maybe it would be better to extend the new preferences lock to include plugins and use that instead?

Perflyst:
Looks like if a proxy is configured the proxy is used for all search engines and not only for onion search engines. Is there an option to use the proxy (in this specific use-case the Tor socks proxy) only for onion search engines?

dalf:
You can specify a proxy per engine.
See https://searx.github.io/searx/admin/settings.html#engine-settings

So you have to copy / paste the proxy section for each onion engine
or you can use YAML anchor ( https://en.wikipedia.org/wiki/YAML#Advanced_components )

PR #584, opened 2016-06-10
add config endpoint

kvch:
Returns `plugins`, `engines`, `categories`, `instance_name`, `locales`, `default_locale`, `default_theme`, `autocomplete` and `safesearch` setting values of an instance.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -49,6 +49,7 @@ from flask import (
     redirect, send_from_directory
 )
 from flask.ext.babel import Babel, gettext, format_date
+from flask.json import jsonify
 from searx import settings, searx_dir
 from searx.engines import (
     categories, engines, get_engines_stats, engine_shortcuts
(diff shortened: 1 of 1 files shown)


asciimoo:
great, thanks

dalf:
great! it can be used by stats.searx.oe5tpo.com or by the version I've started to write in Python : https://github.com/dalf/searx-stats2
(help is welcome) 

I don't know if there is way to add a hash of the source code of each engine. 

Some instances follow the master branch, some others follow the release I guess. That's why the Searx version number is not enough in my opinion.

asciimoo:
> I don't know if there is way to add a hash of the source code of each engine. 

sure, we can store the sha hash of the engine files

> Some instances follow the master branch, some others follow the release I guess. That's why the Searx version number is not enough in my opinion. 

yap, it would be good to add at least the git HEAD id to the version string (e.g.: `0.9.0+5dd2340`).

Issue #560, opened 2016-05-14
Extract more data from wikidata


dalf:
More data could be extracted from wikidata and be shown in the infoboxes.

For example searching for ["searx"](http://searx.me/?q=!wikidata+searx&category_general=on) the license and the programming language could be shown (see the [wikidata page](https://www.wikidata.org/wiki/Q17639196))

Technical note : with the used API, the result contains only Q1131681 and Q28865 without the names (for two mentioned fields). 

But using the `parse` API the problem can be solved. Example for the searx page : 

https://www.wikidata.org//w/api.php?action=parse&format=json&uselang=en&page=Q17639196&redirects=1&prop=text%7Clanglinks%7Crevid%7Cwikitext&disableeditsection=1&disabletidy=1&preview=1&sectionpreview=1&disabletoc=1&utf8=1&formatversion=2

the HTML in the "text" entity can be parsed easily since there are `id` and `class` attributes

MarcAbonce:
currently working on this here: https://github.com/a01200356/searx/commits/wikidata

dalf:
Thanks a lot !
I've created a wiki page about the properties that can extracted : https://github.com/asciimoo/searx/wiki/wikidata

MarcAbonce:
I added many, but not all, of the properties in the list. I tried to give as much relevant information without overwhelming the user with too much data (for example printing the full cast of a film would be too much, imo).
Nevertheless, it should be easy to add, remove or change the order of properties in the engine, so this feature can be tweaked as it gets used.

The extra requests are still synchronous though, so I just kept result_count to 1 to avoid problems.

dalf:
PR has been merged

PR #588, opened 2016-06-12
[enh] More data from Wikidata

MarcAbonce:
See #560

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .coverage
+coverage/
 .installed.cfg
 engines.cfg
 env
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -1,33 +1,57 @@
-import json
+# -*- coding: utf-8 -*-
+"""
+ Wikidata
+
+ @website     https://wikidata.org
+ @provide-api yes (https://wikidata.org/w/api.php)
+
+ @using-api   partially (most things require scraping)
+ @results     JSON, HTML
+ @stable      no (html can change)
+ @parse       url, infobox
+"""
 
 from searx import logger
 from searx.poolrequests import get
-from searx.utils import format_date_by_locale
+from searx.engines.xpath import extract_text
 
-from datetime import datetime
-from dateutil.parser import parse as dateutil_parse
-from urllib import urlencode
+from json import loads
 from lxml.html import fromstring
-
+from urllib import urlencode
 
 logger = logger.getChild('wikidata')
 result_count = 1
+
+# urls
 wikidata_host = 'https://www.wikidata.org'
 url_search = wikidata_host \
     + '/wiki/Special:ItemDisambiguation?{query}'
 
 wikidata_api = wikidata_host + '/w/api.php'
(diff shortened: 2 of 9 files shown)


dalf:
When there is more than one official website, there is a bug : there is one button "official website" with the contacted URLs.

Example : World of warcraft.

A different point : The links to YouTube, Twitter etc are not implemented, is it because these websites may / doesn't not respected privacy ? ( P2002,  P2013,   P2037,  P2003, ...)

MarcAbonce:
Fixed!

About the social media links, I hadn't really thought about it, but I added them anyway. The user can choose whether or not to click those links, just like with normal search results here.

dalf:
:-) 

Few more notes / requests : 
- the JSON format is not used so `%7Cwikitext` can be removed from `url_detail`
- the P18, P154 properties give a reference to an image.

Example, the wikidata page about [Linux](https://www.wikidata.org/wiki/Q388) contains :
https://commons.wikimedia.org/wiki/File:Tux.svg

the image link (png or jpg) is : 
https://commons.wikimedia.org/wiki/Special:FilePath/Tux.svg?width=200

width can be 200, 300, 500...

dalf:
The more I spend time to understand / test the wikidata properties, the more I find...
@asciimoo may be merged and update after ?

Notes on current PR : 
- P2046 (area) may contains a list (for [Paris](https://www.wikidata.org/wiki/Q90) for example). I guess it can be the case for other properties. Not a big deal I guess.
- for the image perhaps get the logo and fall back to the image ( `['P154', 'P18', 'P242', 'P41', 'P2716', 'P2910']` ). For `Ubuntu` it seems more relevant for example. BTW the wikipedia engine replaces the image, and perhaps there is other example where the current implementation is better...

New things that could be nice (another PR ?) :
- the property [P402](https://www.wikidata.org/wiki/Property:P402) gives a number, which can be used to link directly to OSM this way : `https://www.openstreetmap.org/relation/{number}`. When this property is set, it can replace the OSM link where the zoom is most of time not set correctly.
- always about geocoordinate : [P1332](https://www.wikidata.org/wiki/Property:P1332), [P1333](https://www.wikidata.org/wiki/Property:P1333), [P1334](https://www.wikidata.org/wiki/Property:P1334) and [P1335](https://www.wikidata.org/wiki/Property:P1335) can be used to define the OSM link.
- the P242 (locator map image) or P15 (route map) give a link to a map. Useful for to have all the stops of a metro line (see [example for P15](https://www.wikidata.org/wiki/Q50761) and [example for P242](https://www.wikidata.org/wiki/Q19892) ). A link to this image could be nice, and second picture that the user can click would be great.

MarcAbonce:
The infoboxes should now merge correctly.
Merging rows with ddg definitions works, but usually leads to some redundancy, such as "Birth" and "Date of birth" or "Creator" and "Created by". So I still left ddg disabled by default.

MarcAbonce:
This PR should be ready to merge, btw.
There are probably still some things that could be tweaked, but these will only be found as more people use this feature.

asciimoo:
great job! thanks

PR #601, opened 2016-07-02
remove unecessary capitalize of name of instance

### LINKED ISSUE #597: autocapitalise in oscar
in both versions of oscar the instance name is put in title case, which should be up to the instance owner. removing that from pointhi.min.css and logicodev.dev.css solves this, and shouldl be pushed to master

kvch:
closes #597


asciimoo:
thanks

PR #604, opened 2016-07-05
Quote variables and backticks in manage.sh

blyxxyz:
Otherwise the script fails with spaces or asterisks in the directory path.


asciimoo:
@blyxxyz thanks

PR #606, opened 2016-07-06
Add favicon for logicodev style

davidar:
Add a favicon to match the oscar theme redesign #559


asciimoo:
IMHO it's better to have one favicon for all themes, because its main purpose is to help recognize your search tabs and it's confusing if it changes on theme switch.

davidar:
Ok

Issue #607, opened 2016-07-06
Wolfram Alpha engine should show result, not input


LuccoJ:
While the Wolfram Alpha infobox neatly shows all (or most) information returned on a query, that is not (yet) available from API calls, and when one simlpy looks at hits from Wolfram Alpha as part of the standard list of hits, the text returned ("content" field in JSON) is the search query you made (the Wolfram Alpha "input"), instead of the "result".

See this screenshot for clarification: http://i.imgur.com/MWDHmDM.png
Instead of "2+2", the search result should probably just show "4" since the user is likely uninterested in seeing a copy of what they typed, but might like to see the result of the operation (which is available somewhere in the code as the infobox shows).

firebovine:
I merged a01200356 's changes, and made two changes so in the event there is no primary text pod returned, it'll choose the first non-Input text pod (ie: test with a search !wa 127.0.0.1).

Sorry if I violated any git workflow! I think I did it correctly!

dalf:
Is it still valid ?

MarcAbonce:
Fixed in #609.

PR #608, opened 2016-07-06
[fix] Result text in Wolfram|Alpha (#607)

MarcAbonce:
#607


LuccoJ:
Cool, thanks. So now you merged your patch and mine into one? Since I see aspects of mine, but also applied to the noapi version of the engine.

MarcAbonce:
Yes. I did the no api version, so now it looks the same with both versions.

PR #609, opened 2016-07-06
Improving Wolfram Alpha search hit content

LuccoJ:
Making WA search hits contain
- the (parsed) input inside the "title" instead of just "Wolfram|Alpha", to better match other hit titles and to confirm correct parsing of input to the user
- the first output field that contains any text (skipping ones that are only pictures; this is usually the most meaningful "result" field) instead of the raw input as the "content", making it additionally possible to obtain WA computations from JSON API calls

This pull request is intended to address issue #607

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -494,6 +494,14 @@ engines:
     engine : wolframalpha_noapi
     timeout: 6.0
     categories : science
+    
+  - name : urbandictionary
+    engine        : xpath
+    search_url    : http://www.urbandictionary.com/define.php?term={query}
+    url_xpath     : //*[@class="word"]/@href
+    title_xpath   : //*[@class="word"]
+    content_xpath : //*[@class="meaning"]
+    shortcut : ud
 
   - name : dictzone
     engine : dictzone
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -494,14 +494,6 @@ engines:
     engine : wolframalpha_noapi
     timeout: 6.0
     categories : science
-    
-  - name : urbandictionary
-    engine        : xpath
-    search_url    : http://www.urbandictionary.com/define.php?term={query}
-    url_xpath     : //*[@class="word"]/@href
-    title_xpath   : //*[@class="word"]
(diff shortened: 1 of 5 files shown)


MarcAbonce:
Lol, it looks like we both commited the same thing at the same time (#608).

LuccoJ:
Yes, it does. I'm not sure whether my patch is more comprehensive or yours is... with mine, you get a title like "Wolfram|Alpha (2+2)" and a content like "Result: 4" (I thought including "Result:" was important because it doesn't always say "Result:" and sometimes the output is quite obscure without that header to explain what it means), and it skips fields that just contain images or the like.

My patch appears to fail to meet some checks, though. I'm not sure I understand them.

MarcAbonce:
Ok. I merged your commit and now it works.
It failed because of the engine's unit tests (in tests/unit/engines/test_wolframalpha_api.py), which expected the result title to be only "Wolfram|Alpha". I changed that.

LuccoJ:
We changed it a bit further to make sure a result is output even when the Wolfram API does not indicate any field as "primary"; in such cases, it will pick the first field after the "Input" one that contains any text at all. This is useful in a number of practical circumstances where, for whatever reasons, WA doesn't give any field the badge of "primary", even in cases where there's one field in total.

MarcAbonce:
Nice. So this PR should be the one being merged, since it's the most complete now.

searx/settings.yml
```diff
@@ -425,7 +425,7 @@ engines:
     engine : wolframalpha_noapi
     timeout: 6.0
     categories : science
-
+
```
asciimoo: please remove this whitespace change

LuccoJ: That should be fixed now, I reverted settings.yml to your version in master.

searx/settings.yml
```diff
@@ -254,6 +254,12 @@ engines:
     disabled : True
     shortcut : habr
 
+  - name : ina
```
asciimoo, line 298: i'm not sure if this really belongs here

asciimoo:
great, thanks

PR #610, opened 2016-07-06
Adding Urban Dictionary xpath engine

LuccoJ:
Simple xpath-based settings for Urban Dictionary.
Not sure which category to put this into; I think a "words", "dictionary", "definitions" or "language" category would be generic and important enough to be worth considering.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -425,6 +425,14 @@ engines:
     engine : wolframalpha_noapi
     timeout: 6.0
     categories : science
+    
+  - name : urbandictionary
+    engine        : xpath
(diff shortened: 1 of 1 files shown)


asciimoo:
settings.yml aleady contains urbandictionary, it is just commented out. Please delete one of them.
Also, it would be good to disable it by default (`disabled: true`).

LuccoJ:
I hadn't really realized it was already there commented out. The commented out version says it worked erratically. This one has seemed to work okay for me so far, but at this point I suggest you reject or hold this merge request, as I'd like instead to add a few xpath configs for various other dictionaries, and I think we can review the matter again when I have them ready.

asciimoo:
okay, thanks

PR #615, opened 2016-07-09
Add ScanR structures search engine

mmuman:
In theory ScanR should also search for projects but the API is different,
so we'd need another engine.

### DIFF
--- a/searx/engines/scanr_structures.py
+++ b/searx/engines/scanr_structures.py
@@ -1,0 +1,78 @@
+"""
+ ScanR Structures (Science)
+
+ @website     https://scanr.enseignementsup-recherche.gouv.fr
+ @provide-api yes (https://scanr.enseignementsup-recherche.gouv.fr/api/swagger-ui.html)
+
+ @using-api   yes
+ @results     JSON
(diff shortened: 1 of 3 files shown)


searx/engines/scanr_structures.py
```diff
@@ -0,0 +1,77 @@
...
+        content = None
+        if 'highlights' in result:
+            content = result['highlights'][0]['value']
+
+        # append result
+        results.append({'url': url + 'structure/' + result['id'],
+                        'title': result['label'],
+                        # 'thumbnail': thumbnail,
+                        'img_src': thumbnail,
+                        'content': content})
```
asciimoo: please use `searx.utils.html_to_text` or `cgi.escape` to escape the content

mmuman: Oh I didn't notice searx did its own highlighting of the search term in a different way.

mmuman:
Sh*, unit tests :D

mmuman:
Should be working now.

kvch:
:+1:

asciimoo:
cool, thanks

PR #620, opened 2016-07-11
Fix line break before binary operator

ghost:


kvch:
Based on what convention is this the accepted place of these operators? AFAIK PEP8 (https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) states present form is the accepted version.

asciimoo:
@stepshal  we follow pep8 code style recommendations, so this pr is invalid, closing

PR #623, opened 2016-07-13
[fix] multilingual duckduckgo

MarcAbonce:
Language support for duckduckgo engine.

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -16,6 +16,7 @@
 from urllib import urlencode
 from lxml.html import fromstring
 from searx.engines.xpath import extract_text
+from searx.languages import language_codes
 
 # engine dependent config
 categories = ['general']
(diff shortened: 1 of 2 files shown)


searx/engines/duckduckgo.py
```diff
@@ -39,13 +37,22 @@ def request(query, params):
...
-
-    params['url'] = url.format(
-        query=urlencode({'q': query, 'kl': locale}),
-        offset=offset)
+        locale = params['language'].split('_')
+        if len(locale) == 2:
+            # country code goes first
+            locale = locale[1].lower() + '-' + locale[0].lower()
+        else:
+            # doesn't actually do anything because ddg requires both country and language
```
kvch: I tried your PR with a locale that is only one long. It did not work as you had written here. If DDG requires both country and language why did you include the `else`? Is it possible that there will be locales which match the format DDG requires without any transformation?

kvch:
Apart from the question I added inline, I think it is cool. :)

MarcAbonce:
You're right. I just changed it to try to guess the country if none is given.
So language support should work now, even if no country is specified by the user.

asciimoo:
nice, thanks

PR #624, opened 2016-07-14
Add Google Scholar search engine

davidar:


asciimoo:
@davidar please use lxml instead of adding beautifulsoup as a new dependency

davidar:
@asciimoo using the xpath engine instead now :)

asciimoo:
great, thanks

PR #630, opened 2016-07-16
Add Crossref search engine and DOAI rewrite plugin

davidar:


asciimoo:
Wouldn't be better to create a specific engine which rewrites the results by default instead of introducing a new plugin which also checks all other engine's results?

davidar:
@asciimoo Possibly?

The plugin isn't specific to this engine, as any engine could potentially return DOI links. I probably should have made these two separate PRs.

davidar:
@asciimoo Done.

I also updated the plugin so that it's able to catch more links now (not limited to just the doi.org domain)

asciimoo:
great, thanks

PR #633, opened 2016-07-19
[fix] deviantart engine xpaths

kvch:
There were xpath changes on deviantart, so the engine did not return any results. In this patch the xpaths are fixed and the changes are followed up in the unit test.


kvch:
Now it seems to me that it still does not work...

asciimoo:
thanks

Issue #557, opened 2016-05-06
Limit Results by Time Range


ibrokemypie:
Feature request to limit results by time range, eg last 24 hours, last week, last month, last year.

LuccoJ:
Seconded. At least some search engines being used as backends (e.g. Google) support this, so it should be possible to provide it to some extent at least.

dalf:
+1

ammarnajjar:
+1

Yodan:
+1

1xPdd:
+1

Would also be cool if one could set date ranges, e.g., 1998-2000, rather than simply newness, e.g., 2015-present.

asciimoo:
closing, see #634

Issue #383, opened 2015-07-06
Filter search results by date


Francisco-Ordas:
In Google Search, I often use date criteria to filter results. For me, it would it's a key feature to replace Google Search by Searx.

dalf:
Technical note : 
- a way in searx to specify the date criteria (from the user point of view and from the code point of view)
- when a date criteria is asked, searx must not use engines that don't support this criteria
- implements for different search engine this criteria

asciimoo:
closing, see #634

PR #634, opened 2016-07-19
support time range search

kvch:
### Features (as requested in #383, #557)
- advanced search form under search input
- time range search
  - last day
  - last week
  - last month
- supported engines
  - Google
  - Yahoo
  - DuckDuckGo
  - Google Images
  - Deviantart
### Index page with Advanced settings closed

index
### Index page with Advanced settings open

index-advanced
### Search page with Advanced settings closed

search
### Search page with Advanced settings open

search-advanced
### Possible enhancements
- support more search engines
- search based on custom range
- add more settings to Advanced settings (e.g. safe search)


dalf:
Great ! In the result page, would it be more clear if the categories and time range are displayed / expanded ?
[EDIT] Suggestion : Time range as a combo box which takes far less space. So the safe search, the search language / locale could all fit in the screen (to have a live example : duckduckgo does that).

kvch:
@dalf What do you mean by displayed/expanded? Should the categories be shown always without clicking on advanced search?

When I created the UI my main goal was to be user-friendly, not compact. I'va read at a few pages that using "radiobuttons" are more user-friendly that using dropdowns. However, nearly all of the other search services use dropdowns to select time range. Those websites have (I suppose) UX teams with great knowledge, so it is wiser to follow their methods. Also, for users it would be much easier to adopt to searx if selecting the time range is similar to the widely-used alternatives. In a few days I'll push the redesign version with dropdowns. :) Thanks for your suggestion. :)

dalf:
Yes, on the result page I think the categories / time range should be always displayed. I think the UX / UI should give a quick answer without click to the question "there are some results, but what was the request ?". By request I mean everything (text, time range, category, ... and language but that's another story)

It's not that the user will forgot what he has just entered, but he can adjust right away, leave the page and come back after, etc...

asciimoo:
very-very awesome, thanks! =)

1xPdd:
The UI looks great! But, if I could, I'd really like to suggest (again) that dates should not be only measured as relative distance from the present as in last day/week/month. Library search engines and the likes allow dates to be limited to an arbitrary date range, as in from year to year. No general search engine that I know of does this. The advantage of a more flexible date adjustment is that it enables Searx to function as a tool for historical research. As to UI, it could be done via a slider or using a form.

virtadpt:
Will this functionality be accessible from the search API?

will76:
What about 'last year'? It's the best option to search for updated articles on a subject.

PR #639, opened 2016-07-23
add digbt engine - fixes #638

### LINKED ISSUE #638: with KAT being targeted, could we get digbt search working as a backup?
I predict this wont be the last we here about torrent sites being targeted and after KAT domain seizure i believe the mirrors might get targeted as well. could searx get http://www.digbt.org/ as a backup since btdigg is also down

kvch:
Introduce new XPATH engine digbt.org based on request in #638.

### DIFF
--- a/searx/engines/btdigg.py
+++ b/searx/engines/btdigg.py
@@ -16,6 +16,7 @@ from urllib import quote
 from lxml import html
 from operator import itemgetter
 from searx.engines.xpath import extract_text
+from searx.utils import get_torrent_size
 
 # engine dependent config
 categories = ['videos', 'music', 'files']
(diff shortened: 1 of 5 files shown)


dalf:
:-)

I don't know how to keep the xpath engine to add information about magnet, upload, download.
That's really help to choose the right torrent.

kvch:
TODO (so nothing will be forgotten in conenction with this PR)
- [ ] refactor XPATH engine so it can support torrent results
- [ ] let digbt use the refactored XPATH engine

kvch:
I did not refactor the XPATH engine, because I did not want to pollute it with torrent related codes. So I moved digbt into a seperate engine. I did a small refactor: calculation of sizes of torrents were moved to utils to avoid code duplication.

asciimoo:
thanks

PR #645, opened 2016-07-25
[mod] remove the kickass engine from settings.yml

dalf:
There are some mirrors, should the kickass engine be modified to point one of them ?

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -324,10 +324,6 @@ engines:
     timeout : 10.0
     disabled : True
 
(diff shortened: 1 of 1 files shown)


asciimoo:
@dalf do you have any trusted information about these mirror sites?

dalf:
After a quick overview, all websites are mirror of KAT just before everything stops.
The register link goes to ads.
I wouldn't trust these mirrors.

asciimoo:
thanks

PR #655, opened 2016-08-08
set type of cursor to pointer in advanced search

kvch:
From now on cursor will change to pointer when going over "Advanced search" and its input fields.


asciimoo:
thanks

Issue #649, opened 2016-07-29
Idea] An endless research


mcl1234:
Hi,

Is it possible to add an endless research , like here for google search : https://openuserjs.org/scripts/tumpio/Endless_Google

?

davidar:
Here's a quick snippet that does this for the oscar theme:

``` js
$(document).ready(function() {
    var win = $(window);
    win.scroll(function() {
        if ($(document).height() - win.height() == win.scrollTop()) {
            var formData = $('#pagination form:last').serialize();
            $('#pagination').html('Loading next page...');
            $.post('/', formData, function (data) {
                var body = $(data);
                $('#pagination').remove();
                $('#main_results').append(body.find('.result'));
                $('#main_results').append(body.find('#pagination'));
            });
        }
    });
});
```

However, I'm not sure what the best way to generalise it for all the themes would be.

asciimoo:
@davidar currently only oscar theme supports client side plugins.

davidar:
@asciimoo ah, that makes it easier :)

mcl1234:
thanks a lot ! :)

asciimoo:
closing - see #658

PR #658, opened 2016-08-11
[enh] infinite scroll

davidar:
See #649

### DIFF
--- a/searx/plugins/infinite_scroll.py
+++ b/searx/plugins/infinite_scroll.py
@@ -1,0 +1,8 @@
+from flask_babel import gettext
+
+name = gettext('Infinite scroll')
+description = gettext('Automatically load next page when scrolling to bottom of current page')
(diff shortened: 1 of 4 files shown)


searx/plugins/infinite_scroll.py
```diff
@@ -0,0 +1,8 @@
+from flask_babel import gettext
+
+name = gettext('Infinite scroll')
+description = gettext('Automatically load next page when scrolling to bottom of current page')
+default_on = True
```
asciimoo: I'd leave it `False` by default. Extra files need to be downloaded even for users who cannot use this feature (eg js blocking).

davidar: Yeah, I wasn't sure about that. Fixed.

asciimoo:
cool, thanks =]

Issue #334, opened 2015-05-27
[ENGINE] Dict.cc


cy8aer:
https://github.com/rbaron/dict.cc.py seems to have an api for dict.cc. Main problem: how to tag this?
!en-de house

Cqoicebordel:
Nice !!!
And good idea for the tag !

asciimoo:
Great idea! The search syntax could be handled the same way as in the currency convert engine - fetching results from this engine only if the search query format is `{lang1} {lang2} .+`

pointhi:
I propose a generic implementation of the translation interface, because dict.cc isn't the only possible translator.

For example, http://wadoku.de is a very good german<->japanese translator, and the google translator can handle more cases than dict.cc. A generic approach like for normal search engines would be the best way.

Probably done by a plugin, which parse the query, and returning some tags which are detected and used by the translation-engines. And the plugin is merging them together afterwards. This would be a clear way to allow special engines while keeping the core-code clean from specific algorithms.

cy8aer:
wadoku seems to have an api too (the whole server is oss): https://github.com/Wadoku - but it is rails. And yes: multi-engine would be nice if there is the possibility to have a module system (maybe?)

GreenLunar:
Any news on this?

asciimoo:
This can be implemented as a new engine like the currency convert engine. Engines can parse the search query on their own way. Search syntax could be `!dc en-de words to translate`

PR #661, opened 2016-08-14
Online Etymology Dictionary

davidar:
Related: #334

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -145,6 +145,17 @@ engines:
     shortcut : ddg
     disabled : True
 
+  - name : etymonline
+    engine : xpath
+    paging : True
+    search_url : http://etymonline.com/?search={query}&p={pageno}
+    url_xpath : //dt/a[1]/@href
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -145,6 +145,17 @@ engines:
...
+  - name : etymonline
+    engine : xpath
+    paging : True
+    search_url : http://etymonline.com/?search={query}&p={pageno}
+    url_xpath : //dt/a[1]/@href
+    title_xpath : //dt
+    content_xpath : //dd
+    suggestion_xpath : //a[@class="crossreference"]
+    first_page_num : 0
+    shortcut : et
```
asciimoo, line 157: It would be good to disable it by default, since it has no https support

asciimoo: ah, it isn't part of any category, so it will be disabled by default

asciimoo (approved):

asciimoo:
thanks

PR #664, opened 2016-08-15
uncapitalize searx in templates

kvch:
The name of this searx engine is uncapitalized **searx**. However, it is not consistent in the templates. So it can be confusing. I replaced the occurences of "Searx" to "searx" in sentences for the sake of consistency.


asciimoo:
thanks

PR #667, opened 2016-08-20
loosened up the dependency versions

Profpatsch:
Resets stricter versions in order to work with the bounds of the 0.9
release (on NixOS).

They should be as lenient as possible in order to ease the life of maintainers.


dalf:
for requests I would suggest to require at the version 2.10.0 (last version 2.11.1), to support socks proxy (require for Tor).

devonhk:
Why can't nixos use the more recent version of the dependencies? Are they not using pip as the package manager?

dalf:
It seems that the NixOS philosophy is to manage all dependencies with its own package manager  : 
https://nixos.org/wiki/Python

But I see project like this one : 
https://github.com/ktosiek/pip2nix

I guess @Profpatsch can say more about it ?

matejc:
NixOS has a Nix package manager at its core, with witch you can package virtually _anything_, pip2nix and pypi2nix are here to generate Nix code for packaging python packages and its dependencies. I packaged the Searx into Nix, but then there where no tools like pip2nix and pypi2nix, or they where not reliable at the time. With Nix you can package any package with whatever dependencies you want, for example you can have multiple Searx versions with different versions of dependencies without interference. Why I didn't packaged it the right way after those tools became available, is partially because I am using Docker management software on my servers and partially because I am lazy.

matejc:
To make this more clear, you do not need to adjust dependencies of any package for Nix, the Searx has to be properly packaged with Nix.

Profpatsch:
It’s not a problem with nix per-se, but rather every package manager. If the versions are fixed, the packager has to jump through hoops to get the versions right (in a consistent fashion with other packages). Nix makes this easier than most, since you could have more than one version of a package (which you also have to support afterwards).

dalf:
I think a lot people installs searx dependencies in a virtualenv. This is the documented [way to install](https://asciimoo.github.io/searx/dev/install/installation.html#basic-installation) it.

But `certifi` could be would without a specified version, to make sure to install the last one, since it's involved security.

Profpatsch:
> I think a lot people installs searx dependencies in a virtualenv. This is the documented way to install it.

Whatever does that have to do with packaging for distros?

kvch:
I don't think that it is the responsibility of searx devs to guarantee that the project runs on a distro but the package maintainers. Thus, I would not merge this PR because it is packaging related. If someone wants to package it he/she has to collect all the dependencies and make sure it installs correctly.

Currently, only the documented installations are supported. I don't think that we should consider supporting more methods soon. Present ones are sufficient for most of the users as @dalf pointed out.

PR #671, opened 2016-08-22
Custom 404 message - fixes #317

### LINKED ISSUE #317: custom error-sites
We could support custom error sites inside the templates, to show for example 404-Errors in the design of the template instead of some naked text. (continuous design).

But we have to be cautious with 5xx-Errors caused by template problems (fallback-solution, if the error-template is not working like expected)

kvch:
Add custom message on HTTP 404. The default message is "Page not found". I am open to better messages. :) #317

### DIFF
--- a/searx/templates/courgette/404.html
+++ b/searx/templates/courgette/404.html
@@ -1,0 +1,7 @@
+{% extends "courgette/base.html" %}
+{% block content %}
+<div class="center">
+    <h1>{{ _('Page not found') }}</h1>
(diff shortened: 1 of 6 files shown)


dalf:
why not use gettext ?
if there is no language cookie, the error is in English as now, otherwise it is translated.

kvch:
@dalf sorry, i misunderstood the original issue. :) i fixed the files, and amended the commit. thanks for the review.

asciimoo:
thanks

PR #672, opened 2016-08-27
Self information on later pages

### LINKED ISSUE #668: Infinite scroll: answer are repeated on each page
How to reproduce : search for ["user agent"](https://searx.me/?q=user+agent) with Infinite scroll activated.

Should the answer be disabled except the first page ? or should Infinite Scroll hide the answer ?

I vote for the first option : disabled answers except on the first page on the server side.

kvch:
This patch fixes #668. So answers from self information plugin do not show in the pages of results from the second page.


asciimoo:
thanks

PR #674, opened 2016-08-29
add noopener noreferrer to target=_blank

kvch:
Set `noopener` and `noreferrer` for opening new tab. This issue and fix is based on [this blogpost](https://dev.to/ben/the-targetblank-vulnerability-by-example) and [this blogpost](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/).


asciimoo:
Thanks.
I've started to reimplement the `target="_blank"` feature to work without js and it will fix this issue too.

kvch:
\o/ I prefer your solution. So after your fix is merged, this can be closed.

PR #678, opened 2016-09-06
[engine] dictzone + mymemory.translated engine

potato:
engine to https://dictzone.com to act as a dictionary for 1 word searches,
and engine to http://mymemory.translated.net/ to act as a translator for multi-word searches

translates when the search query matches

> [lang1]-[lang2] [word|words]

format, where `lang1` is the language translated from (full name, or 2-char abbreviation) and `lang2` is the language to translate to. if only 1 word is given the dictzone engine triggers (much faster, but can't translate multiple words), otherwise mymemory.translated api is used (which is disabled by default, use it with `!lt`)


asciimoo:
awesome! thanks =]

Issue #324, opened 2015-05-17
Embed a searx search bar inside a web portal


HLFH:
Hi,

Is there some HTML5/JS code to add a searx search bar inside a web portal, and when you use this searx search bar - for example you type "searx is cool', you are redirected from the web portal (example.com) to the searx results (searx.example.com/?q=searx%20is%20cool&pageno=1&category_general).

On this private web portal, there are several shortcuts to subdomains with webapps like ownCloud, gogs, Transmission, etc. I would like to put in the header a basic searx search bar. Is that a thing possible?

Thanks in advance,

dalf:
How do you want to include searx ? with a iframe as you descibe it ? another solution could use the JSON API (even if there is some missing information like the info boxes or the right side).

I think the IFrame solution is possible but 
- I don't know why the search result replace the portal page (this is a simple HTML form without target or something else).
- the theme has to be modify to be nicely include into a portal (source code of the oscar theme : https://github.com/asciimoo/searx/tree/master/searx/templates/oscar ). Perhaps create a new theme ?

Can you post some code to test what you descibe ?

Cqoicebordel:
I think what he wants is that searx provide a small page to be included in a iframe, with only the text input and a search button, which would redirect to a choosen page.
That could be really easy to implement.

HLFH:
@Cqoicebordel Hi and thanks for translation.

@dalf What I would like to add as @Cqoicebordel explained is a "text input and a search button, which would redirect to a chosen page": the dedidated results page of the searx instance. And where? For example here: https://gitnote.eu in the header, before the "webapps boxes". This call-to-action shortcut would improve the use of Searx and reduce the direct use of Google Search. In fact, a lot of open source users love their personal data, so they want to protect them with ownCloud, Cozy, ...and also with dedicated operating systems like YunoHost, arkOS. In this trend, these people use a lot of open source webapps in one single domain. 
![screenshot 2015-05-19 at 13 48 06](https://cloud.githubusercontent.com/assets/2359059/7702084/b9109b60-fe2d-11e4-822e-ff076764b972.png)
For this Yunohost example, I believe we need to add the search bar - maybe in a lighter version - on this homepage/web portal. 
![screenshot 2015-05-19 at 13 53 15](https://cloud.githubusercontent.com/assets/2359059/7702179/74abc764-fe2e-11e4-95bc-53167979d854.png)
If it could be done with an iFrame or using the JSON API, I would love to know how to do it.

[...]

PR #679, opened 2016-09-06
docs: time_range search param, embed

kvch:
- document time_range parameter of search query
- add embedded search example

Closes: #324


asciimoo:
Thanks!

PR #692, opened 2016-09-09
Feature/horriblesubs engine integration

Yuyu:
Includes horriblesubs.info as a searchable engine.
I've made it include airing/release times from shows if there's any matches in there too. If your instance runs for >3 months it's likely the "index" will be obsolete, but I'm not sure how to work around this yet.


asciimoo:
@Yuyu0 it would be good to call `index_schedules()` periodically if a timeout exceeded (e.g. from the `request()` function). Wolframalpha engine's session handling is similar.

asciimoo:
Closing due to inactivity.

PR #699, opened 2016-09-18
kickass.to is seized by ICE

leo9843:
kickass.to domain is seized by ICE and now only one domain kickass.cd is working...


dalf:
Thank you. Could you adjust the test : 
https://github.com/asciimoo/searx/blob/master/tests/unit/engines/test_kickass.py#L17

You can run the tests : 
https://asciimoo.github.io/searx/dev/quickstart.html#how-to-run-tests

PR #724, opened 2016-10-06
[engine] PDBe (Protein Data Bank Europe)

Athemis:
Engine to search for 3d protein structures in the PDBe (https://www.ebi.ac.uk/pdbe). Since the API is queried in fulltext search mode, either PDB codes (e.g. 1KBL) or generic searches (e.g. "amino transferase") are supported.
### Example

pdbe_engine_example

### DIFF
--- a/searx/engines/pdbe.py
+++ b/searx/engines/pdbe.py
@@ -1,0 +1,109 @@
+"""
+ PDBe (Protein Data Bank in Europe)
+
+ @website       https://www.ebi.ac.uk/pdbe
+ @provide-api   yes (https://www.ebi.ac.uk/pdbe/api/doc/search.html),
+                unlimited
+ @using-api     yes
+ @results       python dictionary (from json)
+ @stable        yes
+ @parse         url, title, content, img_src
+"""
+
+from json import loads
+from flask_babel import gettext
+
+categories = ['science']
+
+hide_obsolete = False
+
+# status codes of unpublished entries
+pdb_unpublished_codes = ['HPUB', 'HOLD', 'PROC', 'WAIT', 'AUTH', 'AUCO', 'REPL', 'POLC', 'REFI', 'TRSF', 'WDRN']
+# url for api query
+pdbe_solr_url = 'https://www.ebi.ac.uk/pdbe/search/pdb/select?'
+# base url for results
+pdbe_entry_url = 'https://www.ebi.ac.uk/pdbe/entry/pdb/{pdb_id}'
+# link to preview image of structure
+pdbe_preview_url = 'https://www.ebi.ac.uk/pdbe/static/entry/{pdb_id}_deposited_chain_front_image-200x200.png'
+
+
+def request(query, params):
+
+    params['url'] = pdbe_solr_url
+    params['method'] = 'POST'
+    params['data'] = {
+        'q': query,
+        'wt': "json"  # request response in parsable format
+    }
+    return params
+
+
+def construct_body(result):
+    # set title
+    title = result['title']
+
+    # construct content body
+    content = """{title}<br />{authors} {journal} <strong>{volume}</strong>&nbsp;{page} ({year})"""
+
+    # replace placeholders with actual content
+    try:
+        if result['journal']:
+            content = content.format(
+                title=result['citation_title'],
+                authors=result['entry_author_list'][0], journal=result['journal'], volume=result['journal_volume'],
+                page=result['journal_page'], year=result['citation_year'])
+        else:
+            content = content.format(
+                title=result['citation_title'],
+                authors=result['entry_author_list'][0], journal='', volume='', page='', year=result['release_year'])
+        img_src = pdbe_preview_url.format(pdb_id=result['pdb_id'])
+    except (KeyError):
+        content = None
+        img_src = None
+
+    # construct url for preview image
+    try:
+        img_src = pdbe_preview_url.format(pdb_id=result['pdb_id'])
+    except (KeyError):
+        img_src = None
+
+    return [title, content, img_src]
+
+
+def response(resp):
+
+    results = []
+    json = loads(resp.text)['response']['docs']
+
+    # parse results
+    for result in json:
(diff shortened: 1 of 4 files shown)


searx/engines/pdbe.py
```diff
@@ -0,0 +1,120 @@
...
+        content = None
+        img_src = None
+        pass
+
+    # construct url for preview image
+    try:
+        img_src = pdbe_preview_url.format(pdb_id=result['pdb_id'])
+    except (KeyError):
+        img_src = None
+        pass
```
devonhk: The pass is redundant

searx/engines/pdbe.py
```diff
@@ -0,0 +1,120 @@
...
+                page=result['journal_page'], year=result['citation_year'])
+        else:
+            content = content.format(
+                title=result['citation_title'],
+                authors=result['entry_author_list'][0], journal='', volume='', page='', year=result['release_year'])
+        img_src = pdbe_preview_url.format(pdb_id=result['pdb_id'])
+    except (KeyError):
+        content = None
+        img_src = None
+        pass
```
devonhk: Also redundant

searx/settings.yml
```diff
@@ -339,6 +339,17 @@ engines:
     disabled : True
     shortcut : or
 
+  - name : pdbe
+    engine : pdbe
+    shortcut : pdb
+# Hide obsolete PDB entries.
```
devonhk, line 345: Some of the content on the site is not transmitted over `https` so you might have to disable it by default. One of the project's devs could probably chime in.

devonhk (commented):
Besides the points I brought up the PR looks good.

Athemis:
Thanks for the comments! Pull request has been updated accordingly.

searx/settings.yml
Athemis: Which content exactly? I know, if you access the site directly, and are doing a search there, the images are loaded unencrypted, but the plugin does not crawl the site. Instead it queries the pdbe api using https. The preview image is loaded over https as well.

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+# link to preview image of structure
+pdbe_preview_url = 'https://www.ebi.ac.uk/pdbe/static/entry/{pdb_id}_deposited_chain_front_image-200x200.png'
+
+
+def request(query, params):
+
+    params['url'] = pdbe_solr_url
+    params['method'] = 'POST'
+    # using searx User-Agent
+    params['headers']['User-Agent'] = searx_useragent()
```
kvch: Why is `searx_useragent` is called here? If you set it, you leak that information that a searx instance made the request. If the point was to hide the user agent of the user, it is unnecessary, because searx generates a random user agent when a request is sent.

This function was introduced to authenticate for BASE engine. Doest it use the same auth method as BASE does?

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+
+def response(resp):
+
+    results = []
+    json = loads(resp.text)['response']['docs']
+
+    # parse results
+    for result in json:
+        # reset skip flag
+        skip_result = False
```
kvch: `skip_result` is unnecessary in this function. If you want to skip the result use `continue` to skip this cycle of the loop.

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # catch obsolete entries and mark them accordingly
+        if result['status'] in pdb_unpublished_codes:
```
kvch, line 81: Without `skip_result`:

```
if result['status'] in pdb_unpublished_codes:
    continue
```

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+            skip_result = True
+        elif result['status'] == 'OBS':
+            if not hide_obsolete:
```
kvch: `hide_obsolete` can be moved out from this `if`.

```
if hide_obsolete:
   continue
if result['status'] == 'OBS':
    # process result...
else:
    # process result...
```

This way `else: skip_result = True` is not needed anymore.

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+                    pdb_id=result['superseded_by'], )
+
+                # obsoleted entries don't have preview images
+                img_src = None
+            else:
+                skip_result = True
+        else:
+            title, content, img_src = construct_body(result)
+
+        if not skip_result:
```
kvch: This `if` is not required either, because when `continue` is used, processing never gets here if the result has to be skipped.

kvch (changes requested):

searx/engines/pdbe.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def request(query, params):
+
+    params['url'] = pdbe_solr_url
+    params['method'] = 'POST'
+    # using searx User-Agent
+    params['headers']['User-Agent'] = searx_useragent()
+    params['data'] = {
+        'q': "{}".format(query),
```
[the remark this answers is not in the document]
Athemis: Yep, that's a leftover because the original query was more complex ;)

searx/engines/pdbe.py
Athemis: Ok, then we can leave it out. I'm a bit worried though that a random user agent might lock searx out in the future. The pdbe api usage policy requires users to send a _meaningful_ user agent. We can try though and send something more specific in case searx gets ever blocked.

Athemis:
@kvch Your comments are addressed in the last commits. Thanks!

kvch (approved):

kvch:
@Athemis could you please squash your commits into a single commit? After that it can be merged. :)
Thanks for this cool engine. :)

Athemis:
@kvch Done! Thanks for merging and the helpful advices :)

asciimoo:
@Athemis thank you, very-very cool.
also, @pydo @kvch thanks the useful reviews =]

PR #405, opened 2015-08-24
some Bug fixes

pointhi:
I fixed some minor bugs:
- fix debug switch
- fix #390 
- don't show link to search site when searching with startpage or ixquick

PR #726, opened 2016-10-08
[fix] set the title on the opensearch link tag

mmuman:
Firefox uses the title attributes instead of the ShortName from the xml file
as set in 0fbd7052 which closed #405 

Note I didn't add opensearch to the pix-art template.


kvch (approved):

asciimoo:
thanks!

PR #733, opened 2016-10-22
Search architecture

dalf:
This is a try to clean up the architecture, but this PR most probably falls into this pitfall :  
- https://hackernoon.com/how-to-accept-over-engineering-for-what-it-really-is-6fca9a919263
- http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/

Purposes : 
- isolate the plugins calls
- isolation between parsing the web part and the actual search (Search class). Purpose : to be able to test code easily, to run searx code outside a web server, to filter the search query parameters with plugins more easily, etc...

Details :
- request.request_data contains request.form or request.args (initialize inside pre_request() function)
- Query class is renamed RawTextQuery
- SearchQuery class defines all search parameters
- get_search_query_from_webapp create a SearchQuery instance (basically the previous Search.**init** code)
- Search class and SearchWithPlugins class takes a SearchQuery instance as class constructor parameter  
- SearchWithPlugins class inherites from Search class, and run plugins
- All plugins code is executed inside the try...except block (webapp.py, index function)
- advanced_search HTTP parameter value stays in webapp.py (it is only part of UI)
- multiple calls to result_container.get_ordered_results() doesn't compute the order multiple time (note : this method was call only once before)
- paging value is stored in the result_container class (compute in the extend method)
- test about engine.suspend_end_time is done during search method call (instead of **init**)
- check that the format parameter value is one of these : html, rss, json, rss (before the html value was assumed but some text formatting wasn't not done)

Change of the plugin API (second commit)  :
- pre_search(request, search)
- post_search(request, search)
- on_result(request, search, result)

with
- request is the Flask request
- search a searx.Search instance
- result a searx result as usual

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -25,9 +25,10 @@ from searx.engines import (
     categories, engines
 )
 from searx.utils import gen_useragent
-from searx.query import Query
+from searx.query import RawTextQuery, SearchQuery
 from searx.results import ResultContainer
 from searx import logger
+from searx.plugins import plugins
 
 logger = logger.getChild('search')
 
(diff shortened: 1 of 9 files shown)


asciimoo:
@dalf awesome! This part was a very week point of searx's code - tightly coupled and messy. I'll review it in the following days, thanks! =]

dalf:
There is a lot more to do / more ideas : 
- `get_search_query_from_webapp` is HTTP related. So may be this function should move to another file (webapp.py ? webapp_support.py ?)  
- `.encode('utf-8')` is called on `q` HTTP parameter in autocompleter but not in the normal search (actually it's done at different location to have both values : encoded and not)
- apply the PR #518 ( ? )
- should the [`engine.request` functions](https://github.com/dalf/searx/blob/6aff14f8ce45d22c2a5dbbeb10fff53e3b279e9e/searx/search.py#L311) use the SearchQuery instance instead of copying the value to `request_params` ? If [the lang parameter is specified](https://github.com/dalf/searx/blob/6aff14f8ce45d22c2a5dbbeb10fff53e3b279e9e/searx/search.py#L301), the Search.search method can create a copy of the Search_Query instance with a specific value. It's a lot of work to modified all the engines and tests, but the API to write an engine would be more or less self documented / more clear.

Note(s) :
-  `request` is needed only by the self_info plugin.

searx/search.py
```diff
@@ -268,45 +264,51 @@ def search(self, request):
...
             if hasattr(engine, 'language') and engine.language:
                 request_params['language'] = engine.language
             else:
-                request_params['language'] = self.lang
+                request_params['language'] = search_query.lang
 
             # 0 = None, 1 = Moderate, 2 = Strict
-            request_params['safesearch'] = request.preferences.get_value('safesearch')
-            request_params['time_range'] = self.time_range
-            request_params['advanced_search'] = self.is_advanced
```
asciimoo, line 305: Why `advanced_search` has removed? It indicates that the advanced search panel should be open.

searx/webapp.py
```diff
@@ -354,6 +354,16 @@ def render(template_name, override_theme=None, **kwargs):
...
 def pre_request():
+    # request.request_data
+    if request.method == 'POST':
+        request_data = request.form
+    elif request.method == 'GET':
+        request_data = request.args
+    else:
+        request_data = {}
+
+    request.request_data = request_data
```
asciimoo: It isn't necessary to select request data here. Below in this function all `GET` arguments are merged with `POST` arguments - see L366 (original code). From that point it is enough to use `request.form` it contains all the client arguments. Maybe the name `request.form` is confusing after the merge and `request.request_data` would be better, but merging arguments allows more flexible usage than selecting only `GET` or `POST` data exclusively.

searx/webapp.py
```diff
@@ -501,30 +516,23 @@ def about():
 @app.route('/autocompleter', methods=['GET', 'POST'])
 def autocompleter():
     """Return autocompleter results"""
-    request_data = {}
-
-    # select request method
-    if request.method == 'POST':
-        request_data = request.form
-    else:
-        request_data = request.args
```
asciimoo, line 520: :+1:

asciimoo (changes requested):

searx/webapp.py
dalf: So I basically search / replace `request.request.request` by `request.form` ? And remove request.request_data initialization.

asciimoo: Yes, using `request.form` is the good practice from that point, it contains all the available arguments after GET/POST merge.

asciimoo (approved):

asciimoo:
great, thanks

PR #735, opened 2016-10-22
[mod] the environment variable SEARX_DEBUG enables debug

dalf:
Instead of modifying settings.yml to enable debug and forget it, this PR can enable debug with the environment variable SEARX_DEBUG.

The value can be 1, 0, true or false (upper or lower case).

I don't know if this is security hole.


asciimoo (approved):

asciimoo:
can be handy on debugging, thanks

PR #739, opened 2016-10-24
Compile jinja template in a /tmp zip file when debug is not activated.

dalf:
It's a note not really to merge

On each search request,  templates are read from the disk (if there are 20 results, "result.html"  is read 20 times).

This PR is just an attempt to reduce the number of read.

Looking at (Ubuntu 14.04)

``` bash
strace -f -e open,close python searx/webapp.py 2> log_strace.txt
cut -d\" -f2 log_strace.txt | sort | uniq -c | sort -nr | grep -v "/usr/lib/python" | grep -v $HOME | grep " /"
```

After one search :

```
     48 /etc/ld.so.cache
     35 /usr/lib/x86_64-linux-gnu/liblzma.so.2
     35 /usr/lib/liblzma.so.2
     35 /lib/x86_64-linux-gnu/liblzma.so.2
     35 /lib/liblzma.so.2
     16 /tmp/searxMglL6W.zip
     10 /etc/hosts
      8 /etc/resolv.conf
      5 /dev/urandom
      3 /lib/x86_64-linux-gnu/libc.so.6
      2 /usr/lib/x86_64-linux-gnu/gconv/gconv-modules.cache
      2 /usr/lib/locale/locale-archive
      1 /usr/share/locale/locale.alias
      1 /usr/share/locale-langpack/fr/LC_MESSAGES/libc.mo
      1 /usr/share/locale-langpack/fr_FR/LC_MESSAGES/libc.mo
      1 /usr/share/locale/fr/LC_MESSAGES/libc.mo
      1 /usr/share/locale/fr_FR/LC_MESSAGES/libc.mo
      1 /usr/lib/x86_64-linux-gnu/x86_64/liblzma.so.2
      1 /usr/lib/x86_64-linux-gnu/tls/x86_64/liblzma.so.2
      1 /usr/lib/x86_64-linux-gnu/tls/liblzma.so.2
      1 /usr/lib/x86_64-linux-gnu/libxslt.so.1
      1 /usr/lib/x86_64-linux-gnu/libxml2.so.2
      1 /usr/lib/x86_64-linux-gnu/libffi.so.6
      1 /usr/lib/x86_64-linux-gnu/libexslt.so.0
      1 /usr/lib/x86_64/liblzma.so.2
      1 /usr/lib/tls/x86_64/liblzma.so.2
      1 /usr/lib/tls/liblzma.so.2
      1 /tmp/IHYngN
      1 /sys/devices/system/cpu/online
      1 /sbin/ldconfig
      1 /proc/sys/vm/overcommit_memory
      1 /proc/self/status
      1 /proc/mounts
      1 /proc/meminfo
      1 /lib/x86_64-linux-gnu/x86_64/liblzma.so.2
      1 /lib/x86_64-linux-gnu/tls/x86_64/liblzma.so.2
      1 /lib/x86_64-linux-gnu/tls/liblzma.so.2
      1 /lib/x86_64-linux-gnu/libz.so.1
      1 /lib/x86_64-linux-gnu/libuuid.so.1
      1 /lib/x86_64-linux-gnu/libutil.so.1
      1 /lib/x86_64-linux-gnu/libssl.so.1.0.0
      1 /lib/x86_64-linux-gnu/libresolv.so.2
      1 /lib/x86_64-linux-gnu/libpthread.so.0
      1 /lib/x86_64-linux-gnu/libnss_myhostname.so.2
      1 /lib/x86_64-linux-gnu/libnss_mdns4_minimal.so.2
      1 /lib/x86_64-linux-gnu/libnss_files.so.2
      1 /lib/x86_64-linux-gnu/libnss_dns.so.2
      1 /lib/x86_64-linux-gnu/libm.so.6
      1 /lib/x86_64-linux-gnu/liblzma.so.5
      1 /lib/x86_64-linux-gnu/libgpg-error.so.0
      1 /lib/x86_64-linux-gnu/libgcrypt.so.11
      1 /lib/x86_64-linux-gnu/libdl.so.2
      1 /lib/x86_64-linux-gnu/libcrypto.so.1.0.0
      1 /lib/x86_64/liblzma.so.2
      1 /lib/tls/x86_64/liblzma.so.2
      1 /lib/tls/liblzma.so.2
      1 /etc/timezone
      1 /etc/nsswitch.conf
      1 /etc/localtime
      1 /etc/host.conf
      1 /etc/gai.conf
      1 /dev/null
```

The two files in /tmp are created by the jinja compilation.


asciimoo:
onice! could you fix the pep8 errors?

dalf:
In the current implementation the temporary file is not deleted.

More over, I think this file should be in the safe place or in memory (I've try to use tempfile.TemporaryFile and tempfile.SpooledTemporaryFile without success)

And I'm not sure if it is more efficient or less (at least it's avoid switch between kernel space, user space).

asciimoo:
`NamedTemporaryFile` has `delete` parameter which deletes the file on close, maybe it works: https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile

dalf:
Thank's it works.
I've difficulties to know if it is really faster. At least the compiled module are loaded only once.
Without this PR, template are loaded each there are used (for example : result.html)

One drawback is the behavior which is different when the debug mode is disabled.

dalf:
With 
```
flask==0.12
Jinja2==2.9.5
```
the templates are not loaded on each request. 

May be once per worker or something similar. 
I think this PR can be closed.

PR #748, opened 2016-11-01
[mod] Allow users to search in most engine supported languages

MarcAbonce:
This PR is meant to provide users a search language list that more accurately reflects the language availability from the underlying engines.
The generated language list is expected to encompass most supported languages, while avoiding redundant entries or mistaken language codes. Most of all, the list is meant to be useful.

Changes:
 * ~~Added list of supported languages for most engines that have language support.~~
 * languages.py is now generated automatically from each engine's supported languages.
 * Many fixes in engines for language support (engines follow different language code standards).
 * Language selector available in advanced settings box (Oscar only).
 * Engine selector in preferences tells user if cookie language is supported by each engine.  

Edit:  
 * Engines can have an optional fetch_supported_languages method that gets the languages from the site/api itself.  
 * The list of supported languages for each engine is currently stored in a json file. This should probably be moved to a [database](https://github.com/asciimoo/searx/wiki/database-structure) eventually, though.  
* utils/fetch_languages.py generates both the json for each engine as well as languages.py. This script is run manually to update both lists. It runs extremely slow, but it doesn't really need to be called often. I think running it once before every major release would be enough (similar to translations in that regard).


asciimoo:
@a01200356 awesome!
I'm thinking on if there is a way of automatic gathering of the engines supported languages, or at least a method to validate our language list against the search service, because now it is relatively hard to detect if a search service's language settings changed.
Not sure it is the scope of this pull request, but it would be good to find a way to test search services in that manner - e.g. in https://github.com/dalf/searx-stats2 ? =]

MarcAbonce:
Supported languages are now read directly from their websites (or apis). I edited the PR description with more information to reflect this changes.

asciimoo:
@a01200356 supercool! i've started the review, looks very promising, thanks!

utils/fetch_languages.py
```diff
@@ -0,0 +1,164 @@
...
+
+engines_languages = {}
+languages = {}
+
+
+# To filter out invalid codes and dialects.
+def valid_code(lang_code):
+    # filter invalid codes
+    # sl-SL is technically not invalid, but still a mistake
+    if lang_code[:2] == 'xx'\
```
kvch, line 37: Could you make this condition less complex? For example you could create lists like `invalid_codes = ['sl-SL', 'wt-WT', 'jw']` and use those in the condition, instead of enumerating all invalid codes.

utils/fetch_languages.py
```diff
@@ -0,0 +1,164 @@
...
+def fetch_supported_languages():
+    for engine_name in engines:
+        if hasattr(engines[engine_name], 'fetch_supported_languages'):
+            try:
+                engines_languages[engine_name] = engines[engine_name].fetch_supported_languages()
+            except Exception as e:
+                print e
+
+    # write json file
+    f = io.open(engines_languages_file, "w", encoding="utf-8")
```
kvch: IMHO using the `with` context is much cleaner.

kvch:
All in all, it is a cool PR. I don't insist on modifying the code based on my notes, because most of it are just nitpicking.

Why are supported languages shown only when the user selected a language? To me it is counter-intuitive. I would have expected the checkboxes of engines that have language support to be checked in case of default search language.

kvch (approved):

asciimoo:
@a01200356, works well and the code is clean and understandable.
Preferences is the only part which is still a bit confusing to me.
If search language is the default, then all the `language support` checkboxes are checked in the engines tab, but if e.g. `Brezhoneg` is selected as language, only dailymotion is checked however it still searches in all engines. Probably a more informative solution would be to display the list of supported languages for all engines as a dropdown or just keep the previous behavior: check the checkbox if the engine supports any language customization.

asciimoo:
Pulling and creating the preferences mod, thank you again the awesome pr

PR #751, opened 2016-11-05
Simplify search.py

dalf:
Basically it's an update of PR #518

The timeouts in settings.yml is about the total time (not only the HTTP request but also the prepare the request and parsing the response)
It was more or less the case before since the threaded_requests function ignores the thread after the timeout even the HTTP request is ended.

New / changed stats :
* page_load_time : record the HTTP request time
* page_load_count: the number of HTTP request
* engine_time : the execution total time of an engine
* engine_time_count : the number of "engine_time" measure

The avg response times in the preferences are the engine response time (engine_load_time / engine_load_count)

To sum up :
* ```Search.search()``` filters the engines that can't process the request
* ```Search.search()``` calls ```search_multiple_requests``` function
* ```search_multiple_requests``` creates one thread per engine, each thread runs the ```search_one_request``` function (it is the ```threaded_requests``` function which has been renamed)
* ```search_one_request``` calls the ```request``` function, makes the HTTP request, calls the ```response``` function, extends the ```result_container```
* ```search_multiple_requests``` waits for the the threads to finish (or timeout)

I don't know if it is involved too much changes.


asciimoo (approved):

asciimoo:
Much cleaner and more understandable! thanks

Issue #709, opened 2016-09-29
feature: add more date search options to combo control


atomGit:
right now it's day, week, month - would like to see 3, 6 and 12 months also

gggirlgeek:
+1. Yes please!

I am surprised that there is not "past year". It is my default search parameter because older information is just too out of date. 24 months would be nice too. Most tech info is still relevant within 2 years.

Meanwhile:  is there a workaround? I tried adding "time_range=year" but the page just reverted back to "Anytime."  Also is there a way to specify a number of months, or even a specific date range?

PR #783, opened 2016-12-11
add year support to engines which support it

kvch:
Please, note that some of the engines only support time range search up to a month. Others which support year are extended with this function.

Engines supporting it: Bing Images, Flickr, Google, Google Images, Google News
Engines not supporting it: Bing News, DeviantArt, DuckDuckGo, Yahoo, YouTube

As only Flickr could be able to support 3 and 6 month ranges with a workaround, these ranges are not introduced in this PR. IMHO searx should not support it, because most of the engines does not even allow it. If adding year to the existing time ranges search functionality is sufficient, #709 can be closed.


asciimoo (approved):

asciimoo:
nice! thanks

asciimoo:
In the future it would be good to support custom time ranges.

PR #786, opened 2016-12-23
Add ccc-tv to searx

kekscode:
This adds https://media.ccc.de to searx.

Tested in docker container. Content extraction works and deeplinks work, too.
Sadly paging is not provided by ccc-tv itself (yet?).


kvch:
Huge :+1: for CCC. :) However, could you squash your commits into a single one? Five commits for an XPATH engine is a lot.

kekscode:
I opened #788 with a single commit history and close this.

PR #788, opened 2016-12-24
Add ccctv engine to settings.yml

kekscode:
This adds https://media.ccc.de to searx.

Tested in docker container. Content extraction works and deeplinks work, too.
Sadly paging is not provided by ccc-tv itself (yet?).

This supersedes pull request #786 which had unnecessary commits.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -60,3 +60,4 @@ generally made searx better:
 - Thomas Renard @threnard
 - Pydo `<https://github.com/pydo>`_
 - Athemis `<https://github.com/Athemis>`_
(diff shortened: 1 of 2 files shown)


kvch (approved):

asciimoo:
Thanks for the great source.
I think disabling it by default would be better (`disabled : True`), because it isn't a general purpose video search service

kekscode:
As suggested: `disabled : True` is added to the configuration.

asciimoo (approved):

asciimoo:
Awesome, thanks

PR #794, opened 2016-12-29
Changes in search language list (continuation of #748)

MarcAbonce:
This is a continuation of my previous PR (#748), based on the last comment there.
There are mainly two things that I changed here: The list of selectable languages and also I reverted the column in preferences.

Originally, the language list includes any language that at least one engine supports. But then selecting a language like Breton is pretty useless if it's only supported on one engine. Displaying this barely supported languages may even be a bit deceiving.
So maybe instead of displaying all languages supported by at least one engine, we could only display those languages which are supported by a minimum amount of engines, so that we can at least ensure that selecting any of the available languages will give some results in that language.
Other languages can still be queried through the : bang as long as it looks like a valid language code.

About the column in preferences, I think it makes more sense to tell the user if *their* language is supported by that engine. I don't think the user cares at all if Yandex, for example, allows you to switch between Slavic languages if that user understands none. I changed the column name so that hopefully it makes more sense at first sight what it means.


asciimoo (approved):

asciimoo:
agree, thanks

dalf:
What if the user changes the selected language in the General tab, and then switches to the Engine tab ?

PR #797, opened 2016-12-29
Engine can be offline and has a full access to the search_query object.

dalf:
It would allow sympy to be an engine instead of a plugin. Since it would be thread, it would be more easier to deal with timeout.

The full access to the search_query object would allow translation by specifying two languages in the request for example. 

Changes : 
* an engine can define the ```search(search_query, selected_categories, start_time, timeout_limit)``` function. If not the default behavior doesn't change, see ```get_default_search function``` in  ```searx/engines/__base__.py```   
* an engine can define the ```can_accept(search_query)``` function. It would allow the sympy engine to accept or not the request. The default behavior doesn't change, see ```get_default_can_accept``` function.

Side effects : 
* For one user request, multiple HTTP were done using the same user agent. It's no more the case with this PR.
* About the global timeout in the ```send_http_request``` function ( ```if search_duration > timeout_limit + timeout_overhead``` ) : before, the ```engine.continuous_errors``` was reseted, it's not the case anymore. It can be reverted.

Notes : 
* there are three commits, to show the incremental changes.
* the engine.continuous_errors is set inside ```searx/search.py``` and ```searx/engines/__base__py``` not very nice. They can be squashed.

### DIFF
--- a/searx/engines/__base__.py
+++ b/searx/engines/__base__.py
@@ -1,0 +1,94 @@
+import threading
+import requests.exceptions
+import searx.poolrequests as requests_lib
+from time import time
+from searx import logger
+
+logger = logger.getChild('engine')
+
+
+def send_http_request(engine, request_params, timeout_limit):
+    # for page_load_time stats
+    time_before_request = time()
+
+    # create dictionary which contain all
+    # informations about the request
+    request_args = dict(
+        headers=request_params['headers'],
+        cookies=request_params['cookies'],
+        timeout=timeout_limit,
+        verify=request_params['verify']
+    )
+
+    # specific type of request (GET or POST)
+    if request_params['method'] == 'GET':
+        req = requests_lib.get
+    else:
+        req = requests_lib.post
+        request_args['data'] = request_params['data']
+
+    # send the request
+    response = req(request_params['url'], **request_args)
+
+    # is there a timeout (no parsing in this case)
+    timeout_overhead = 0.2  # seconds
+    search_duration = time() - request_params['started']
+    if search_duration > timeout_limit + timeout_overhead:
+        raise requests.exceptions.Timeout(response=response)
+
+    with threading.RLock():
+        # no error : reset the suspend variables
+        engine.continuous_errors = 0
+        engine.suspend_end_time = 0
+        # update stats with current page-load-time
+        # only the HTTP request
+        engine.stats['page_load_time'] += time() - time_before_request
+        engine.stats['page_load_count'] += 1
+
+    # everything is ok : return the response
+    return response
+
+
+def get_default_search(engine):
+
+    def search(query, request_params, timeout_limit):
+        # update request parameters dependent on
+        # search-engine (contained in engines folder)
+        engine.request(query, request_params)
+
+        # check if there is a HTTP request
+        if request_params['url'] is None:
+            return []
+
+        if not request_params['url']:
+            return []
+
+        # send request
+        response = send_http_request(engine, request_params, timeout_limit)
+
+        # parse the response
+        response.search_params = request_params
+        return engine.response(response)
+
+    return search
+
+
+def get_default_can_accept_search_query(engine):
+
+    def can_accept_search_query(search_query):
+        # if paging is not supported, skip
+        if search_query.pageno > 1 and not engine.paging:
+            return False
+
+        # if search-language is set and engine does not
+        # provide language-support, skip
+        if search_query.lang != 'all' and not engine.language_support:
+            return False
+
+        # if time_range is not supported, skip
+        if search_query.time_range and not engine.time_range_support:
+            return False
+
+        return True
+
+    return can_accept_search_query
--- a/searx/engines/__base__.py
+++ b/searx/engines/__base__.py
@@ -2,12 +2,13 @@ import threading
 import requests.exceptions
 import searx.poolrequests as requests_lib
 from time import time
+from searx.utils import gen_useragent
 from searx import logger
 
 logger = logger.getChild('engine')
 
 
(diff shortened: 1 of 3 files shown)


asciimoo:
@dalf answerers are implemented for exactly this purpose: https://github.com/asciimoo/searx/tree/master/searx/answerers (not documented yet)
I think it is a cleaner interface for offline calculations. What are the benefits of the offline engines compared to the answerers?

dalf:
I know it may be not a need, or just out of the scope of searx.

Some purposes :
* Implement an engine that look up results into anything that is not HTTP : news group, a local database.
* It use thread as the others engines with timeout (I think about Sympy).

@asciimoo, whatever, what do you think this commit : https://github.com/asciimoo/searx/pull/797/commits/12b077c8fbe49f329d4b4c792fbaef7e56315cf6

asciimoo:
> Implement an engine that look up results into anything that is not HTTP : news group, a local database.

This is how answerers should work =]

> It use thread as the others engines with timeout (I think about Sympy).

Answerers can be threaded too, but it still does not solve the main problem, that the thread cannot be killed on the given timeout and eats all the cpu times after a while.

> what do you think this commit : 12b077c

Cool! Agree with the changes and the small refactor makes the code more modular and understandable to me.

dalf:
Sorry for my dumb answer.

answerers doesn't mix with other engines : when an answerer responds to a query, all other engines are ignored. I think the purpose of answerers is respond to specific requests (math, get the color from hex code, duckduckgo has a bunch of that).

The purpose here is more about an usual/normal engine but doing a request that is not a HTTP request  : querying a SQL server / NoSQL server for example. Example : a list of addresses that can used in the map category. In this case, I think the user still want to results from photon and OSM.

I know this PR doesn't solve the following problem, but when there is more than one HTTP request to get the results : why searx blocks the engine if there is an HTTP error on the first request, but not on the second one (done in the response function of the engine without the right user agent, without time-out). 

Example : https://github.com/asciimoo/searx/blob/master/searx/engines/wikidata.py#L83
Actually, in this case, it would super cool to use again the  search_multiple_requests  function. But I agree the current implementation is fine in nearly of the cases (I mean, preparing the HTTP request, sending the request, parsing the result).

I know that a thread can't be killed, but process can be (not sure if SymPy haven't issue about that) : it's a first stone to implement this. Only an idea.

But honestly it was more a try than anything else.

I hope it's more clear.

asciimoo:
@dalf thanks the explanation.

> answerers doesn't mix with other engines

I implemented this way for privacy reasons. I thought it is a good idea to stop if there is an answer already from the answerers - so data leakage to external resources can be prevented.
As a solution, we can add a flag to the answerers which could indicate if the search can be stopped or not.
(If no results provided by the answerers, the search will continue in the current code too, which can allow (pre)computation that can be used by the engines)

> The purpose here is more about an usual/normal engine but doing a request that is not a HTTP request : querying a SQL server / NoSQL server for example. Example : a list of addresses that can used in the map category. In this case, I think the user still want to results from photon and OSM.

This is more like a scheduled/periodical event than a user triggered if i understand correctly.

> I know that a thread can't be killed, but process can be (not sure if SymPy haven't issue about that) : it's a first stone to implement this. Only an idea.

You are right, it could be a child process terminated on search timeout.

> I know this PR doesn't solve the following problem, but when there is more than one HTTP request to get the results : why searx blocks the engine if there is an HTTP error on the first request, but not on the second one (done in the response function of the engine without the right user agent, without time-out).

I agree, it would be good to parallelize those requests.

dalf:
> This is more like a scheduled/periodical event than a user triggered if i understand correctly.

Use case : in the map category, the user searches for "Riviere Du Loup", and results comes from from OSM, Photon, and airport_offline engine.

The airport_offline engine queries an database (the administrator has to setup the database on the searx instance or not far).
The data are loaded from https://github.com/jpatokal/openflights/tree/master/data
(see airports.dat file).

So, doing this, searx becomes an search engine without the meta since it starts to host data.

In the general category, the engine can suggest "Riviere Du Loup" when the user searches for "CYRI" (the airport code). Actually it's what wikipedia engine does when the english language is selected, not in other languages.

Side note : If the user searches for "weather CYRI", an answer can be provided, but the user may be interested in the usual results.

I admit that :
* this is most probably out of the scope of searx and / or useless (just a little experiment)
* the use is most probably to implement infoboxes or answers in an asynchronous way (I understand now the privacy choice for the answers).

asciimoo:
> Use case : in the map category, the user searches for "Riviere Du Loup", and results comes from from OSM, Photon, and airport_offline engine.

Ah, okay. Then, i think we are talking about the same thing with a different terminology (offline engine vs answerer).  My goal with answerers was exactly the same: create offline result providers which can replace online ones.

It is true, sometimes it's not enough to only replace the results, answerers should handle this too.

> this is most probably out of the scope of searx and / or useless (just a little experiment)

I don't think so. It can strengthen privacy and give more useful features and another level of customization.
That's why i created a dedicated module for it, instead of polluting the engines with another slightly different use case.
I admit, answerers are still very basic and they need to be more polished, but in long term if the number and the size of the answerers/offline-engines grows,  the separation (at least in code level) has advantages.
With the current design of answerers (one folder for each) it is possible, to include external answerers as submodules and if one of the answerer's code grows, it has it's own directory. It would be harder to implement these to engines.

dalf:
I just open the PR #798 with only the first commit of this PR. 

@asciimoo, feel free to close, leave it in stand by for the future / as note / reference.

dalf:
See commit a6f20caf32af463b57a026ee7cb7ed6317db6b8b

PR #798, opened 2016-12-30
[mod] add a search_one_request_safe function wrapper

dalf:
Add a search_one_request_safe function wrapper to call search_one_request. 
All exceptions are catched in this wrapper.

Note that ```if search_duration > timeout_limit + timeout_overhead``` check is done before setting ```engine.continuous_errors = 0```.

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -20,6 +20,7 @@ import threading
 from thread import start_new_thread
 from time import time
 from uuid import uuid4
+import requests.exceptions
 import searx.poolrequests as requests_lib
 from searx.engines import (
     categories, engines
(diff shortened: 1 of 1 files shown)


asciimoo (approved):

asciimoo:
Thank you

asciimoo:
@dalf how can cause problem the mentioned search duration check?

dalf:
Not sure if I understand correctly your question. 

Without this commit : 
* if an engine get a timeout with Requests, it will be suspended.
* if an engine get a timeout with searx (the ```if search_duration > timeout_limit + timeout_overhead``` check), the engine will not be suspended

With this commit, in both case the engine will be suspended. 

Actually the ```if search_duration > timeout_limit + timeout_overhead``` check could be part of poolrequests.py. The purpose is to have a global timeout, not a timeout for the connect, and another one for the read.

asciimoo:
ah, i see, great, thanks.
Global+read timeout sounds reasonable to me.

PR #801, opened 2016-12-31
[fix] extract_text: use html.tostring instead html_to_text.

### LINKED ISSUE #711: Tags <xml> are hidden from result titles

dalf:
I'm not sure about this PR. 

What does ```html_to_text``` more than this new code ?
Should the encoding be unicode ?

Fix #711


asciimoo (approved):

asciimoo:
I don't think extract text does anything else, and it works fine with unicode characters (tested with `!gos 没有那个文件或目录`).
Thanks!

Issue #720, opened 2016-10-05
swisscows doesn't work


ghost:
I tried using the swisscows search engine on several different instances including https://searx.me but results aren't being fetched.

devonhk:
Just tested locally can confirm. The regex parsing string is what seems to be broken. I'll see about making a PR to fix it later.

devonhk:
So I found the swisscows endpoint that returns the search results.
`https://swisscows.ch/?query={query}+hundefutter&region=de-CH&uiLanguage=browser&_={Date}&apiGuard={token}'`

The `Date` is a derived from javascript `new Date().valueOf()`.
The api token comes from making a `GET` request to `https://swisscows.ch/generateApiGuardToken`,

Here's some example code to retrieve the search results:

```
import requests
import time
import datetime

resp = requests.get('https://swisscows.ch/generateApiGuardToken')
api_token = resp.json()['token']
time_stamp = time.mktime(datetime.datetime.now().timetuple()) * 1000
headers = {'X-Requested-With': 'XMLHttpRequest', 'Cookie': 'hash={}'.format(resp.cookies['hash'])}
resp = requests.get('https://swisscows.ch/?query=dogdog+hundefutter&region=de-CH&uiLanguage=browser&_={}&apiGuard={}'.format(time_stamp, api_token), headers=headers)
print(resp.status_code)
print(resp.json())
```

This problem is the request only works about 1 out of 10 times. Most of the time the 2nd request just returns a 403. 
If anyone has any ideas on how to consistently call this endpoint so it returns 200 reponses please let me know. If not I suggest we drop support for **swisscows**.

dalf:
The swisscow engine is not supported.
See https://github.com/searx/searx/issues/1859#issuecomment-589952307

HechtYehuda:
Was this ever addressed? I'm experiencing similar issues now. Most searches return "Server doesn’t respond or something else happened."

PR #809, opened 2017-01-03
[fix] remove the swisscow engine (see #720)

dalf:


asciimoo:
It would be good to fix it instead of remove. I figured out that the request requires an `apiGuard` parameter which can be obtained here: https://hulbee.com/generateApiGuardToken. As I see you can do multiple requests with one `apiGuard` token, but not sure if it has time or any other limits.

UPDATE:
Token expires after 1-2hrs

PR #811, opened 2017-01-03
[mod] ./manage.sh can download geckodriver

dalf:
```./manage.sh check_geckodriver``` : 
* checks if geckdriver is installed anywhere in the PATH.
* if not, get the URL for the right arch from github
* download it and install the binary into the virtual environment OR just give the URL to download if there is no virtual envionment

```./manage.sh tests``` calls check_geckodriver to make sure that the geckodriver is installed.


asciimoo (approved):

asciimoo:
@dalf nice! thanks

Issue #815, opened 2017-01-05
Please allow installing static files and configuration to a different path


josch:
Hi,

currently, setuptools will install everything into a path like `/usr/lib/python3/dist-packages/searx` on the system where searx gets installed. Unfortunately some of the content that thus ends up in `/usr/lib` then violates the Filesystem Hierarchy Standard. Thus, in Debian, we must put some files into different locations, like:

 - putting `README.rst` into `/usr/share/doc/searx`
 - putting `searx/static` into `/usr/share/searx` and adding a symlink to that into `/usr/lib/python3/dist-packages/searx`
 - putting `searx/settings.yml` into `/etc/searx` and using `SEARX_SETTINGS_PATH` when executing searx

While I could now maintain a patch in Debian that does all this, it would be easier for me if you could integrate at least some of that functionality on your side. That would then not require me to rebase my patch upon every release you make. There are different things you can do:

 - allow installing data files and configuration files into a custom directory through setuptools
 - make the path to the data files configurable either through a variable in the code that we can patch in Debian or through an environment variable

In the best case of course, I do not need to carry any patch against searx on the Debian site but I wanted to raise the issue here to see what your position on the issue is. Depending on what you are willing to do, I can then provide you with a patch.

Thanks!

cheers, josch

josch:
Let me propose the following patch:

```patch
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -181,15 +181,24 @@ class UnicodeWriter:
 def get_themes(root):
     """Returns available themes list."""
 
-    static_path = os.path.join(root, 'static')
-    templates_path = os.path.join(root, 'templates')
+    if 'SEARX_STATIC_PATH' in os.environ:
+        static_path = os.environ['SEARX_STATIC_PATH']
+    else:
+        static_path = os.path.join(root, 'static')
+    if 'SEARX_TEMPLATES_PATH' in os.environ:
+        templates_path = os.environ['SEARX_TEMPLATES_PATH']
+    else:
+        templates_path = os.path.join(root, 'templates')
 
     themes = os.listdir(os.path.join(static_path, 'themes'))
     return static_path, templates_path, themes
 
 
 def get_static_files(base_path):
-    base_path = os.path.join(base_path, 'static')
+    if 'SEARX_STATIC_PATH' in os.environ:
+        base_path = os.environ['SEARX_STATIC_PATH']
+    else:
+        base_path = os.path.join(base_path, 'static')
     static_files = set()
[...]

PR #816, opened 2017-01-06
[mod] the static and templates directories can be defined in the settings.yml

dalf:
Add ```static_path``` and ```templates_path``` settings as asked by #815

The ```themes_path``` setting is removed.

Note about previous implementation : the ```themes_path``` was not used for ```global_favicons``` according [this line](https://github.com/asciimoo/searx/blob/master/searx/webapp.py#L118) : 
```python
 theme_img_path = searx_dir + "/static/themes/" + theme + "/img/icons/"
```

Moreover, there is a second commit which allow settings.yml to be loaded from /etc/searx/settings.yml

Searx looks for settings.yml in this order:
* from SEARX_SETTINGS_PATH if defined (if the environment variable reference an file that doesn't exist an exception is raised)
* if not found then from searx code base,
* if not found then from /etc/searx/settings.yml
* if not found an exception stops searx loading

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -18,7 +18,7 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 import certifi
 import logging
 from os import environ
-from os.path import realpath, dirname, join, abspath
+from os.path import realpath, dirname, join, abspath, isfile
 from ssl import OPENSSL_VERSION_INFO, OPENSSL_VERSION
 try:
     from yaml import load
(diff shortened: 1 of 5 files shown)


asciimoo:
@dalf very cool!
It would be good to wait for the py3 branch merge with this, because debian packaging will use that code. I've finished with the rebase of the py3 branch to the current master, but there still can be bugs both in py2/py3 side.

asciimoo:
@dalf could you rebase and resolve the conflicts?

asciimoo (approved):

asciimoo:
Great, thank you.

PR #817, opened 2017-01-06
Remove pyasn1-modules from dependencies

dalf:
According to https://github.com/kennethreitz/requests/issues/749#issuecomment-19187417 there is no need to install pyasn1-modules.


asciimoo (approved):

asciimoo:
thanks

PR #826, opened 2017-01-13
[mod] add a __common__ template

dalf:
this template can't be selected but it provides a common place for shared templates.

What has been moved into this template / directory :
* opensearch*.xml is always the same whatever the themes
* the text inside */about.html

Purpose : avoid duplicate by hand with possible error (in the master branch, pix-art theme haven't opensearch*.xml files)

Possible extension : if statistics are improved, one or more template can be stored into this directory.


asciimoo (approved):

asciimoo:
Onice! Very-very useful PR. It drastically reduces template code duplication and makes the about page translation easier. However travis fails - seems both `opensearch.xml` and `opensearch_response_rss.xml` are missing. Could you fix it?

PR #831, opened 2017-01-17
[WIP] Add suffix compatibility

johackim:
Force engine with a prefix works perfectly (example: `!yt matrix`) but not suffix (example: `matrix !yt`)

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -93,6 +93,14 @@ class RawTextQuery(object):
                         if lang == lang_id:
                             break
 
+            # this force a engine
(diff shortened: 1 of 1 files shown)


asciimoo:
It would be a useful feature indeed.
This part of the original code is a bit messy and hard to understand. It would be good to refactor.
And the PR does not cover all the custom query syntax types, just the engine shortcuts. Allowing position independent engine/category name and language setting would be useful too, i think.
So, I'd keep this PR open while the above issues are not resolved.

PR #832, opened 2017-01-20
Input check

dalf:
supersede #826 and add input validation.

The safesearch parameter can be set in the query URL.

[WIP] I think it should be added to the generated URL of the json/rss/csv format.


asciimoo (approved):

asciimoo:
awesome, thanks

PR #845, opened 2017-01-28
Remove secret key

### LINKED ISSUE #834: What’s the secret key used for?
secret_key : used for cryptography purpose

What is a “cryptography purpose”?

Profpatsch:
Instead of having to “spoil” the config with a secret key (which hinders
uploading to a public repo), we can generate it on application start and put it
into the user cache directory (which should only be readable by the searx user).

Adds `utils.get_secret_app_key` which generates a key if none exists and returns
the existing one otherwise (declarative style).
Also test every codepath of that function.

- [x] Tests run w/o errors
- [x] proxying works with the new generated key

fixes https://github.com/asciimoo/searx/issues/834

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -28,6 +28,7 @@ import hmac
 import json
 import os
 import requests
+import xdg
 
 from searx import logger
 logger = logger.getChild('webapp')
@@ -59,7 +60,7 @@ from searx.engines import (
 from searx.utils import (
     UnicodeWriter, highlight_content, html_to_text, get_themes,
     get_static_files, get_result_templates, gen_useragent, dict_subset,
-    prettify_url
+    prettify_url, get_secret_app_key
 )
 from searx.version import VERSION_STRING
 from searx.languages import language_codes
@@ -103,7 +104,7 @@ app = Flask(
 
 app.jinja_env.trim_blocks = True
 app.jinja_env.lstrip_blocks = True
-app.secret_key = settings['server']['secret_key']
+app.secret_key = get_secret_app_key()
 
 if not searx_debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
     initialize_engines(settings['engines'])
@@ -280,7 +281,7 @@ def image_proxify(url):
     if settings.get('result_proxy'):
         return proxify(url)
 
-    h = hmac.new(settings['server']['secret_key'], url.encode('utf-8'), hashlib.sha256).hexdigest()
+    h = hmac.new(app.secret_key, url.encode('utf-8'), hashlib.sha256).hexdigest()
 
     return '{0}?{1}'.format(url_for('image_proxy'),
                             urlencode(dict(url=url.encode('utf-8'), h=h)))
@@ -684,7 +685,7 @@ def image_proxy():
     if not url:
         return '', 400
 
-    h = hmac.new(settings['server']['secret_key'], url, hashlib.sha256).hexdigest()
+    h = hmac.new(app.secret_key, url, hashlib.sha256).hexdigest()
 
     if h != request.args.get('h'):
         return '', 400
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -104,6 +104,10 @@ app = Flask(
 
 app.jinja_env.trim_blocks = True
 app.jinja_env.lstrip_blocks = True
+
+# notify the user that the secret_key is no longer used
+if 'secret_key' in settings['server']:
+    logger.warning(' The "secret_key" config key is no longer used.')
 app.secret_key = get_secret_app_key()
 
 if not searx_debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
--- a/README.rst
+++ b/README.rst
@@ -18,8 +18,7 @@ Installation
    ``git clone https://github.com/asciimoo/searx.git && cd searx``
 -  install dependencies: ``./manage.sh update_packages``
 -  edit your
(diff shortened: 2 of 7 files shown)


searx/webapp.py
```diff
@@ -265,7 +270,7 @@ def proxify(url):
                                            url.encode('utf-8'),
                                            hashlib.sha256).hexdigest()
 
-    return '{0}?{1}'.format(settings['result_proxy']['url'],
+    return '{0}?{1}'.format(settings['re sult_proxy']['url'],
```
dalf: There is an extra space I guess

asciimoo:
I feel a bit overkill to introduce a new dependency for this feature.
Wouldn't be enough to add the ability to use environment variable for `secret_key` which overwrites the value defined in `settings.yml`? You can store it in a file and source it before starting searx.

Profpatsch:
> which overwrites the value defined in settings.yml

The whole point is to remove the  `secret_key`, since it’s sensitive data in the config which is completely unnecessary.

We could of course skip the dependency and implement this part of xdg-directory manually, but I don’t see what harm a pure python implementation of xdg does.

searx/webapp.py
Profpatsch: Right, that slipped in.

Profpatsch: I rebased the line, it should be correct now.

kvch:
I have multiple problems with this PR:

1. You submitted multiple things in a single PR: fixing dependency versions and adding a new feature.

2. You overcomplicated it. Publishing secret keys could be avoidable by setting it as an environment variable. This solution requires less change in the current behaviour then yours. The result would be the same. Security would not be a problem as you install/run searx as a user. There is no need to change `settings.yml`, adding a new dependency, generating a secret key. adding new error classes and so on.

3, Introduction of a new dependency. Not only lengthens the installation of searx, but makes the code harder to maintain, because if the API changes, it should be followed up.

So I would not merge this PR at all, as there is a simpler solutions to this problem.

kvch:
I have multiple problems with this PR:

0. Travis fails.

1. You submitted multiple things in a single PR: fixing dependency versions and adding a new feature.

2. You overcomplicated it. Publishing secret keys could be avoidable by setting it as an environment variable. This solution requires less change in the current behaviour then yours. The result would be the same. Security would not be a problem as you install/run searx as a user. There is no need to change `settings.yml`, adding a new dependency, generating a secret key. adding new error classes and so on.

3. Introduction of a new dependency. Not only lengthens the installation of searx, but makes the code harder to maintain, because if the API changes, it should be followed up.

So I would not merge this PR at all, as there is a simpler solutions to this problem.

Profpatsch:
> 0. Travis fails.

There were minor pep warnings, which I fixed.



> You submitted multiple things in a single PR

I can create a new PR for the dependencies if it is wanted.

> setting it as an environment variable. This solution requires less change in the current behaviour then yours. The result would be the same.

That is not a correct.
There is no change in the practical behaviour, as the config variable is superfluous in any case, since the user cannot influence much by setting it manually. It could even be changed to not persist the key and just create one on program start, which would simplify the patch considerably actually. The new dependency could also be removed again.

kvch:
> That is not a correct.
> There is no change in the practical behaviour, as the config variable is superfluous in any case, since the user cannot influence much by setting it manually. 

Generating a secret key automatically is a huge change compared to the current behaviour, because right now it is done manually by the user. In the solution described above secret key is still set manually by the user, so it requires less change. Your reason is partly true, the line in the `settings.yml` would become unnecessary, but the user would be able to set it.

Your solution fails in case of load balanced searx instances. All of the running searx server would have different secret key. So an URL of a proxified image generated by an instance behind LB, could only be served by the instance which generated it. Thus, if you sent the URL to your friend who is served by a different running server, could not open the image, as that server had a different secret key. While having the secret key set in an environment variable or in `settings.yml` would enable all of the instances to use a single secret key.

Profpatsch:
The load-balancing aspect of a bigger setup is very valid, yes. The example of sharing links with friends reminds me of Windows working groups.

What do you think of making local keys (different one for each run, so it doesn’t have to touch the file system) the default case, optionally grabbing a provided key from the environment?
I still think it should be removed from the config if possible.

asciimoo:
@Profpatsch I still can't see the benefit of removing the secret key from `settings.yml` and disallowing the option of setting the secret key there.

If you place the key to another file, searx needs read permission for that file too. What extra safety is provided by this solution compared to the optional redefinition of settings.yml's key with an environment variable (which is ~5 lines of straight-forward code and backward compatible)?

virtadpt:
It looks as if this would make life much more difficult for those of us who use Searx for its API functionality only.

Profpatsch:
> It looks as if this would make life much more difficult for those of us who use Searx for its API functionality only.

Care to elaborate?

virtadpt:
I wrote a number of interactive bots that use Searx's API to issue search requests.  Having the secret key in a configuration file means it's static, which means that I can put it into the configuration files of my bots.  Making it dynamic and cached makes that more difficult to handle.

While this would be merely annoying to me, there are other people out there who are doing much the same thing in more constrained environments, and they won't be able to read the cached file on disk to get the secret key.  This breaks things completely for them.

r14c:
i would never want the application to generate its own secret key, that's definitely something that should be handled explicitly by the operator or some automation system. secret management is **not** the responsibility of the application. using environment variables to define secrets is a much cleaner solution imo. 

my searx cluster uses a config file that is generated with confd, but i'd be happy if i could set the secrets directly with environment variables.

PR #851, opened 2017-02-02
[RFR] Fix dockerfile

### LINKED ISSUE #850: Docker Errors
I did this 

run -d --name searx -p 8888:8888 -e IMAGE_PROXY=True searx

and received this error message 

docker: Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "exec: \"/usr/bin/tini\": stat /usr/bin/tini: no such file or directory".

johackim:
Fix #850

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.3
+FROM alpine:3.4
 MAINTAINER searx <https://github.com/asciimoo/searx>
 LABEL description "A privacy-respecting, hackable metasearch engine."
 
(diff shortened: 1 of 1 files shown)


Wonderfall (commented):
tini moved its binary from `/usr/bin/tini` to `/sbin/tini`. I don't know if it's the case with alpine 3.4 (at least it was displaying a warning if I recall correctly), but it is with alpine 3.5 and will be for future versions. That was just the problem, I suggest this to be fixed, so you can even use alpine 3.5 instead which is the latest stable version.

If needed, you can take a look at mine : https://github.com/Wonderfall/dockerfiles/tree/master/searx

PR #853, opened 2017-02-04
Add my ip answer

ghost:
I added an answer that retrieves the public IP and information of the current client, since the plain "ip" command only returns the user's LAN IP. The command also returns a general location and an ISP.


dalf:
I'm not sure to understand. With 
```python
ip_info = get('http://ip-api.com/json', proxies=outgoing_proxies).json()
```
```ip_info``` contains the IP of the searx instance. Of course if the searx instance is run locally, you get your IP address.

**Comment by ghost** [comment]
You are actually right about that; It contains the instance IP. I was mostly thinking in terms of when you run it on your own or through a Tor network, so that you can have visibility over what identity you are given. Perhaps the plugin should be renamed to `instanceip`? Unless there is an elegant way of passing the `request.remote_addr` from Flask into the plugin, so that it could be passed to the ip-api request.

**Comment by dalf** [comment]
Tor network allows to have hidden servers : the IPs of the servers are hidden. 
Why display the instance IP in this case ?

**Comment by ghost** [comment]
Well, the [Tor exit nodes are known and are public](http://iplists.firehol.org/?ipset=tor_exits). The thing is that no one - that is on the other side - knows who's connected to them and this would allow one to know their current (disposable) identity, since they switch every so often. It may be useless, but I found it useful while using it on a Raspberry Pi at home. However, this could still be used if `request.remote_addr` could be passed as a parameter in the GET request, so that the user's IP info would be found.

**Comment by dalf** [comment]
> However, this could still be used if request.remote_addr could be passed as a parameter in the GET request, so that the user's IP info would be found.

https://github.com/asciimoo/searx/blob/master/searx/plugins/self_info.py#L39 ?

**Comment by ghost** [comment]
Ok, I'll create a separate pull request for that. Feel free to close this pull request if you don't see a use case for this plugin (I had seen that line, but was unsure what that was used for).

## Edit

I'll keep this PR and reiterate.

**Comment by kvch** [comment]
I know that I am late to the game. This is a cool feature, however it mustn't be an answerer. Answerers are are offline and don't make any requests to external services. @conmarap could you please implement it as an engine or plugin?

**Comment by ghost** [comment]
Absolutely @kvch !

**Comment by ghost** [comment]
I tested with `./manage.sh tests` locally. Not sure why it's failing, but I'll figure it out. In the meantime, @kvch I made the change that you requested and it looks something like this:

image

File: searx/plugins/self_info.py [hunk 1]
```diff
@@ -25,20 +29,42 @@
...
+
+    if ip == '127.0.0.1':
+        ip = ''
+
+    # Find the outgoing proxies settings from the user configuration.
+    outgoing_proxies = settings['outgoing'].get('proxies', None)
+
+    # Initiate a GET request and set the outgoing proxies, if any were set in
+    # settings.yml.
+    ip_info = get('http://ip-api.com/json/' + ip, proxies=outgoing_proxies).json()
```
dalf: HTTP request without TLS : the IP is in clear between the searx instance and ip-api.com server(s). Unfortunately ip-api.com doesn't seem to support HTTPS requests.

ghost: Any suggestions?

ghost: Maxmind's datasets are freely available, but they need an attribution link. So this could be handled internally, maybe? http://dev.maxmind.com/geoip/legacy/geolite/, otherwise I've set up a service on Heroku with HTTPS support https://geoipnode.herokuapp.com/json/, but doesn't provide any ISP information.

dalf: Seeing
* http://dev.maxmind.com/geoip/legacy/geolite/  (the new format has far less information)
* https://github.com/appliedsec/pygeoip/blob/master/docs/getting-started.rst

I would say that searx can handle that internally ? 
It would make sure that the IP is not spread. But when to download the database, when to update ?

Otherwise there is https://ipinfo.io/ (a comparaison tool : https://www.iplocation.net/ hoping it is fair).

ghost: Ok, I can change it to ipinfo.io for now, and look into implementing it internally. The Maxmind datafiles have plenty of information, the same that ipinfo.io provides, so it should be fine.

searx/plugins/self_info.py
```diff
@@ -46,14 +46,14 @@ def get_ip(request):
 
     # Initiate a GET request and set the outgoing proxies, if any were set in
     # settings.yml.
-    ip_info = get('http://ip-api.com/json/' + ip, proxies=outgoing_proxies).json()
+    ip_info = get('http://ipinfo.io/' + (ip + '/' if ip else '') + 'json/', proxies=outgoing_proxies).json()
```
dalf: I think a "s" in missing :-p

ghost: Haha, yes it is. I'll take care of it. I was mostly testing and wanted to see why the assertion test is failing on travis ci.

ghost:
ping

dalf:
@conmarap sorry the late answer. There are two issues : 
* since the code sends a HTTP request, it should be converted to an engine similar to the currency engine.
* could you merge all the commits ?

ghost:
No problem, @dalf! 😄 I actually changed it to https in cec07cf80ac3325b98f40d8fe7eb8e6d9a7f9161 (edit: I got what you mean after I read the #903), but I can turn it into an engine if that's what you'd like. What do you mean by merging the commits though?

dalf:
Yep, got it for the https : good.

There are 15 commits : only one should be enough.

ghost:
I'll resubmit.

Issue #828, opened 2017-01-15
Please consider adding a minimal theme that saves bandwidth


josch:
The default theme (oscar) is unbearably slow with my internet connection (128 kbit/s). Loading only the front page takes 106 seconds. To fix this, I implemented a new theme which does not need css and javascript and does not load any images or icons.

To illustrate my situation, I took a screenshot of the firebug network graph:

![firebug output](https://mister-muffin.de/p/EWvO.png)

As you can see, nearly a third of a megabyte has to be downloaded only to display the front page.

And the above output is even a best-case scenario. Sometimes (probably since all resources are loaded at once) some javascript or css files will not be loaded at all because the request would timeout.

Yet another reason for my efforts are costs of mobile internet. When I'm abroad I easily pay 15 EUR per 10 MB. I don't want searx to already take up 0.3 MB of that...

Thus, I created a more minimal theme. Basically I just took oscar and removed the css, javascript and images from it. With that theme, the front page requires only 1.1 KB to be downloaded which can be done in under a second.

I just would like to ask you: would you appreciate adding a minimal theme like this?

I thought of adding certain styles to the theme to adjust how minimal the user wants it. For example: also disable images in the image search or the map preview in the map search. Or maybe allow some very minimal CSS to make things look better. Or maybe even allow some very minimal javascript for some very basic usability features.

For this to work better, I also have two wishlist bugs:

It would be nice if the theme could be chosen via the url. Otherwise, whenever I change to my slow internet connection, I first have to wait painfully long until the preferences page loaded which allows me to change the theme from oscar to the minimal one. Allowing something like `?theme=foo` in the url would be nice.

Secondly, the list of available themes is currently determined by the existance of a directory with the theme name in `searx/static/themes`. But that doesn't make any sense for the minimal theme because it ships nothing in that directory. It would make more sense to check `searx/templates` for the list of available themes because there cannot be a theme without a template, right?

You can grab the current status of my template from here as a gzipped tarball:

https://mister-muffin.de/p/Ss7y.gz

It follows two screenshots of the search results (top and bottom). As you can see, everything is still nicely usable.

![top](https://mister-muffin.de/p/3noV.png)
![bottom](https://mister-muffin.de/p/dAwf.png)

kvch:
Just being curious... Why is the legacy theme not sufficient?

josch:
[...]

Issue #687, opened 2016-09-07
[task] Abandon bootstrap


GreenLunar:
By my opinion. it is too big, wasteful, and unnecessary.  It does not offer anything special aside from having even worsen and weaken and more dependent web designers.

The new green/black oscar theme is very good, but even that theme should be bootstrap-free.

**The bandwidth it requires, and the consequent load it consumes, do not worth it, by my opinion.**

dalf:
The simple theme is far lighter.
Reopen the issue, if you want to replace the default theme.

GreenLunar:
The default theme is fine.
The bootstrap CSS files of the default theme are the files that I want to replace.

PR #858, opened 2017-02-12
Simple theme

dalf:
Based on legacy theme, this theme is a copy cat of some existing layout.

Only the results page is modified for now, and still unfinished.

A not exhaustive TODO list :
- [x] mootools can be replaced by https://github.com/autocompletejs/autocomplete.js to create less heavy pages (minus 90k).
- [x] be able to include javascript only if needed (no need to include the image layout javascript if there is no image result)
- [x] use grunt to concat, minimize the javascript files.
- [ ] results page : consistent layout, tweak some details 
- [ ] results page / index page : double check all form parameters (pass all parameters : safesearch and all) 
- [ ] image template : improve display ( https://www.sitepoint.com/recreating-google-images-search-layout-css/ ?)
- [ ] default template : fix height of small image (object-fit and object-position display the real size image, but the overall size remains big). Example : search for "osmand" with the file category.  
- [x] index page layout
- [x] preferences page layout
- [x] map template doesn't work
- [x] torrent template is broken
- [x] video player
- [x] music player
- [x] vim_hotkeys plugin
- [x] search_on_category_select plugin
- [x] open_results_on_new_tab plugin
- [ ] infinite_scroll plugin
- [x] mobile layout, test responsive design.
- [ ] update the RTL style

Related to #828 and #687

capture d ecran de 2017-02-12 15-25-34

capture d ecran de 2017-02-12 15-27-25

capture d ecran de 2017-02-12 15-29-08

capture d ecran de 2017-05-13 13-36-08


dalf:
I've updated my instance with the simple branch : https://al-f.net/searx/ so everyone can give a try and add bugs reports, reviews, comments. 

### About network usage
A result page without images and without autocompleter downloads 37,6Kb. With autocompleted activated, 67.9kb are downloaded (using mootools, http://autocomplete-js.com/ can be used but seems buggy).

More over, the full glyphicons-halflings font is downloaded (17.7kb): a subset can be created to reduce the network usage (may be 10kb less ?).

The magnify can be replace by an icons ( glyphicon-search ) : 2.3kb.

### About plugins
They are heavily tied to jquery (and the oscar templates). In simple theme, a quick implementation is to load jquery when a plugin add some js files. A more long term is to implement plugin with native javascript (or to transpile from ES6 / ES7). Another third way is to [load minimal jquery](http://projects.jga.me/jquery-builder/) version, or use a compatible version like [zepto](http://zeptojs.com/).

dalf:
Some updates : 
* without image, the network usage is 22.1kb for results of the general category. 
* the vim keyboard has been included and modify to use vanilla JS (the help doesn't work for now)
* the browser focus follows the vim selection
* images size are set in a way that each row is always full (the last commit fixes a bug when there are different block of images)   
* the autocompleter uses POST when the main search uses POST too. 

What can improved : 
* search, the input has the focus, go to another page, go back to the searx page : the autocomplete choices are shown. I don't think the user expects that behaviour.
* the "send search on category click" is missing.
* still the preferences page.
* use the full page width when the user search for images only.
* which logo to used ?
* the customize ion font could be included in the css which should reduce the response time.
* refactor the less sources : for now, I think the RTL modification won't be easy.

dalf:
There are still plenty of different things to do, but I think it could be the time to merge, so more people could test it, or even contribute, @asciimoo what do you think ? 

It is more than usable except with RTL languages.

asciimoo (approved):
Thank you.

dalf:
see for remaining work on the simple theme : https://github.com/asciimoo/searx/issues/1025

PR #865, opened 2017-02-22
Dockerfile nitpicks

matejc:
removing removal of py-pip fixes this terrifying message at start of searx in docker environment:

```
CRITICAL:searx.webapp:The pyopenssl, ndg-httpsclient, pyasn1 packages have to be installed.
Some HTTPS connections will fail
```

(searx works with or without this fix)


Wonderfall:
The fix should be merged, this message appeared a while ago. I still can't see why the removal is required but this is it... Searx still works but some search engines such as Google are unavailable due to HTTPS errors. Now it's resolved with this fix.

kvch:
Closing, as #873 was merged.

PR #867, opened 2017-02-25
A couple of fixes with search languages

### LINKED ISSUE #863: qwant  bug with specific languages
Hello ,
From my tests I can say that qwant images works but only with "default language", if I set a language like french or english it give nothing

MarcAbonce:
The first commit simply fixes #863.

The second commit allows users to query in any correctly formatted language code even if it's not on languages.py.
languages.py can change (as the above commit causes), so users may query on a language that is not on the list anymore, even if it is still recognized by a few engines.

Also made no and nb the same because they seem to return the same, though most engines will only support one or the other.


kvch (approved):

asciimoo (approved):
great, thanks

PR #881, opened 2017-03-21
Add framalibre engine

mmuman:
framalibre.org is a catalogue of Free Software, edited by Framasoft.

For now we pass the thumbnail as img_src as it doesn't seem to be used
for IT...

Looks like: https://twitter.com/mmu_man/status/844309927535480832


asciimoo (approved):
Works well, thank you.

Issue #884, opened 2017-03-28
ERROR:searx.engines:Cannot load engine "soundcloud"


hylinux1024:
i got this problem when i run searx

ERROR:searx.engines:Cannot load engine "soundcloud"
Traceback (most recent call last):
  File "/Users/lancelot/PycharmProjects/searx/searx/engines/__init__.py", line 63, in load_engine
    engine = load_module(engine_module + '.py', engine_dir)
  File "/Users/lancelot/PycharmProjects/searx/searx/utils.py", line 300, in load_module
    module = load_source(modname, filepath)
  File "/Users/lancelot/PycharmProjects/searx/searx/engines/soundcloud.py", line 62, in <module>
    guest_client_id = get_client_id()
  File "/Users/lancelot/PycharmProjects/searx/searx/engines/soundcloud.py", line 41, in get_client_id
    response = http_get("https://soundcloud.com")
  File "/Users/lancelot/PycharmProjects/searx/searx/poolrequests.py", line 86, in get
    return request('get', url, **kwargs)
  File "/Users/lancelot/PycharmProjects/searx/searx/poolrequests.py", line 79, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/Users/lancelot/PycharmProjects/searx/searx-ve/lib/python2.7/site-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/lancelot/PycharmProjects/searx/searx-ve/lib/python2.7/site-packages/requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "/Users/lancelot/PycharmProjects/searx/searx-ve/lib/python2.7/site-packages/requests/adapters.py", line 487, in send
    raise ConnectionError(e, request=request)
ConnectionError: HTTPSConnectionPool(host='soundcloud.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x107dea090>: Failed to establish a new connection: [Errno 65] No route to host',))
Traceback (most recent call last):
  File "searx/webapp.py", line 107, in <module>
    initialize_engines(settings['engines'])
  File "/Users/lancelot/PycharmProjects/searx/searx/engines/__init__.py", line 221, in initialize_engines
    engines[engine.name] = engine
AttributeError: 'NoneType' object has no attribute 'name'

dalf:
I see
```
No route to host
```
[...]

PR #888, opened 2017-04-01
[mod] searx doesn't crash at startup when an engine can't be loaded

dalf:
see #884

With this PR, searx ignores engine that can't be loaded (the module), instead of crashing.

A reason to keep the actual behavior : 
* it's up to the administrator to adjust the settings.yml and carefully disable the engines which doesn't work.

A reason to accept this PR : 
* the setup is more easy, there are logs. Administrator can start to use searx and then adjust seeing the logs.


asciimoo (approved):

asciimoo:
thank you

PR #897, opened 2017-04-08
Img src

dalf:
Various fix about images (related to the simple theme, but should be ok whatever the theme) :
* bing_news : the thumbnail are too big compare to the other
* framalibre : remove the 'thumbnail', 'img_src' is enough.
* qwant : add images (news, social)
* twitter : smaller thumbnail (similar to qwant)


asciimoo (approved):

asciimoo:
Great, thanks.

PR #913, opened 2017-05-09
Add Python3 compatibility

asciimoo:
This is a complete rewrite of the existing codebase to be compatible with both python2 and python3.
I also added a compatibility layer for the url manipulation functions to hide differences of naming in different python versions.

Please review the code and test it both with py2 and py3, if possible. The setup should be the same for both python versions.


searx/engines/bing_news.py
```diff
@@ -11,14 +11,18 @@
...
-from urlparse import urlparse, parse_qsl
 from datetime import datetime
 from dateutil import parser
 from lxml import etree
 from searx.utils import list_get
 from searx.engines.bing import _fetch_supported_languages, supported_languages_url
 
+try:
+    from urlparse import urlparse, parse_qsl
+    from urllib import urlencode
```
kvch: Why don't you import `urlencode` from `searx.url_utils` just like in other engines?

kvch (approved):

asciimoo:
@kvch thank you the review. Seems, there isn't any other note on the code, so i'm merging it.

PR #919, opened 2017-05-13
[fix] bing images

MarcAbonce:
Their html changed a little bit, so it wasn't working anymore.

The only issue now is that apparently they don't display the image's titles anymore in their own results.


asciimoo (approved):
Great, thank you

PR #921, opened 2017-05-15
Load engines paralelly

kvch:
Unfortunately, `load_module` must be locked, otherwise engines like `xpath` can't be loaded.


kvch:
Closing, because the locking takes away the point of loading engines paralelly.

PR #931, opened 2017-05-27
https more secure ?

dalf:
when searx connects to a search engine :
* using HTTPS, TLS1.2 and perfect forward secrey ciphers will be used (and AES256-GCM-SHA384 see the code)
* using HTTP, a warning is logged.

Moreover :
* fix gigablast engine
* engines don't use ```requests``` but only ```searx.https``` (formely ```poolrequests```)

I'm not a security expert, so may be the cipher list is wrong in a way or another. 

Important note : when I activate all engines in the general category and make a search, the server quits without an error (not always reproducible).

[EDIT] see https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility


dalf:
More or less related to this PR and / or outgoing http(s) connections.

Create a per engine HTTP configuration :
- forbidden HTTP or not (could be set to true by default)
- the cipher suite to use.
- the proxy to use. The global proxy configuration remains : it would easy to send some requests through TOR for only some engines.

In addition, the set of IP address could be per engine instead of per process : for the engine, searx would never use the IP twice in row (now it is not guarantee).

dalf:
This branch is outdated: some ciphers must be removed.

Different approach in this branch : https://github.com/dalf/searx/blob/httpx_firefox/searx/poolrequests.py#L19

* same cipher list than Firefox 73, to look like Firefox.
* use [httpx](https://github.com/encode/httpx) so use HTTP/2.

If we want to combine the two:
* Firefox cipher list because it helps for some engines (need to check).
* only use secure ciphers

It is possible to stop the connection if an unsecured cipher is choosen by the server (it requires a dedicated [httpx backend](https://github.com/encode/httpx/blob/master/httpx/_backends/asyncio.py#L84) that check the ssl connection).

PR #932, opened 2017-05-28
[fix] fix duckduckgo's offset

MarcAbonce:
First page now starts with 0 offset, rather than starting on the 30th result.

DuckDuckGo returns 30 results on each page.


asciimoo (approved):
Thanks the fix.

PR #933, opened 2017-05-28
searx fetchs automaticaly firefox last versions

dalf:
At launch and every week, searx fetches versions from https://ftp.mozilla.org/pub/firefox/releases/

The user agent sent to search engines is based on this list, but only the last two major versions are kept (including all  minor versions).

The purpose is to match the most widespread versions at any time.

The linux 32bits architecture is removed in the same purpose.

Currently, the list is : 
* 53.0.3
* 53.0.2
* 53.0
* 52.0.2
* 52.0.1
* 52.0

### DIFF
--- a/searx/data/useragents.json
+++ b/searx/data/useragents.json
@@ -1,0 +1,15 @@
+{
+    "os": [
+        "Windows NT 10; WOW64", 
+        "X11; Linux x86_64"
+    ], 
+    "ua": "Mozilla/5.0 ({os}; rv:{version}) Gecko/20100101 Firefox/{version}", 
+    "versions": [
+        "59.0.2", 
+        "59.0.1", 
+        "59.0", 
+        "58.0.2", 
+        "58.0.1", 
+        "58.0"
+    ]
+}
--- a/searx/data/useragents.json
+++ b/searx/data/useragents.json
@@ -1,15 +1,14 @@
 {
-    "os": [
-        "Windows NT 10; WOW64", 
-        "X11; Linux x86_64"
-    ], 
-    "ua": "Mozilla/5.0 ({os}; rv:{version}) Gecko/20100101 Firefox/{version}", 
+    "ua": "Mozilla/5.0 ({os}; rv:{version}) Gecko/20100101 Firefox/{version}",
     "versions": [
-        "59.0.2", 
-        "59.0.1", 
-        "59.0", 
-        "58.0.2", 
-        "58.0.1", 
-        "58.0"
+        "61.0.1",
+        "61.0",
+        "60.0.2",
+        "60.0.1",
+        "60.0"
+    ],
+    "os": [
+        "Windows NT 10; WOW64",
+        "X11; Linux x86_64"
     ]
 }
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -10,8 +10,10 @@ from codecs import getincrementalencoder
 from imp import load_source
 from numbers import Number
 from os.path import splitext, join
+from io import open
 from random import choice
 import sys
+import json
 
 from searx import settings
 from searx.version import VERSION_STRING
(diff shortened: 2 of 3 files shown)


asciimoo:
As I see, it tries to fetch the latest FF versions on every startup (why only if debug mode is turned off?), which makes the startup time longer. It would be good to cache the versions or fetch it parallel. What do you think?

dalf:
> why only if debug mode is turned off?

because it makes the startup time longer :-) 
when in production, I don't think it's an issue if startup time is a little bit longer.

At first I was thinking to store the versions in a file, but I don't know where ?
In the ```data``` directory, I think it will make trouble with the deb package.
In the ```/tmp``` may be ?

The other solution about parallel version could be simpler even it's more hackish. ```update_firefox_version()``` can create a thread to do the real work.

asciimoo:
>  when in production, I don't think it's an issue if startup time is a little bit longer.

Well, it matters for me e.g. on searx.me if a webapp restart takes more time after every update. And, the current init time is far from the optimal, so I wouldn't increase it.
Also, I'm thinking on adding an optional `init()` function to the engines, which could be parallelized, so engines with heavy initialization (like wolframalpha or soundcloud) wouldn't affect the startup time of searx.

> The other solution about parallel version could be simpler even it's more hackish. update_firefox_version() can create a thread to do the real work.

I can't see why would be it hackish, this is the best solution for me =)

dalf:
> I can't see why would be it hackish, this is the best solution for me =)

So thread it will be. About the hackish : it's my wish to remove the one HTTP request = one thread.
Moreover if there are 4 process (with uwsgi), the firefox versions will be fetched 4 times.

+1 for the init() method.

dalf:
Done.

I've added a 2 seconds timeout on the URL fetch and updated standalone_searx.py.

```time_to_update check``` is not thread safe so once a week, ```schedule_update_firefox_version()``` can be called multiple time in a row. The ```update_lock``` prevents multiple URL fetches at this moment.

dalf:
Just to be sure, Is it the fetch at the start of the searx that block this PR ?

A static tmp file could solve the issue : 
- a file is create in a defined location
- the file is updated once a week
- when an instance is (re)loaded, the file is only loaded

Only issue :  define this static file (or database).

asciimoo:
>Just to be sure, Is it the fetch at the start of the searx that block this PR ?
A static tmp file could solve the issue :
    a file is create in a defined location
    the file is updated once a week
    when an instance is (re)loaded, the file is only loaded
Only issue : define this static file (or database).

Yes, it would be good to avoid hitting mozilla on every app reload.
The main reason why this PR is still open that I have doubts if the complexity of this feature isn't bigger than the original problem itself. It's not that hard to update the user-agent version strings occasionally (before releases) by hand and I can't see why would be beneficial to keep it more up-to-date.

dalf:
I agree it may be too much code.

According to http://caniuse.com/usage-table  version 40 to 47 represents 0.37% of user requests.

The purpose of this PR is to merge with the 4.08% which is a little better (last two versions of Firefox).
Using the last versions of Chrome would be even better.

But anyway, I guess search engines can easily find out that searx is not a real browser.

dalf:
Another implementation : 
* create a utils/fetch_firefox_version.py which create searx/data/firefox_versions.json
* searx loads that file

So : 
* The start time of searx is not impacted
* The versions can be updated whenever the administrator decided (requires to reload searx)

MarcAbonce:
I ran it locally and I just have two minor points to nitpick:

The script fails with python 3. First, the versions list can't be sorted because some version numbers mix numbers and strings. Since you are already filtering out the ESR versions later on in the code, you could just exclude them from NORMAL_REGEX and the result should look the same. There's also an encoding error when writing the json file with python 3.

Also, when opening useragents.json, the relative path assumes the script is being run from searx/utils, so it fails if you run the script from any other directory. Importing searx_dir from searx and using that instead of the relative path solves this.

dalf:
I have updated the branch.

The useragents.json file is located in ```searx/data/useragents.json```

Because this file can be updated, it would be interresting to have this file in another location when searx is packaged.

Some possible locations : 
* the same directory than settings.yml
* a directory referenced by settings.yml ( relative path "searx/data/useragents.json" by default )

dalf:
The design has changed a lot since this pull request has been opened.

In the current implementation : 
* the ```searx/data/useragents.json``` contains the user agents to be used when requests are sent to the different search engines.
* this file can be updated by the searx administrator either manually, either automatically using the script ```utils/fetch_firefox_version.py```. This script is never called automatically : someone has to call it.

So, in the current implementation, I don't think there is privacy breach ?

josch:
@dalf cool! Thanks for the update! :smile:

dalf:
Moreover, it is possible to get the last Firefox versions using [wikidata]( https://www.wikidata.org/wiki/Q698 ) the same way searx answers user requests.

It is less realiable : anyone can modify the wikidata page to highlight searx searches (if searx administrator don't check the results from the script).

asciimoo (approved):
Great, now this is a useful automation without exposing any additional information about the service, thanks!

PR #941, opened 2017-06-11
[fix] wrap suggestions inside suggestion box

MarcAbonce:
Long suggestions now wrap around the box instead of going overboard.

The buttons still adapt to the word's length, but only within the limits of the parent box.


kvch (approved):

asciimoo (approved):
Thank you.

Issue #943, opened 2017-06-13
Largest images engines, Bing and Google, no results


jodo13:
If we can trust http://stats.searx.oe5tpo.com/ and by the results we can, Searx instances can't retrieve results for the largest engines.

Searx becomes almost useless for image searching then with the rest of available engines.

pointhi:
please note this site is testing the connectivit on my instance, which wasn't update a long time now. Many engines likely have fixes, as well as some engines are probably blocked. Dont trust the results to much :-)

jodo13:
I see, that is why I said too that by the results (at least I have).

Anyway, I use laquadrature.net instance (0.11) and searx.me (0.12) and along the last week and these last couple of days, both couldn't retrieve from Google and Bing. Now Searx.me can't still retrieve from Google and intermittent with Bing.

So, let me tell it here, just in case something is broken too ;-)

Thanks anyway for all your work :)

kvch:
@jodo13 Do you mean Bing or Bing Images? Both of them are seem to work on searx.me.

jodo13:
@kvch as the subject of the issue says, I'm talking about images engines, or images results, so, Bing Images :)

As told in the previous message to yours, it is intermittent with Searx.me instance. Sometimes display Bing images results, and some doesn't. Might be high load on Bing Images or whatever.

Thanks for the google fix. I hope to see it soon on instances :)

kvch:
@jodo13 Thank. :) So far I haven't been able to reproduce it so far nowhere. So yes, it is possible that it is a Bing problem.

PR #944, opened 2017-06-13
[fix] fix xpath of google images

kvch:
A new class was added to the div we looked for in the xpath. I followed up the changes in the code, so now it works.

This partly solves the problem mentioned in #943.


asciimoo (approved):
Thank you

PR #945, opened 2017-06-15
[fix] dictzone and translated engines

potato:
changes in languages.py broke the dictzone and translated engines caused by the new elements in the languages dict. returning the correct values from utils/is_valid_lang now.


asciimoo (approved):

asciimoo:
Thanks.

PR #951, opened 2017-07-04
Italian localisation improvement

lcpz:
I corrected typos and enhanced the translation.


kvch:
Thanks for the improvements. However, translations are maintained on https://www.transifex.com/asciimoo/searx/. Please, correct mistakes there. :)

lcpz:
I did it. I don't understand how it works, though. Does it get a review there?

kvch:
No, review of translation happens on transifex. But translation files are not updated immediately in this repo, after changes are made on transifex. It usually happens before a release or when lots of things are changed in the translations.

But if you want to get it merged now, please add the corresponding generated `.mo` to your PR. If you need help with that, there is a tutorial on pulling and generating bins from translation files: https://asciimoo.github.io/searx/dev/translation.html.

lcpz:
Requested `.mo` file added.

---

Your [translation](https://asciimoo.github.io/searx/dev/translation.html) page is wrong:

1. the Transifex instance to specify in `tx init` is just the default one (otherwise client won't login)
2. the `tx set` command currently does not work as is

For the point 2, I used `-l it` instead of `--auto-local` but it gave me chinese; I tried again, and it gave me hebrew. I ended up downloading the `.po` file manually.

---

The Transifex client stores password and token in plain (but the site keeps it *secure* by showing it to you only once!). Plus, it's really frustrating to use (see above).

One could just:

1. `git clone` searx
2. modify/create `searx/translations/<lang>/LC_MESSAGES/messages.po`
3. `pybabel compile -d searx/searx/translations`

kvch (approved):

kvch:
Thank you for updating your PR.

I will update the documentation based on your feedback.

I agree with you, transifex can be frustrating to use. However, we are not planing on leaving it, because it makes localization of searx accessible to people who don't know git or pybabel. We don't want to lose those translators. :)

lcpz:
@kvch Since it's intended for devs, I suggest to remove the use of Transifex client from your translation page and just add my previous commands, so that you won't loose other people like me either. 😃

asciimoo:
First of all, thank you for updating the translations.

> Since it's intended for devs, I suggest to remove the use of Transifex client from your translation page and just add my previous commands, so that you won't loose other people like me either. 😃

In long term the currently documented method is the most comfortable way to sync translations with transifex imo. But, we can document that the use of the transifex client is fully optional, only synchronization of the repo and transifex is important, not the way you do it.

asciimoo (approved):

PR #952, opened 2017-07-05
[fix] duckduckgo images doesn't fail with countryless language

MarcAbonce:
Right now duckduckgo images fails when the language has no country, e.g. `:de !duckduckgo_images nein` so this fixes that bug.


asciimoo (approved):
Great, thank you.

Issue #948, opened 2017-06-19
Extend global timeout when there is no result


dalf:
Currently, if all search engines timeout, the user doesn't have any result to read.
In this scenario, the user usually refresh the page to make sure there is really no result : and sometimes it displays some results.

Technically, this can happen when the HTTPS connections have been closed. On the second try, the connections are already opened, and the results arrived before the global timeout. 

See : https://news.ycombinator.com/item?id=14579491#14581065

I see two solutions (most probably there are a lot of more) : 
* advise the administrator to increase the global timeout in this case. How ?
* change the searx behavior : if there is no results after the timeout, double the timeout and wait more.

PR #961, opened 2017-07-12
show engine errors in infobox && add new error alert

kvch:
From now on engine errors with type name are visible to users. I also removed the unnecessary close button from the "no result" alert.

error_wo_results

error_w_result

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -135,6 +135,7 @@ class ResultContainer(object):
         self._number_of_results = []
         self._ordered = False
         self.paging = False
+        self.unresponsive_engines = []
 
     def extend(self, engine_name, results):
         for result in list(results):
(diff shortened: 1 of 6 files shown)


kvch:
I added the error list to the alert box shown when there is no results.

error_wo_results_more

asciimoo (approved):
Nice PR and feature, thanks =)
#948 is related to this and it would be also a useful addition.

dalf:
https://github.com/asciimoo/searx/blob/243d3e4298c93014f37ca6f1f957a60cb09f4ae1/searx/search.py#L182

Timeout in the ```search_multiple_requests``` function are not displayed.

implementation idea:
- call ```add_unresponsive_engine``` method.
- the ```add_unresponsive_engine``` checks if the engine was already added.

kvch:
@dalf You are right. I will fix it. Thanks for the review. :)

PR #969, opened 2017-07-19
Add missing timeouts

kvch:
As pointed out by @dalf in https://github.com/asciimoo/searx/pull/961#issuecomment-315597756, a timeout error was missing. I added this to the existing errors.

Furthermore, to eliminate duplication of errors `unresponsive_engines` is changed to `set` from `list`.


asciimoo (approved):
Thanks

Issue #887, opened 2017-03-31
pyparsing missing from install


micah:
If I follow the installation instructions (on Debian Jessie), on the ```./manage.sh update_packages``` step, the following happens:

```
...
  Found existing installation: setuptools 5.5.1
    Uninstalling setuptools:
      Successfully uninstalled setuptools
  Running setup.py install for cffi
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/setuptools/__init__.py", line 12, in <module>
        import setuptools.version
      File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/setuptools/version.py", line 1, in <module>
        import pkg_resources
      File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 72, in <module>
        import packaging.requirements
      File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/packaging/requirements.py", line 9, in <module>
        from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
    ImportError: No module named pyparsing
    Complete output from command /srv/searx/searx-ve/bin/python2 -c "import setuptools, tokenize;__file__='/tmp/pip-build-DzEoyT/cffi/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-HZc3xx-record/install-record.txt --single-version-externally-managed --compile --install-headers /srv/searx/searx-ve/include/site/python2.7:
    Traceback (most recent call last):

  File "<string>", line 1, in <module>

  File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/setuptools/__init__.py", line 12, in <module>

    import setuptools.version

  File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/setuptools/version.py", line 1, in <module>

    import pkg_resources

  File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 72, in <module>

    import packaging.requirements

  File "/srv/searx/searx-ve/local/lib/python2.7/site-packages/packaging/requirements.py", line 9, in <module>

    from pyparsing import stringStart, stringEnd, originalTextFor, ParseException

ImportError: No module named pyparsing

----------------------------------------
Cleaning up...
[...]

Issue #962, opened 2017-07-14
Import error searx 0.12 at launch


Eliesemoule:
Hello,

When I launch searx with this command : `python searx/webapp.py`, I got this error :

`Traceback (most recent call last):
  File "searx/searx/webapp.py", line 33, in <module>
    from searx import logger
  File "/usr/local/searx/searx/__init__.py", line 18, in <module>
    import certifi
ImportError: No module named certifi`

Searx was installed on CentOS 7 with the procedure in the wiki. The 0.11 of searx worked fine on my server

Eliesemoule:
Issue solved and install procedure updated to work on CentOS 7

PR #972, opened 2017-07-20
"./manage.sh update_packages" updates pip and setuptools

dalf:
Fix issue #887, #962


dalf:
The previous commit c0993ece546716e485478a5b86e30c41e81f5154 avoids ```--upgrade```, so I don't know if this is the right place to do an upgrade of pip and setuptools

dalf:
What about this one?

asciimoo (approved):
Thank you and sorry for the delay

PR #973, opened 2017-07-20
Remove 'all' option from search languages

MarcAbonce:
The meaning of the all/Automatic/Default language setting has always been inconsistent. For about half of the engines the setting is hardcoded back to English and only a few engines actually parse the setting, returning either multiple languages or the engine's default language.

So I think it would be better to simply drop it and use an actual language as the default. I put `en-US` as the new default instead of just `en` because some engines require a country code.

### DIFF
--- a/searx/engines/archlinux.py
+++ b/searx/engines/archlinux.py
@@ -26,7 +26,7 @@ xpath_results = '//ul[@class="mw-search-results"]/li'
 xpath_link = './/div[@class="mw-search-result-heading"]/a'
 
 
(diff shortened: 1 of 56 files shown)


MarcAbonce:
I just updated this and added a couple more unit tests.  

At this point, after #1085, none of the default engines in the general category even support a separate behaviour for the 'all' option.

asciimoo (approved):
@MarcAbonce thanks for fixing this in all engines.
I agree, we can discard `all` from the languages

PR #974, opened 2017-07-22
New engine: Google videos

MarcAbonce:
Thumbnail and embedded video are not supported, though.

I also made a small fix to include all result parameters in the merged result. For example, in this case a merged result will always include the thumbnail as long as one of the source engines has it, regardless of the merging order.

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -218,6 +218,11 @@ class ResultContainer(object):
                     result_content_len(duplicated.get('content', '')):
                 duplicated['content'] = result['content']
 
+            # merge all result's parameters not found in duplicate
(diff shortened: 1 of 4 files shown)


searx/results.py
```diff
@@ -218,6 +218,11 @@ def _merge_result(self, result, position):
                     result_content_len(duplicated.get('content', '')):
                 duplicated['content'] = result['content']
 
+            # merge all result's parameters not found in duplicate
```
asciimoo, line 221: 👍

asciimoo (approved):
Great! Thank you.

Issue #975, opened 2017-07-22
Core: HTTP timeout for all requests.


dalf:
Right now, search.py set a timeout on the main HTTP request, but there is no timeout for additional requests (see [duckduckgo_images](https://github.com/asciimoo/searx/blob/077d8efed85e468d863a52e8a32feb88ba158c11/searx/engines/duckduckgo_images.py#L36), [wikidata ](https://github.com/asciimoo/searx/blob/52e615dede8538c36f569d2cf07835427a9a0db6/searx/engines/wikidata.py#L80)for example).

One idea : 
* remove the   ```timeout=timeout_limit,``` in the ```send_http_request``` (search.py), and the ```if search_duration > timeout_limit + timeout_overhead:```
* at the begining of ```search_one_request_safe``` (or ```search_one_request```), call a function of ```searx.poolrequests``` to set the timeout of all HTTP request for this thread. 
* This function technically set a threadlocal value (store the timestamp when the engine must timeout).
* inside ```searx.poolrequests``` : automatically set the timeout (if the threadocal is set) for HTTP request.
* the thread is killed after each request, so there is no need to reset the threadlocal value.

So any additional HTTP requests would have a timeout without additional code in the engine.

Side effect : the ```send_http_request``` could be more simple (no start_time, timeout_limit parameters).

PR #977, opened 2017-07-23
[enh] timeout and total HTTP time are managed by searx.poolrequests

dalf:
Implement #975

### DIFF
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -1,8 +1,9 @@
 import requests
 
 from itertools import cycle
-from threading import RLock
+from threading import RLock, local
 from searx import settings
+from time import time
 
 
 class HTTPAdapterWithConnParams(requests.adapters.HTTPAdapter):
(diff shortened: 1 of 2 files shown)


dalf:
About [```if search_duration > timeout + timeout_overhead```](https://github.com/asciimoo/searx/pull/977/files#diff-678059f39d0697a791718f3a0ddb2f8fR114)
I don't know if it is relevant to keep it ?

Without it, it means the parsed results will be ignored later, inside ```search_multiple_requests```

In addition, this code in search.py
```python
user_agent = gen_useragent()
...
request_params['headers']['User-Agent'] = user_agent
```
could be changed to have a default user agent on all HTTP requests. The default headers could be set at the beginning of ```search_one_request_safe``` (or perhaps ```search_one_request```) using ```poolrequests.set_headers_for_thread```

dalf:
Since a new version has been released, what about merging this one, and take time to test it in master branch ?

asciimoo:
I agree, this is the good time to experiment. We could do the same with #1034.

asciimoo (approved):
👍

PR #992, opened 2017-08-12
[mod] fix #903 : add a function is_accepted to engine.

### LINKED ISSUE #903: An useless thread is created on each request in the general category
The commit 01e2648e93e718f84c77aed16326a14f2b498a2e introduces a regression : 
* before this commit, the request function was calls then if an url was requested a thread was created
* with this commit a thread is created for each engine 

It's not an issue except for one engine in the general category: currency.

Most of the time the currency engine doesn't send a request, but a thread is still created (which take time)

The call to the request function in the main thread is also a problem since some engine sends an HTTP request ([wolframalpha](https://github.com/asciimoo/searx/blob/master/searx/engines/wolframalpha_noapi.py#L66) for example).

A better solution would be a function can_accept defined inside the engine ? (similar to the PR #797) 

Edit : the PR #853 could be implemented as an engine too.

dalf:
fix #903
search.py calls this function before creating a thread for the engine.
if the function doesn't exist, the function is .... not called.

without this requests a useless thread is created for the dictzone and currency engines.

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -399,6 +399,11 @@ class Search(object):
             request_params['safesearch'] = search_query.safesearch
             request_params['time_range'] = search_query.time_range
 
+            # check if the engine accepts the request
+            if hasattr(engine, 'is_accepted'):
+                if not engine.is_accepted(search_query.query, request_params):
+                    continue
+
             # append request to list
             requests.append((selected_engine['name'], search_query.query, request_params))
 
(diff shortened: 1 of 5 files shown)


dalf:
Another more simple implementation : just define a regex for an engine.
If search.py see a defined regex, it checks if the query matches this regex.

dictzone and translated do extended check of the languages in ```is_accepted```, with the regex version it won't be possible.

searx/search.py
```diff
@@ -399,6 +399,11 @@ def search(self):
             request_params['safesearch'] = search_query.safesearch
             request_params['time_range'] = search_query.time_range
 
+            # check if the engine accepts the request
+            if hasattr(engine, 'is_accepted'):
```
dalf, line 403: Move this check in the get_search_query_from_webapp function.

dalf:
Supersed by #2159

PR #994, opened 2017-08-12
[fix] fix no_results.html template of the simple theme

dalf:
I forgot this file when I made the commit  10a24bdc2c3870f07ec62dd710841628d325aaf6


asciimoo (approved):
Thanks

PR #1009, opened 2017-08-24
[fix]torrentz search engine fixed for new version

misnyo:
Torrentz.eu service has been shut down, however, the successor is now working properly in another url(https://torrentz2.eu). Minor changes to the engine are required.


asciimoo (approved):
Great, thanks

PR #1014, opened 2017-08-30
[fix] faroo json api and image layout fixed

misnyo:
Faroo doesn't require an api if referer is set correctly, thus obsolating the api_key.
Also search results having an image now have videos.html template.


asciimoo (approved):
:+1:

Issue #1025, opened 2017-09-03
Simple theme : bug fixes, improvements


dalf:
### General : UI
- [ ] center or move the results to the right.
- [ ] footer at the botton of the page ( https://codepen.io/cbracco/pen/zekgx http://matthewjamestaylor.com/blog/keeping-footers-at-the-bottom-of-the-page )
- [x] use https://gist.github.com/dalf/1ace543608ccc04ac8a5c6068eaa693e

see https://joe-crick.github.io/pure-css-components/

### General : UX
- [x] Display "Search URL of the currently saved preferences" in the preferences.
- [x] Show "Answerers" somewhere.
- [ ] do the RTL style 
- [ ] engine crash warning : add a close button
- [ ] infinite_scroll plugin
- [ ] infoboxes : some properties value should be a link to another search ( ```Programming language :Python```  --> be able to click on ```Python``` to search that on searx)

###  Search
- [x] add a clear text button on the left of the search button
- [ ] add safe search right to Language combobox and Time combobox
- [ ] apply #1226 to Simple theme

### On mobile
- [ ] do not set the focus on the search input text : the keyboard hides the results.
- [ ] make the search input smaller when without focus to give space for the searx logo and the preferences button. With focus, make it full width wide as now.
- [ ] embedded videos on mobile : too large, outside the screen
- [ ] selected category on mobile : scroll to the first selected category.

### Image category
- [ ] image category : improve display ( https://www.sitepoint.com/recreating-google-images-search-layout-css/ ?)
- [x] image category : full screen width when there is only images.

### Reduce the network usage
- [x] remove searx/static/themes/simple/img/loader.gif using https://projects.lukehaas.me/css-loaders/
preferences button.
- [ ] create a small bandwidth version using javascript (use JS to add languages choices, cached links, categories)

kvch:
@dalf I started working on adding a clear button to the search input. I created a css style, but I hasn't been able to compile it, because of dependencies. How did you install `clean-css`? I tried using npm, but still no success.

dalf:
@kvch I have followed the "recipe" inside [.travis.yml](https://github.com/asciimoo/searx/blob/master/.travis.yml) : 
```bash
npm install less@2.7 less-plugin-clean-css grunt-cli
export PATH=`pwd`/node_modules/.bin:$PATH
./manage.sh install_geckodriver
./manage.sh npm_packages
./manage.sh update_dev_packages
```

then
```bash
./manage.sh styles
[...]

PR #1029, opened 2017-09-07
Chiefr integration - distributed contribution model for searx

asciimoo:
Hello,

I'd like to try to decentralize the contribution model of searx because of the following reasons:
 - Reduces the load of maintainers
 - Increases the responsiveness if a contribution arrives
 - Explicitly defines ownership of project parts (segments)
 - Experimenting is always good =)

I've created a tool (https://github.com/asciimoo/chiefr) to support the above idea.
Chiefr requires to define a maintainers file which contains all information about the maintainers and resources of project segments. I already created a minimal `.maintainers.ini` file and added some segments, but it is far from complete.

Please comment, if you have any thoughts about the idea or you want to be a maintainer of a segment of searx, and I'll update the maintainers file. Segment can be any set of files (except empty set) in the repo, even a single engine or just a function.

### DIFF
--- a/.maintainers.ini
+++ b/.maintainers.ini
@@ -1,0 +1,56 @@
+[engines]
+Repository = https://github.com/asciimoo/searx
+IssueTracker = https://github.com/asciimoo/searx/issues
+Chiefs = asciimoo, kvch
+FilePatterns = searx/engines/.*
+Topics = engine
+Priority = 1
+
+[http backend]
+Repository = https://github.com/dalf/searx
+IssueTracker = https://github.com/dalf/searx/issues
+Chiefs = dalf
+FilePatterns = searx/poolrequests\.py
+Topics = http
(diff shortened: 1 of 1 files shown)


dalf:
What is the process to merge in asciimoo/searx ?
Make a huge PR from time to time ? as they arise ?

So I should move #1025 to my fork ?

asciimoo:
> What is the process to merge in asciimoo/searx ?
> Make a huge PR from time to time ? as they arise ?

It's totally up to us. If a maintainer receives or writes some code which adds value to searx, then she/he can submit it to the main repo.

> So I should move #1025 to my fork ?

Yes, that would be the proper way in the future, but I think, it's not necessary to put any extra effort to change previous contributions to fit to this model, so it's ok here.

asciimoo:
closing due to lack of interest

Issue #1028, opened 2017-09-05
limit virtual interfaces to specific engines


micah:
Because so many people have problems with google blocking results due to a single IP making many queries (see #729), it seems the only way to get around this is to have a pool of IPs that can be used to make queries, spreading the load around. 

This can be done by setting the outgoing interfaces in virtual interfaces in https://github.com/asciimoo/searx/blob/master/searx/settings.yml#L43, it works fine if you have an excess of IPv4 addresses to spare for this purpose. However, most people do not have spare IPv4. Many people do have large numbers of ipv6 addresses, but if you configure several virtual interfaces to use ipv6 addresses, then the engines that do not support ipv6 will fail.

It would be good if there were a possibility to configure which engines are used in virtual interfaces, so you can configure a bunch of ipv6 virtual interfaces, but only limit them to those engines that support ipv6.

dalf:
Technical informations :
* https://github.com/requests/requests/issues/1691
* https://stackoverflow.com/questions/33046733/force-requests-to-use-ipv4-ipv6

misnyo:
I'll take a look on this soon.

xinomilo:
startpage, bing, reddit and others, aren't accessible to ipv6. no AAAA records.  in ipv6 vm, most engines produce same error.
wikidata is working here in a test ipv6 instance, so don't know why the exception there. what does uwsgi say?

micah:
xinomilo <notifications@github.com> writes:

> startpage, bing, reddit and others, aren't accessible to ipv6. no AAAA
> records.  in ipv6 vm, most engines produce same error.

Same with duckduckgo, no AAAA records.

> wikidata is working here in a test ipv6 instance, so don't know why the exception there. what does uwsgi say? 

```
ERROR:searx.search:engine wikidata : requests exception(search duration : 0.77158498764 s, timeout: 3.0 s) : HTTPSConnectionPool(host='www.wikidata.org', port=443): Max retries exceeded with url: /w/api.php?action=parse&format=json&page=Q10443&uselang=en&redirects=1&prop=text%7Cdisplaytitle%7Clanglinks%7Crevid&disableeditsection=1&disabletidy=1&preview=1&sectionpreview=1&disabletoc=1&utf8=1&formatversion=2 (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7f0b9dfe6ad0>: Failed to establish a new connection: [Errno -9] Address family for hostname not supported',))
[...]

PR #1034, opened 2017-09-12
Source ip select

misnyo:
Basically how it works is that in poolrequests.py, the source_ips from the config are extracted to `http(s)_adapters['source_ip']` by ip versions respectively.
Then, when it is needed, in `SessionSinglePool` the adapters are assigned by whether the engine supports ipv6 and there is at least one ipv6 source ip is present, or if not, if an ipv4 source ip is present. In any other case, the default interface is used.
`searx.utils.dictrandom` is an iterator wrapper for `random.choice`
The decision if an engine supports ipv6 is the following: a new attribute for every engine is introduced, called "meta", and the key ipv6_support of this dict decides it, which defaults to False. Overwriting this key in either the [engine].py or in settings.yml will cause the engine to "support ipv6".
The configuration of source ips are fault tolerant in a way that if there is an entry which is not a valid ip address, it ignores it, and writes a debug message telling that entry is invalid.
This PR should solve #1028
Since this PR rewrites some parts of how the poolrequests works, **I recommend extensive testing** especially with the combination of different source ips.

### DIFF
--- a/searx/engines/1337x.py
+++ b/searx/engines/1337x.py
@@ -7,6 +7,7 @@ url = 'https://1337x.to/'
 search_url = url + 'search/{search_term}/{pageno}/'
 categories = ['videos']
 paging = True
+meta = {'ipv6_support': True}
 
 
 def request(query, params):
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -49,7 +49,9 @@ engine_default_args = {'paging': False,
                        'disabled': False,
                        'suspend_end_time': 0,
                        'continuous_errors': 0,
-                       'time_range_support': False}
+                       'time_range_support': False,
+                       'meta': {
+                           'ipv6_support': False}}
 
 
 def load_engine(engine_data):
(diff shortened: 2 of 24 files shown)


dalf:
:+1: 
Is ipv6 still used in case of redirect ?

misnyo:
In my tests, it uses the same ip for one request "chain".

unixfox:
@micah the "captcha issue" won't be resolved until we find a good solution to combat against bots because Google will keep blocking the IP if he finds repetitive abusive searches.
This may be a temporary fix but not a permanent one.

ilf:
Unfortunately, I don't have a fix. But I want to stress the importance. Searx is awesome and neccessary. But in the last months, most instances have become useless, because they are blocked by Google. This hurts the very goal of searx. :(

unixfox:
@ilf Yeah that's an issue. Like I said [here](https://github.com/asciimoo/searx/issues/1584#issuecomment-493323951) I found a way to filter most of the bots but I won't be able to release that experimental project right now because it's not stable and I don't have the time to work on it anymore (I'm currently in my exams sessions) but I promise after my exams I'll publicly release the same good protection against bots that I currently use on my instance.

Meanwhile, you can use my Searx instance: https://searx.be. It works by default with Google without any issue.

ilf:
> Meanwhile, you can use my Searx instance: https://searx.be. It works by default with Google without any issue.

Now that's blocked, too :(

unixfox:
> Now that's blocked, too :(

Yeah I know... sometimes my instance gets blocked by Google because my antibot program isn't perfect but I'm working on a better solution that will directly block access to every bot instead of blocking them after 2 retries.

asciimoo:
We are just before changing our network stack, so it wouldn't be efficient to revive this PR. Note that currently both ipv4 and ipv6 source addresses are supported as long as they are not mixed together (only ipv4 or only ipv6 source ips). ! keep open #1028 as a reminder to implement mixed source ip support in the future.

Issue #1036, opened 2017-09-14
[engine] add PubMed to Science tab


jibe-b:
PubMed: Index agregating scientific publications in life science.

doc: https://www.ncbi.nlm.nih.gov/home/develop/api/

self-assigned

kvch:
It was added by #1075

PR #1039, opened 2017-09-19
add pubmed engine

jibe-b:
PubMed is an index agregating life science literature.

Specificities:
-  to get results, two requests have to be made: one to get PMIDs (PubMed IDs) and one to get abstracts, hence snippets. The first one is made in `request`, the second one by making a second request. This might break the design principles of Searx, please tell me if there is a convention on how to do it
- from a result, it is possible to link to the PubMed notice, as well as use the DOI when available. How can one add HTML to the "content" section of a result? That would enable to use both

### DIFF
--- a/searx/engines/pubmed.py
+++ b/searx/engines/pubmed.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python
+
+"""
+ PubMed (Scholar publications)
+
+ @website     https://www.ncbi.nlm.nih.gov/pubmed/
+ @provide-api yes (https://www.ncbi.nlm.nih.gov/home/develop/api/)
+
+ @using-api   yes
+ @results     XML
+ @stable      yes
+ @parse       url, title, publishedDate, content
+ More info on api: https://www.ncbi.nlm.nih.gov/books/NBK25501/
+"""
+
+from lxml import etree
+from datetime import datetime
+from searx.url_utils import urlencode
+import urllib2
+
+
+categories = ['science']
+
+base_url = 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi'\
+           + '?db=pubmed&{query}&retstart={offset}&retmax={hits}'
+
+# engine dependent config
+number_of_results = 10
+pubmed_url = 'https://www.ncbi.nlm.nih.gov/pubmed/'
+
+
+def request(query, params):
+    # basic search
+    offset = (params['pageno'] - 1) * number_of_results
+
+    string_args = dict(query=urlencode({'term': query}),
+                       offset=offset,
+                       hits=number_of_results)
+
+    params['url'] = base_url.format(**string_args)
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    # First retrieve notice of each result
+    pubmed_retrieve_api_url = 'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?'\
+                              + 'db=pubmed&retmode=xml&id={pmids_string}'
+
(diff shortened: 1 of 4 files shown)


jibe-b:
achieving #1036

searx/engines/pubmed.py
```diff
@@ -0,0 +1,117 @@
...
+ @results     XML
+ @stable      yes
+ @parse       url, title, publishedDate, content
+ More info on api: https://www.ncbi.nlm.nih.gov/books/NBK25501/
+"""
+
+from lxml import etree
+from datetime import datetime
+from searx.url_utils import urlencode
+try:
```
kvch: Is it required beacuse of py2/py3 compatibility? If yes, please move it to `url_utils`, as it is where all proper URL functions are imported based on Python version.

searx/engines/pubmed.py
```diff
@@ -0,0 +1,117 @@
...
+
+    retrieve_url_encoded = pubmed_retrieve_api_url.format(**retrieve_notice_args)
+
+    search_results_xml = urlopen(retrieve_url_encoded).read()
+    search_results = etree.XML(search_results_xml).xpath('//PubmedArticleSet/PubmedArticle/MedlineCitation')
+
+    for entry in search_results:
+        title = entry.xpath('.//Article/ArticleTitle')[0].text
+
+        # Optional: redirect directly to article page and if no DOI, redirect to PubMed
```
kvch, line 73: Could you make the optional stuff configurable? So instead of commenting lines out, there could be a boolean which is `False` by default and can be configured from here or from `settings.yml`. This way the code can be kept cleaner, and those admins who don't like touching code can set it from config file. Also, if something changes in these lines and an admin who enabled this part has to deal with conflicts, too.

tests/unit/engines/test_pubmed.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertRaises(AttributeError, pubmed.response, '[]')
+
+        response = mock.Mock(text='<PubmedArticleSet></PubmedArticleSet>')
+        self.assertEqual(pubmed.response(response), [])
+
+        xml_mock = """<eSearchResult><Count>1</Count><RetMax>1</RetMax><RetStart>0</RetStart><IdList>
+<Id>1</Id>
+</IdList></eSearchResult>
+"""
+# The engine retrieves abstracts with a second api call, hence testing indirectly
```
kvch, line 32: Could you remove these lines? It makes reading test code harder.

kvch:
All in all I like the idea. It will be a great addition to the existing science engines. After my notes are address, I will be glad to merge it.

searx/engines/pubmed.py
jibe-b: It is the case. It is quite usual to deal with urllib2 and urllib.request but as engines are not supposed to make requests by their own, it has not been used until now.

searx/engines/pubmed.py
jibe-b: I will make it configurable.

Are the parameters in the object

>  - name : pubmed
    parameter : True

accessible from the engine, _i.e._ callable without having to parse settings.yml and match the item with the good name.

Otherwise, I'll simply make it a parameter outside of the engine, but it will be less maintainable.

jibe-b: I finally don't add a parameter in settings. Only display DOI when available. And if a plugin handles DOIs, then it can be made clickable.

jibe-b:
I was lost with mess in my repo so I open a new PR.

https://github.com/asciimoo/searx/pull/1045

PR #1047, opened 2017-09-22
[enh] oa_doi_rewrite plugin broadens doai_rewrite

jibe-b:
Improved version of PR of https://github.com/asciimoo/searx/pull/966

Enables to choose the DOI resolver used in plugin to directly resolve DOIs found in Science tab.

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -22,7 +22,7 @@ if version_info[0] == 3:
 
 logger = logger.getChild('plugins')
 
-from searx.plugins import (doai_rewrite,
+from searx.plugins import (oa_doi_rewrite,
                            https_rewrite,
                            infinite_scroll,
                            open_results_on_new_tab,
@@ -78,7 +78,7 @@ class PluginStore():
 
 
 plugins = PluginStore()
-plugins.register(doai_rewrite)
+plugins.register(oa_doi_rewrite)
 plugins.register(https_rewrite)
 plugins.register(infinite_scroll)
 plugins.register(open_results_on_new_tab)
--- a/searx/plugins/oa_doi_rewrite.py
+++ b/searx/plugins/oa_doi_rewrite.py
@@ -1,13 +1,18 @@
 from flask_babel import gettext
 import re
 from searx.url_utils import urlparse, parse_qsl
+from flask import request
+from searx import settings
+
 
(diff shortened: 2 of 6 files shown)


searx/plugins/oa_doi_rewrite.py
```diff
@@ -21,12 +26,20 @@ def extract_doi(url):
...
+
+
 def on_result(request, search, result):
     doi = extract_doi(result['parsed_url'])
     if doi and len(doi) < 50:
         for suffix in ('/', '.pdf', '/full', '/meta', '/abstract'):
             if doi.endswith(suffix):
                 doi = doi[:-len(suffix)]
-        result['url'] = 'http://doai.io/' + doi
+        result['url'] = get_doi_resolver() + doi
```
asciimoo, line 43: You can pass `request.args` to `get_doi_resolver()` as a parameter from this function, this way you don't need the `from flask import request` line.

searx/plugins/__init__.py
```diff
@@ -22,7 +22,7 @@
 
 logger = logger.getChild('plugins')
 
-from searx.plugins import (doai_rewrite,
+from searx.plugins import (oa_doi_rewrite,
```
asciimoo, line 25: What is the benefit of this rename? What does `oa` mean?

searx/webapp.py
```diff
@@ -164,6 +164,14 @@ def get_locale():
     return locale
 
 
+def get_doi_resolver():
```
asciimoo, line 167: Seems this function is identical to `searx/plugins/oa_doi_rewrite.py:get_doi_resolver()`. You could import that function to avoid code duplication.

asciimoo (changes requested):
Good improvement. Please check the notes.

searx/plugins/__init__.py
jibe-b: OA stands for Open Access, _i.e._ scientific publications that are _at least_ available on the web without a paywall (webpage handling payment of a fee to access another webpage).

doia.net is one DOI resolver service, oadoi.org, doi.org are others. The plugin should have a name that is generic enough.

The plugin could be renamed to doi_resolver_rewrite too.

searx/webapp.py
jibe-b: Sure!

PR #1048, opened 2017-09-23
[add] openaire publication and datasets (twin) engines

jibe-b:
OpenAIRE engines retrieving scientific publications and datasets.

http://openaire.eu/

### DIFF
--- a/searx/engines/openairepublications.py
+++ b/searx/engines/openairepublications.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+"""
+ OpenAIRE (publications) (Scholar publications)
+
+ @website     https://openaire.eu
+ @provide-api yes (http://api.openaire.eu/)
+
+ @using-api   yes
+ @results     json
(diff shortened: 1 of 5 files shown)


searx/engines/openairepublications.py
```diff
@@ -0,0 +1,79 @@
...
+        except:
+            res_dict['publishedDate'] = publishedDate
+
+        try:
+            url = entry_bis['children']['instance']['webresource']['url']['$']
+            res_dict['url'] = url
+        except:
+            pass
+
+#        res_dict = {'url': url,
```
kvch, line 72: Please remove unnecessary comments.

kvch:
Why isn't it implemented as a JSON engine like Crossref or Hoogle? If it can be done, please add it as a JSON engine to `settings.yml`.

jibe-b:
It can surely be done. I hadn't checked json_engine so from now I'll use json_engine whenever possible.

PR #1059, opened 2017-10-06
preferences: make supported/not supported labels

### LINKED ISSUE #558: SafeSearch Boxes Not Changeable
The safesearch buttons under engines cannot be changed, a couple are on and most are off, but they cannot be ticked or unticked. Major problem

kvch:
* replace checkboxes which indicated if an engine supported selected language, safe search and time range search
* add display of time range search support to RTL locales

Closes #558


asciimoo (approved):
👍

PR #1061, opened 2017-10-10
[fix] Language support for Bing Images and Videos

MarcAbonce:
This fixes a current problem where both engines always return their results localized in the instance's country. This could leak the country of origin of a hidden instance.


kvch (approved):

asciimoo:
Great, thanks! Both the PR and the review =)

PR #1066, opened 2017-10-19
[fix] Missing favicon images in oscar and simple themes

MrPetovan:
On my setup:
- Searx 0.12.0
- Apache 2.4.25
- Debian 9

Everything is fine but the favicons (i.e. when a Wikipedia page appears in the results) whose URL features a double-slash `//`, throwing off searx and generating 404 errors.

Ex: https://searx.domain.com/static//themes/oscar/img/icons/wikipedia.png

I found the mistake in the template macro with the parameter `filename` of the function `url_for` that doesn't need a leading `/`. Removing the leading slash fixes the missing image issue.

Successfully tested locally.

### DIFF
--- a/searx/templates/oscar/macros.html
+++ b/searx/templates/oscar/macros.html
@@ -5,7 +5,7 @@
 
 <!-- Draw favicon -->
 {% macro draw_favicon(favicon) -%}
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Thanks for the fix.

PR #1067, opened 2017-10-23
update search api with params and description on syntax

kvch:
This PR adds
- minimal description on how search syntax works
- more parameters to existing search API after exposing every parameter to support saving preferences to search URL


asciimoo (approved):
Great, thanks =)

PR #1078, opened 2017-11-03
results: make engines of results set

kvch:
Previously, if a same result was returned from the same engine, the name of the engine would be added more than once. To eliminate engine name duplication engine list was changed to set.


asciimoo (approved):
Thanks

PR #1082, opened 2017-11-14
Add flake8 to find syntax errors & undefined names

cclauss:
There is at least one print statement that is not Python 3 compatible and many undefined names which may raise NameError at runtime.  Flake8 runs in two passes: The first looks at critical issues in stop-the-build mode and the second looks at style violations in everything-is-a-warning mode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,12 @@ before_install:
 install:
   - ./manage.sh npm_packages
   - ./manage.sh update_dev_packages
-  - pip install codecov
+  - pip install codecov flake8
+before_script:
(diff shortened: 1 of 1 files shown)


asciimoo (commented):
Perhaps this could be in `manage.sh`. It already does static code checks and it would be useful to catch these errors locally. What do you think?

cclauss:
Sure.  __manage.sh flake8__ would make a lot of sense as long as it comes after flake8 is pip installed.

We need to fix the issues raised at https://travis-ci.org/asciimoo/searx/builds/301878446 before this can go live anyway.

PR #1084, opened 2017-11-15
fix bug: TypeError: set(['bing', 'google']) is not JSON serializable,…

moon2l:
… when curl 'http://127.0.0.1:8888/?q=xxxxxxxx&categories=general&format=json'


asciimoo (approved):
Nice finding, thanks!

PR #1090, opened 2017-11-22
Dependencies upgrade

dalf:
https://pypi.python.org/pypi/pyOpenSSL : fix memory leak
http://lxml.de/4.1/changes-4.1.1.html : speed improvement


asciimoo (approved):
👍 great, thanks!

PR #1105, opened 2017-11-30
Fix time range selection during search

### LINKED ISSUE #1102: Time range filter doesn't work on Chrome browser
on Chrome browser nothing happens when changing value of this filter

kvch:
The previous version did not work in Chrome browser.

Closes #1102


asciimoo (approved):
Great, thank you =]

PR #1109, opened 2017-12-05
simple theme : magnet.svg is replaced by a home made version

dalf:
This PR replaces magnet.svg with a homemade version to solve the license issue.
see comment https://github.com/asciimoo/searx/issues/1091#issuecomment-348020765


asciimoo (approved):
Thanks!

PR #1112, opened 2017-12-10
[mod] simplify manage.sh

dalf:
build : 
- ```npm install less@2.7 less-plugin-clean-css grunt-cli``` is moved inside manage.sh
- if it is possible to source manage.sh to update the PATH (to start grunt)
- tidy the help documentation
- update the geckodriver version to 0.19.1


asciimoo (approved):
👍 Thanks

PR #1124, opened 2017-12-19
Make Python 3 able to read settings files with Unicode characters

JosephKiranBabu:
SearX currently doesn't start up when run with Python 3 as it tries to parse the
settings.yml file with ASCII codecs.
Python 3 requires that files with Unicode characters be read with a 'b' flag.
This also works with Python 2 and hence can be integrated into the main source
code.

Tested with the latest Python 3.6.4rc1 on Debian unstable.

Signed-off-by: Joseph Nuthalapati <njoseph@thoughtworks.com>

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -50,7 +50,7 @@ if not settings_path:
     raise Exception('settings.yml not found')
 
 # load settings
-with open(settings_path) as settings_yaml:
+with open(settings_path, 'rb') as settings_yaml:
     settings = load(settings_yaml)
 
 '''
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -36,7 +36,7 @@ engines = {}
 
 categories = {'general': []}
 
-languages = loads(open(engine_dir + '/../data/engines_languages.json').read())
+languages = loads(open(engine_dir + '/../data/engines_languages.json', 'rb').read())
 
 engine_shortcuts = {}
 engine_default_args = {'paging': False,
--- a/searx/engines/currency_convert.py
+++ b/searx/engines/currency_convert.py
@@ -94,7 +94,7 @@ def load():
     global db
 
     current_dir = os.path.dirname(os.path.realpath(__file__))
-    json_data = open(current_dir + "/../data/currencies.json").read()
+    json_data = open(current_dir + "/../data/currencies.json", 'rb').read()
 
     db = json.loads(json_data)
 


kvch:
Just curiosity: which character causes the problem?

JosephKiranBabu:
@kvch Sorry, I should have posted the error log.
I've installed Searx 0.13.1+dfsg1 from Debian (unstable). The Python version is 3.6.4~rc1-2 which is also a Debian package.
After installation, I copied the settings.yml from examples into /etc/searx/ and then ran SearX using /usr/bin/searx-run.


```
root@freedombox:~# /usr/bin/searx-run 
  File "/usr/bin/searx-run", line 11, in <module>
    load_entry_point('searx==0.13.1', 'console_scripts', 'searx-run')()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 572, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2752, in load_entry_point
    return ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2405, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2411, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3/dist-packages/searx/__init__.py", line 54, in <module>
    settings = load(settings_yaml)
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 72, in load
    return loader.get_single_data()
  File "/usr/lib/python3/dist-packages/yaml/constructor.py", line 35, in get_single_data
    node = self.get_single_node()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 133, in compose_mapping_node
    item_value = self.compose_node(node, item_key)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 82, in compose_node
    node = self.compose_sequence_node(anchor)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 111, in compose_sequence_node
    node.value.append(self.compose_node(node, index))
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
  File "/usr/lib/python3/dist-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/usr/lib/python3/dist-packages/yaml/parser.py", line 428, in parse_block_mapping_key
    if self.check_token(KeyToken):
  File "/usr/lib/python3/dist-packages/yaml/scanner.py", line 116, in check_token
    self.fetch_more_tokens()
  File "/usr/lib/python3/dist-packages/yaml/scanner.py", line 156, in fetch_more_tokens
    self.scan_to_next_token()
  File "/usr/lib/python3/dist-packages/yaml/scanner.py", line 777, in scan_to_next_token
    self.forward()
  File "/usr/lib/python3/dist-packages/yaml/reader.py", line 101, in forward
    self.update(length+1)
  File "/usr/lib/python3/dist-packages/yaml/reader.py", line 153, in update
    self.update_raw()
  File "/usr/lib/python3/dist-packages/yaml/reader.py", line 178, in update_raw
    data = self.stream.read(size)
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd8 in position 2584: ordinal not in range(128)
```

JosephKiranBabu:
If the Unicode characters are removed from the settings.yml file then startup fails again - with the same error - as data/engines_languages.json has Unicode characters too.

If my fix is not comprehensive, kindly consider making changes in any other files where this might be a problem.

JosephKiranBabu:
@kvch This is the specific section in settings.yml that has Unicode characters.

```yml
locales:
    en : English
    ar : العَرَبِيَّة (Arabic)
    bg : Български (Bulgarian)
    cs : Čeština (Czech)
    da : Dansk (Danish)
    de : Deutsch (German)
    el_GR : Ελληνικά (Greek_Greece)
    eo : Esperanto (Esperanto)
    es : Español (Spanish)
    fi : Suomi (Finnish)
    fr : Français (French)
    he : עברית (Hebrew)
    hr : Hrvatski (Croatian)
    hu : Magyar (Hungarian)
    it : Italiano (Italian)
    ja : 日本語 (Japanese)
    nl : Nederlands (Dutch)
    pt : Português (Portuguese)
    pt_BR : Português (Portuguese_Brazil)
    ro : Română (Romanian)
    ru : Русский (Russian)
    sk : Slovenčina (Slovak)
    sl : Slovenski (Slovene)
    sr : српски (Serbian)
    sv : Svenska (Swedish)
    tr : Türkçe (Turkish)
    uk : українська мова (Ukrainian)
    zh : 中文 (Chinese)
```
Removing everything except `en : English` fixes the problem.
These language strings are also present in data/engines_languages.json

JosephKiranBabu:
I have handled similar issues in two more files - engines_languages.json and currencies.json
This fixes the Unicode character problems in all the files now.

josch:
The bug was also reported for the Debian searx package: https://bugs.debian.org/886093

Our CI system as well as my own live instance does not have any Python 2 packages installed. Only Python 3 packages are installed on both systems. Thus, it is unlikely that this problem occurs because of Python 3. See the Debian bts link for more details of my analysis.

JosephKiranBabu:
@josch 

I am trying to run searx on this box.
https://app.vagrantup.com/freedombox/boxes/plinth-dev

I noticed that searx failed to run when I was logged in as `root` user, but works with `vagrant` user. It was because the locale settings were different. The vagrant user has utf8 in their locale but not the root user.

This issue can be reproduced in the following cases:
```
$ LC_ALL=C python3 -c 'import searx'
$ LC_ALL=POSIX python3 -c 'import searx'
```

Maybe the CI server supports utf8 in its locale. 

In short, searx doesn't work when the default locale on the server doesn't support Unicode.
This pull request makes the application independent of the locale settings of the system on which it is run.

josch:
Okay. But I still don't see how your patch is the proper fix for this issue. What you do is to just treat the input as pure binary. But what we want is to read the file in utf-8 encoding. So would it not be better to "fix" this properly by doing:

    with codecs.open("settings.yml", "r", "utf-8")) as f:
        yaml.load(f)

But I still don't think that the current behaviour of searx even classifies as a bug. If the environment sets `LC_ALL=C` then this is to tell all applications that are run in this environment to apply this locale setting. By forcing searx to ignore this and use binary or utf8 instead, you essentially take away from the user the ability to run searx in the locale of their choosing. You effectively say that searx is limited to utf8 only.

I don't see yet, why the fix has to happen in searx and why the proper solution to this problem is not for you to set your environment into the state that you want it. It seems obvious to me that you want to run applications with utf8 support, so why did you configure your OS with the C locale? To me it sounds like you want to fix a problem with the configuration of your OS in the application that you are running. You will be surprised when you find other applications that actually respect your locale settings. Do you also want to submit patches to them so that they also ignore your locale setting?

I suggest you just adjust your system configuration to use the locale that you prefer. For you, that might be `C.utf8`.

Disclaimer: I'm just the Debian maintainer of the searx package. The opinion of the searx developers on this topic might differ from mine. If searx developers think it's a good idea to ignore the system's locale setting and force utf8 then so be it.

SunilMohanAdapa:
I believe this patch is correct. My reasoning follows:

Locales:

- Locale is a way to tell a program how to interact with the user, that is,
  which currency to use, how to format a date and which language and character
  encoding to use. This applies when showing messages, dates, currencies etc.
  and when accepting input from the user. It does not dictate how a program
  should do internal processing of strings and configuration parsing. These
  files being read here are internal workings of the program, especially in case
  of currencies and engine configuration. They are not user interaction.

- Searx has *chosen* to store the files in question as UTF-8 not as Ascii,
  UTF-16 or UTF-32. Both JSON and YAML allow for UTF-8, UTF-16 and UTF-32[1][2].
  As a consequence, Searx *will have to* read them as UTF-8 irrespective of how
  it is interacting with the user (or which locale is set). Setting a locale
  does not mean that Searx should read its own internal files in that locale.
  Most programs operate internally in UTF-8 and interact with users in various
  locales.

- For a web program like Searx, the locale set in the environment is of much
  less use. It has to interact with users of multiple locales at once. Locale is
  specified on a per-request basis via HTTP header or via preference set in the
  session. This is unlike the command line tools and GUI programs that interact
  with a single user at a time and rely on locale set in the environment to read
  user preference. Further, choosing to read an internal file as UTF-8 does not
  mean Searx will not be serving users of different locales.

- I don't believe setting a locale to something other than UTF-8 means that
  system has no Unicode support. What does having 'system having unicode
  support' mean anyway in this context? We have programming language, its
  libraries, the HTTP protocol and i18n/l10n library all supporting Unicode
  unconditionally.

Changing the system default locale:

- I just installed Debian from a latest testing ISO and during installation, I
  selected 'No localization' selected instead of the default 'English'. I
  installed 'searx' package and `python3 -c "import searx"` failed. The locale
  in this case is 'C' and not 'C.utf8'. We will have go and update Debian's (and
  all other such distributions') default locale to fix this problem.

- Say there is another daemon in the system that has its configuration files
  encoded in UTF-16 (for some reason). It would not be possible to set the
  default locale of the system to UTF-8 (for the sake of Searx) and UTF-16 (for
  the sake of the other program) at the same time. This means that Searx will
  have to be passed its locale separately during start. In this case, it is best
  done in startup files (.service file or /etc/init.d scripts). Should there be
  a provision such that administrators can set Searx's locale separately by
  editing startup configuration?

- Different users on the system have different locales set in their environment.
  System's default locale is simply a way to set, by default, what users are
  most like to choose. On Debian, of the 492 locales, 189 were not using UTF-8
  as character encoding. People use different default locales for great many
  reasons. I don't think it is intended that Searx should catastrophically fail
  in all of those cases, especially when it need not.

Python's behavior:

- Please note the following change in Python's behavior from Python 2 to
  Python 3.

```
$ LC_ALL=en_US.UTF-8 python2 -c "open('hello.txt', 'r').read()"
$ LC_ALL=C python2 -c "open('hello.txt', 'r').read())"
$ LC_ALL=en_US.UTF-8 python3 -c "open('hello.txt', 'r').read()"
$ LC_ALL=C python3 -c "open('hello.txt', 'r').read()"
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe0 in position 0: ordinal not in range(128)
```

- When opening a file in Python 3, the file encoding is assumed from the current
  locale. So, the only correct way to read a known utf-8 file in all locales is
  to explicitly specify its encoding as utf-8.

Alternatives:

- The suggestion to use codecs is also correct. Python seems to use this library
  internally doing various kinds of string decoding.

- Yet another way of doing this is `open(file_name, encoding='utf-8')`.

- All three solutions are roughly equivalent. In case of opening in binary,
  UTF-16 and UTF-32 encoded files are also accepted as supported by JSON and
  YAML specifications and their corresponding libraries.

Links:
1) https://tools.ietf.org/html/rfc7159#page-9
2) http://www.yaml.org/spec/1.2/spec.html#id2771184

asciimoo:
@SunilMohanAdapa thanks for the detailed explanation, you are right.

asciimoo (approved):
Thank you.

SunilMohanAdapa:
@JosephKiranBabu @josch @asciimoo Thank you all very much.

Looking forward to having Searx in FreedomBox real soon :)

josch:
But even with the explanation by @SunilMohanAdapa, is the correct solution to this problem not opening the file with `codecs` and then forcing utf-8 instead of reading it with `b` as suggested by this pull request?

SunilMohanAdapa:
@josch, If we open as binary, we will be supporting all codecs valid under the YAML/JSON specifications (UTF-8, UTF-16, UTF-32). If an administrator creates a /etc/searx/settings.yml with UTF-16 encoding searx should not fail to load it as it is a valid YAML file.

When given binary data, JSON and YAML parsers interpret them with the correct encoding automatically.

Consider the following where binary works better than 'utf-8':
```python
>>> import yaml
>>> open('/tmp/test.yml', 'w', encoding='utf-16').write('a: b')
4
>>> content = open('/tmp/test.yml', encoding='utf-8').read()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
>>> content = open('/tmp/test.yml', 'rb').read()
>>> content
b'\xff\xfea\x00:\x00 \x00b\x00'
>>> yaml.load(content)
{'a': 'b'}
```

josch:
I was unaware that yaml would automagically detect the right codec! That's wonderful! In that case I take back everything I said. Thank you for taking the time to cure my ignorance on the matter! :)

JosephKiranBabu:
Closed the bug on Debian BTS.
https://bugs.debian.org/886093

JosephKiranBabu:
@asciimoo Can we have a patch release with this fix included? Searx is still unusable for my use case.
If a release is difficult, can @josch  please make a new patch release for the Debian package?

josch:
@JosephKiranBabu depending on the reply from @asciimoo I'll make a new release of the Debian package. This is because there is no point of me backporting the patch if there is a new release around the corner.

asciimoo:
> Can we have a patch release with this fix included?

Sure, I'm planning to fix some engines and we can release at the end of this week.

JosephKiranBabu:
Thanks for making the release as promised. Eagerly waiting for the release of the Debian package.

josch:
Version 0.14.0 was just packaged and uploaded to unstable.

PR #1127, opened 2017-12-20
remove the option of sourcing manage.sh

kvch:
From now on, there is no point in sourcing `manage.sh`.

My `/bin/sh` does not support this form of detection if a script is sourced or not. Thus, I am not able to run any of the commands, because it always returns before performing anything. Because of incompatibility, I would like to remove it.
Another solution could be changing `/bin/sh` to `/bin/bash`, so sourcing would work. However, `node_modules/.bin` is placed under the folder `searx` which is not a universal path to grunt. In my case it is under a different directory. I propose to let the developer set his/her own `PATH` without any interference from `manage.sh`.
Removing this part ensures that the script can be run without any problem and lets developers manage their own environment.

What do you think @dalf?

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -3,14 +3,6 @@
 BASE_DIR="$(dirname -- "`readlink -f -- "$0"`")"
 export PATH="$BASE_DIR/node_modules/.bin":$PATH
 
-# the script can be sourced to update the PATH
-# see https://stackoverflow.com/questions/2683279/how-to-detect-if-a-script-is-being-sourced
-if [ $_ != $0 ]; then
-    unset BASE_DIR
-    # sourced : exit now
-    return
-fi
-
 # subshell
 PYTHONPATH="$BASE_DIR"
 SEARX_DIR="$BASE_DIR/searx"
--- a/manage.sh
+++ b/manage.sh
@@ -48,7 +48,7 @@ install_geckodriver() {
 
     if [ -z "$1" ]; then
         if [ -z "$VIRTUAL_ENV" ]; then
-            echo "geckodriver can't be installed because VIRTUAL_ENV is not set, you should download it from\n  $GECKODRIVER_URL"
+            printf "geckodriver can't be installed because VIRTUAL_ENV is not set, you should download it from\n  %s" "$GECKODRIVER_URL"
             exit
         else
             GECKODRIVER_DIR="$VIRTUAL_ENV/bin"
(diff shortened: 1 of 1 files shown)


dalf:
I've to admit it's a little experiment.

I've a local uncommitted / unpushed change that add autocompletion to a ```searx``` command.

I will post this version here for information. 

About the node_modules/.bin path, I've found a command which give the installation path, but I don't remember which one.

[EDIT] ```npm root```

kvch:
Does it mean that you agree with my proposed changes?

Which `searx` command do you mean?

Having a command to find the installation path would be nice.

dalf:
I suggest to 

1/ replace ```/bin/sh``` by ```/bin/bash```

2/ replace
```sh
export PATH="$BASE_DIR/node_modules/.bin":$PATH
```
by 
```sh
```
see https://docs.npmjs.com/cli/bin

dalf:
The unfinished not fully tested version :  https://gist.github.com/dalf/0afe985fd7129cf7aa731eecbf681284

As I said it's an experiment.

dalf:
The safe way to detect sourced file seems to be : 
```sh
sourced=0
if [ -n "$ZSH_EVAL_CONTEXT" ]; then 
  case $ZSH_EVAL_CONTEXT in *:file) sourced=1;; esac
elif [ -n "$KSH_VERSION" ]; then
  [ "$(cd $(dirname -- $0) && pwd -P)/$(basename -- $0)" != "$(cd $(dirname -- ${.sh.file}) && pwd -P)/$(basename -- ${.sh.file})" ] && sourced=1
elif [ -n "$BASH_VERSION" ]; then
  [ "$0" != "$BASH_SOURCE" ] && sourced=1
else # All other shells: examine $0 for known shell binary filenames
  # Detects `sh` and `dash`; add additional shell filenames as needed.
  case ${0##*/} in sh|dash) sourced=1;; esac
fi
```

(according to https://stackoverflow.com/questions/2683279/how-to-detect-if-a-script-is-being-sourced )

funkyfuture:
maybe i'm missing some point, but i suggest to consider to use `make`.

kvch:
I rebased and updated the PR with changing `echo` to `printf` as suggested by https://github.com/asciimoo/searx/issues/1206#issue-298593155.

asciimoo (approved):
Great, thanks!

PR #1135, opened 2018-01-06
Fix yacy engine

### LINKED ISSUE #1064: Yacy results crash
Getting:
Engines cannot retrieve results:
yacy (unexpected crash)

> ERROR:searx.search:engine yacy : exception : 'url'
> Traceback (most recent call last):
>   File "/home/leo/searx/searx/search.py", line 118, in search_one_request_safe
>     search_results = search_one_request(engine, query, request_params, start_time, timeout_limit)
>   File "/home/leo/searx/searx/search.py", line 110, in search_one_request
>     return engine.response(response)
>   File "/home/leo/searx/searx/engines/yacy.py", line 80, in response
>     results.append({'url': result['url'],
> KeyError: 'url'

kvch:
Previously, "url" was not found in the result json, so the engine crashed. Now the engine looks for both "url" and "link" in results to get the URL of a result.

Closes #1064

### DIFF
--- a/searx/engines/yacy.py
+++ b/searx/engines/yacy.py
@@ -74,8 +74,17 @@ def response(resp):
     for result in search_results[0].get('items', []):
         # parse image results
         if result.get('image'):
+
+            result_url = ''
+            if 'url' in result:
+                result_url = result['url']
+            elif 'link' in result:
+                result_url = result['link']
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
great, thanks.

PR #1136, opened 2018-01-06
Add findx engine

kvch:
Does not include images, videos and maps category.


asciimoo:
Findx could be a dedicated engine to be able to extract image, video and map results as well.

asciimoo (approved):
Great, thank you =)

PR #1148, opened 2018-01-17
[fix] Read utf-8 files (settings, languages, currency) with Python3.5

MarcAbonce:
For some reason, the fix in #1124 causes an [error when using Python3.5](https://travis-ci.org/MarcAbonce/searx/builds/329737381?utm_source=github_status&utm_medium=notification).
Python3.5 is still the default Python3 version in the latest stable releases of [Debian (Stretch)](https://packages.debian.org/source/stretch/python3-defaults) and [Ubuntu (16.04 LTS)](https://packages.ubuntu.com/source/xenial/python/python3-defaults), so I added a Travis test to cover that Python version for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
 language: python
 python:
   - "2.7"
+  - "3.5"
   - "3.6"
 before_install:
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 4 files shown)


asciimoo (approved):
Good to know. Thanks for the fix.

PR #1190, opened 2018-02-13
Change alpine version to permit ARM and amd64 deployment 

Themimitoof:
Hello,

I edited the Dockerfile to permit the deployment of Searx on _amd64 (x86_64) and ARM architectures_ (for example: Scaleway and SBCs like Raspberry). The alpine-3.5 version only supports x86 architecture but the 3.6 and later purpose x86_64, armhf and arm64 architectures.

I checked if errors occured when I built the docker image and no error in _Scaleway ARM64_ instance and on a _Raspberry Pi 3_.


asciimoo (approved):
Thanks!

Issue #1183, opened 2018-02-10
Problems with JSON search engines


scroom:
Hallo,

yesterday I've set up an searx instance. Thereby I discovered some problems with some of the search engines. I've got the following error messages:

```
Die folgenden Suchmaschinen können die Ergebnisse nicht empfangen:
soundcloud (unexpected crash: No JSON object could be decoded)
```

```
Die folgenden Suchmaschinen können die Ergebnisse nicht empfangen:
microsoft academic (unexpected crash: No JSON object could be decoded)
```

`google images (unexpected crash: No JSON object could be decoded) `

kvch:
Google images is working for me. Could you provide more details on how to reproduce the issue? E. g. query, instance name, etc.

zuglufttier:
I have the same problem, this is from the logs on my own server:
```
ERROR:searx.search:engine google images : exception : No JSON object could be decoded
Traceback (most recent call last):
  File "/usr/local/searx/searx/search.py", line 104, in search_one_request_safe
    search_results = search_one_request(engine, query, request_params)
  File "/usr/local/searx/searx/search.py", line 87, in search_one_request
    return engine.response(response)
  File "/usr/local/searx/searx/engines/google_images.py", line 69, in response
    g_result = loads(resp.text)
  File "/usr/local/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
ERROR:searx.search:engine 500px : exception : 'photos'
Traceback (most recent call last):
  File "/usr/local/searx/searx/search.py", line 104, in search_one_request_safe
    search_results = search_one_request(engine, query, request_params)
  File "/usr/local/searx/searx/search.py", line 87, in search_one_request
    return engine.response(response)
  File "/usr/local/searx/searx/engines/www500px.py", line 56, in response
    for result in response_json['photos']:
KeyError: 'photos'
```

LP0101:
Any updates for when the next release will be happening? I rely on the Dockerfile for my instance, so I'll need 0.15.0 to be released to get these fixes.

kvch:
Fix was released in the latest version.

PR #1199, opened 2018-02-17
Fix Microsoft Academic engine

kvch:
This PR fixes the existing Microsoft Academic engine.
It was reported to be broken in #1183.


asciimoo (approved):
Great, thanks!

PR #1226, opened 2018-03-05
Fix Vim mode on Firefox

cozybuild:
Firefox automatically opens a built-in quick search when user
starts typing without selected input and the Vim shortcuts
are not triggered


asciimoo (approved):
Thanks for the fix!

PR #1227, opened 2018-03-05
[UI] fix custom-select in oscar

Pofilo:
Fixes https://github.com/asciimoo/searx/issues/1223.


Pofilo:
Well, result is not so good as well ..

image

Should it be larger (or adding a padding to the container) ?

asciimoo (approved):
Thanks for the fix!

PR #1233, opened 2018-03-10
Use new LABEL syntax for Dockerfile

angristan:
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


asciimoo (approved):
Thank you.

PR #1238, opened 2018-03-15
[fix] Change duckduckgo url to avoid error response

MarcAbonce:
This error is hard to reproduce, but sometimes requests to Duckduckgo will return an error.
Removing some parameters on the search URL seems to fix the error.

For example:
https://duckduckgo.com/html?q=query&api=/d.js&o=json will sometimes return an error message.
https://duckduckgo.com/html?q=query should never fail.


asciimoo (approved):
Great, thanks!

PR #1249, opened 2018-03-24
use base_url for prefix of sharing 'currenly saved preferences'

matejc:
if behind reverse proxy, http://localhost:8888/ is a prefix, this fix should solve this problem


Repooc:
Would like to see this added as the ip address is showing for my server instead of the base_url

asciimoo (approved):
Great, thanks!

PR #1260, opened 2018-04-09
[fix] Engine fixes

MarcAbonce:
* Yahoo: Make all results appear.
* Google news: Fix results xpath.
* Wikidata: Make search more accurate by using the [generic search](https://www.wikidata.org/w/index.php?search=&title=) instead of [searching by label](https://www.wikidata.org/wiki/Special:ItemDisambiguation).


asciimoo (approved):
Great, thanks!

PR #1268, opened 2018-04-13
first galician (gl) translation

xmgz:
first .po file for galician (gl) translation.

The commit contains why I had to translate it here and not in transifex.

Please consider merging it on the project. Thanks


kvch:
I accepted your language request on Transifex. Do you mind submitting the translation there?

PR #1273, opened 2018-04-19
[fix] Force English results in Google when using en-US

MarcAbonce:
This should remove the need to use en-GB in place of en-US results.


asciimoo (approved):
Great, thanks!

sven222:
I still have the captcha issue. Updated with git to the last master branch and I do not receive any search results. Just the Captcha messsage in red. Can I provide you some log files?

Issue #1209, opened 2018-02-21
Splitting settings.yml


josch:
Hi,

currently, `settings.yml` maintains the site-specific *settings* (language, secret key, proxy settings...) as well as the list of engines and list of languages.

Problems arise when upgrading searx from one version to the next. While settings like language, secret key and proxy settings probably should stay the same, the `settings.yml` has to be adapted to incorporate any added or removed languages or engines. This means, that currently, upgrading searx involves some manual work.

This manual work could be avoided if `settings.yml` were somehow split into more than one file. I can imagine two scenarios:

`settings.yml` only contains the site-specific settings (like language, secret key, proxy settings) while a new configuration file `engines.yml` contains the list of engines and `languages.yml` contains the list of languages. `settings.yml` would live in `/etc/searx` because it is definitely a site-specific configuration option which does *not* change on each upgrade. `engines.yml` and `languages.yml` could live in `/usr/share/searx` because those files *do* change upon each upgrade. Site admins who wish to override their `engines.yml` or `languages.yml` could add these files into `/etc/searx` but would then naturally have to take care to keep these up to date with their specific requirements when upgrading searx. searx itself would then prefer files in `/etc/searx` over files in `/usr/share/searx` but fall back to the latter if the former doesn't exist.

The second scenario, similar to the first one, maintains a minimal `settings.yml` in `/etc/searx` and a `languages.yml` in `/usr/share/searx` but instead of putting engine information into a `engines.yml`, each engine would have their own small yml file in `/usr/share/searx/engines`. Adding and removing engines by searx upgrades would then work by adding and removing files from that directory. If a system admin wants to overwrite the available engine, then they can create a `/etc/searx/engines.d` which contains all engine files that they want active. These could just be symlinks to the desired files in `/usr/share/searx/engines`. This concept is similar to what for example apache does with its `sites-enabled` and `sites-available` directories.

What do you think?

unixfox:
Maybe related to https://github.com/searx/searx/issues/2252

PR #1278, opened 2018-04-22
Implement #1209

dalf:
Implement #1209 :
- settings.yml has the same format as before
- doi.yml, locales.yml and engines.yml are loaded before (the disabled attributes are removed)
- if yaml search.engines path is defined, then all engines are disabled, and only engine specified in this list are enabled.
- settings.yml contans only general, search, server, ui sections. The search section contains the enabled engines.
- settings_robot.yml is unchanged


dalf:
Some notes : 
* In this implementation an administrator can't define its own engine. I think that "search.engines_yaml" and "search.include_engines_yaml" should be added in settings.yml. The first one to replace engines.yml path, the second to add another engines.yml.
* May be doi.yml, locales.yml, and engines.yml should be moved to the same directory than currencies.json and engines_languages.json ?

dalf:
Add two new settings : 
* search.engines_path : path of engines.yml
* search.include_engines_path : additional engine definitions (doesn't replace the content of engines.yml)
    
doi.yml, engines.yml, locales.yml are loaded directly into doi, engines, locales (not more prefix). Avoid to have settings about engines inside doi.yml for example. 

Note that the DOI settings have changed.

Some bug fixes on oa_doi_rewrite.

jfowl:
What do the core contributers think about this? I think this is a very smart idea all in all.

asciimoo:
Separating the engines part of the settings is a good idea, but the other files (languages, doi) aren't really necessary. What is the advantage of spreading the settings to multiple small files?

dalf:
Replaced by #2291

PR #1279, opened 2018-04-22
Switched currency converter to duckduckgo and added Bitcoin to currency list

rinpatch:
Switched currency converter to duckduckgo because parsing google finance is not working anymore (when trying to request https://finance.google.com/finance/converter?a=1&from=USD&to=EUR it either responds with "finance search result" or times out).
Also added Bitcoin to currency list since it is supported by duckduckgo's currency converter


asciimoo (approved):
Great, thank you!

PR #1283, opened 2018-04-27
[Feature] Acgsou as a searchable engine

rinpatch:
Includes acgsou.com as a searchable engine. Doesn't show seed/leech/download information because, as I figured out, for some reason it is randomly generated on client's side every time the page is reloaded


asciimoo (approved):
Great, thank you.

return42:
@rinpatch https://www.acgsou.com/ is redirected to 36dm.club .. can you confirm, that Acgsou is dead?

If so, I would like to remove the engine / or did you know about a replacement for?

rinpatch:
@return42 not sure if it's dead, but I do not plan on maintaining the engine, so feel free to remove it.

PR #1289, opened 2018-05-03
Switch acgsou to http

rinpatch:
On May 1'st acgsou disabled https and now when trying to request using https it times out


asciimoo (approved):
:+1: Thanks

asciimoo:
Please fix the URL in the test too, because travis fails.

PR #1290, opened 2018-05-05
Fix acgsou encoding error on python2

rinpatch:
Acgsou failed to retrive results on python 2.7 since default text type in it is str and not unicode like in python3. This pr fixes it and adds unicode characters into acgsou unit test


asciimoo (approved):
👍 Thanks.

PR #1303, opened 2018-05-20
Fix bing "garbage" results

### LINKED ISSUE #1275: Is Bing feeding garbage results to searx?
I've noticed most results from bing are irrelevant. Not even based on any of the words I use on my queries.

For example, the query "youtube dash stream full hd" returns as a frist results a tesco link to buy a tv. I guess the "full hd" triggered that maybe, but if you scroll down the 7th result is a link to the FBI podcast (???). For that particular query out of 7 bing results only 1 is maybe relevant.

Here are other query examples that yield garbage resuts.

Query: easy privacy list . 1st result is a page called muscchoice, completely irrelevant. 3rd bing result is a page called growweedeasy, again nothing to do with my query. out of 10 bing results for this query once again maybe just one is partially relevant.

Query: adblock plus rules div tag. This time the first result is correct from both bing and google, but the 4th one (3rd bing result) is for a page called globolister. with the title "vote for icefilmsdotinfo" which is unrelated to my query

Query: "reply all" podcasts. For this one I quoted the term "reply all" which is the name of the podcast. and the very first result from bing point to buzzfeed podcast. I feel like this is just plain advertisement

Query: can vitamin c audio. For a more confusing term, since the the query contains words used in other contexts, the first bing result links to amazonand what  I'm actually looking for is a song. google gets it right .

Query: bert jansch "needle of death" lyrics. Another quoted query, the first result is from a page called popmatters talking about another song, and it doesn't even contain the lyrics. the 3rd link is from a newspaper that makes a small reference to the singer,nothing else. 4th result is for a page called michaelchapman, irrelevant. Out of 6 bing results, again, maybe a couple are partially relevant, none are what I'm looking for.

I could keep adding queries but I think I made my point very clear. In my opinion Bing is feeding garbage to searx. Or maybe it is just that bing is that bad. So I would like to request the removal of Bing as a default search engine

MarcAbonce:
This should fix #1275.

For some reason, it looks like using only Windows based user-agents gets the good results, while using a Linux based user-agent leads to more inconsistent results. Strangely enough, I can't reproduce this bug in a browser. I tried changing other HTTP headers, but didn't find any significant improvement.

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -16,7 +16,7 @@
 from lxml import html
 from searx.engines.xpath import extract_text
 from searx.url_utils import urlencode
-from searx.utils import match_language
+from searx.utils import match_language, gen_useragent
 
 # engine dependent config
 categories = ['general']
@@ -43,6 +43,9 @@ def request(query, params):
         offset=offset)
 
     params['url'] = base_url + search_path
+
+    params['headers']['User-Agent'] = gen_useragent('Windows NT 6.3; WOW64')
+
     return params
 
 
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -57,9 +57,9 @@ blocked_tags = ('script',
                 'style')
 
 
-def gen_useragent():
+def gen_useragent(os=None):
     # TODO
-    return ua.format(os=choice(ua_os), version=choice(ua_versions))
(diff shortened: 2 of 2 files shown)


infosisio:
I had noticed that too, glad someone fixed it !

asciimoo (approved):
Great. thanks!

PR #1321, opened 2018-06-14
Open Semantic Search Engine

### LINKED ISSUE #1318: Support for Open Semantic Search Integration
We use Open Semantic Search, which is arguably the best free & open source intranet search engine out there. Adding a plugin that gets the results from the users local Open Semantic Search instance (for those who host a searx instance only I assume) would be of tremendous value. We are willing to sponsor the task, if necessary and there is interest. Thanks a lot.

nicholasks:
This new engine will be using Solr search server used by an Open Semantic Search setup.

This should close #1318 .


searx/engines/opensemantic.py
```diff
@@ -0,0 +1,47 @@
...
+ @stable     yes
+ @parse      url, title, content, publishedDate
+"""
+from dateutil import parser
+from json import loads
+from searx.url_utils import quote
+
+# If your OSS is hosted on another machine you would need to change
+# your base_url.
+base_url = 'http://localhost:8983/solr/opensemanticsearch/'
```
Pofilo: Is this the classical URL given by an installation of Open Semantic Search ?
You should maybe put it in `settings.yml`. This way, the configuration is only done in the configuration file.
(So you can remove the comment maybe)

What do you think ?

Pofilo (changes requested):
base_url in configuration

Pofilo (approved):

searx/settings.yml
```diff
@@ -464,6 +464,11 @@ engines:
     shortcut : oap
     timeout: 5.0
 
+  - name : opensemanticsearch
```
kvch, line 467: Engines which require action (installation, setting base_url) before being usable are usually commented out. See yacy as an example. Could you please comment it out?

searx/engines/opensemantic.py
```diff
@@ -0,0 +1,45 @@
...
+    for current in docs:
+        item = {}
+        item['url'] = current['id']
+        item['title'] = current['title_txt_txt_en']
+        if current.get('content_txt'):
+            item['content'] = current['content_txt'][0]
+        item['publishedDate'] = parser.parse(
+            current['file_modified_dt']
+        )
+        item['originalData'] = current
```
kvch, line 42: Why is this needed? AFAIK `originalData` is not used anywhere.

PR #1330, opened 2018-06-23
[fix] Update user agent versions

MarcAbonce:
DuckDuckGo is currently returning error on every query.
Updating the user agent to more recent versions seems to fix this.
I think this would also make #1303 redundant, since this is a cleaner version of the same fix.


dalf:
Side note : there is #933 too.

MarcAbonce:
I hadn't noticed that pull request. It would probably be better to merge that one instead.

asciimoo (approved):
Thanks!

Pofilo:
Shouldn't we merge https://github.com/asciimoo/searx/pull/933 instead of this one ?

asciimoo:
@Pofilo  it would be the futureproof solution, but there are still unresolved questions.

PR #1355, opened 2018-08-01
manage.sh : minor change

dalf:
Use ```npm bin``` rather than the hard coded ```node_modules/.bin``` path.


asciimoo (approved):
Thanks.

dadosch:
I get `./manage.sh: 8: ./manage.sh: npm: not found` when I run manage.sh. *(But I'm unsure wether this is because of a broken installation or not)*

PR #1363, opened 2018-08-06
Fix pronunciation on README

beatgammit:
The previous pronunciation read "sucks", now this reads "surks", which is also in [Wikipedia's entry](https://en.wikipedia.org/wiki/Searx).


kvch:
Thanks for the contribution. We use British English, so the "r" is silent. See: https://github.com/asciimoo/searx/issues/1264
The page on Wikipedia has to be fixed.

beatgammit:
Lol, nice. Does it sound like "sucks" in British English too? Because `searx` definitely *doesn't* suck. :smile:

PR #1366, opened 2018-08-10
Fix etymonline engine

### LINKED ISSUE #1361: The following engines don't work
* etymonline
* gigablast (#1293)

kvch:
* follow up changes in xpath expressions
* removed suggestions as they now longer show up on the results page but on the "details" page of each result

Partially fixes #1361


asciimoo (approved):
Great, thanks!

PR #1369, opened 2018-08-12
Change the local name of the Polish language

LiquidLemon:
"Polszczyzna" is a very archaic name for the Polish. "Polski" much better reflects the common way to refer to the language.


kvch (approved):

asciimoo (approved):
Thanks!

Issue #1011, opened 2017-08-27
when a request is made using the POST method, don't set the title of the HTML page.


dalf:
Right now, the query is displayed in the title of the HTML page.

When the GET method is used, the history of the browser will keep the URL anyway.

The POST method allows to hide the query : 
* from the server logs (the explicit rules is not to log queries)
* from the browser history
* something else ?

When the POST method is used, the history of the browser won't keep the URL, but it will keep the different titles, which means the different queries.

It would make sense not to show the query in the HTML page title when the POST method is used.

ghost:
@dalf  Is there a solution to this problem?

dalf:
Technically, it is simple : 
https://github.com/asciimoo/searx/blob/master/searx/templates/oscar/results.html#L11

replace by
```
 {% block title %}{% if method == 'GET' %}{{ q|e }} -{% endif %}{% endblock %}
```

Warning : not tested, may be there is a syntax error, but that's the idea.

PR #1370, opened 2018-08-14
simple theme update

dalf:
this PR implements some items of #1025

- npm package update
- apply #1226
- implement vim help dialog
- display cookies and search URL with preferences
- allow to enable / disable Open Access DOI rewrite
- add a clear text button on the left of the search button
- implement #1011 : the HTML title page is not set when using POST
- remove searx/static/themes/simple/img/loader.gif
- use full width when only there are only images as result


asciimoo (approved):
Onice! Thanks =)

PR #1374, opened 2018-08-18
[WIP] [engine] Duden.de (German dictionary) 

dadosch:


dadosch:
Tests needed

searx/settings.yml
```diff
@@ -713,6 +713,11 @@ engines:
     engine : 1337x
     shortcut : 1337x
     disabled : True
+
```
kvch: My vim shows unnecessary whitespaces here. Could you please remove those?

searx/settings.yml
```diff
@@ -713,6 +713,11 @@ engines:
     engine : 1337x
     shortcut : 1337x
     disabled : True
+    
+  - name : Duden
+    engine : duden
+    shortcut : duden
```
kvch: Why not have a shorter shortcut? "du" is not yet taken.

dadosch: Ok, I can change that.

dadosch:
~~aarg i messed up the squash~~ This is better now. Are tests ok like this?

asciimoo (approved):
Great, thanks.

aliceinwire:
I think dictionaries need to have their own category

asciimoo:
Well, if I use dictionary, I only activate that engine from the search query, like `!dc en-hu apple`, so its' category doesn't really matter for me. What would be the benefit of an own category?

aliceinwire:
That's true using search special command is really useful.
The only problem that I can think of,
is that having dictionary result coming up in a general search doesn't look right. 
Would be nice to have a dictionary category where I can activate the dictionary that I use, and I can look up on more than a dictionary that I choose.

asciimoo:
>  is that having dictionary result coming up in a general search doesn't look right.

True, that's why I deactivate dict engines by default and activate only from queries when I need it.

Btw, category creation is quite easy and fully dynamic, so if you change the category of any engine in settings.yml to any custom one, that engine will be displayed on your instance under the new category.

> and I can look up on more than a dictionary that I choose.

You can activate multiple engines from the search query without grouping them. E.g. query `!wp !go !bi x` will search for `x` in google, bing and wikipedia

Issue #1344, opened 2018-07-11
[Suggestion] More then only one search language


GambaJo:
I have to choose a language for the search. Thst's ok. But often this is not enough.
In other search engines, like duckduckgo or google I can choose the search language too, but they show me also some english results.

I'm a software developer and I often search for some code or solutions. I want to search in my language (german) but often the solutions I'm looking for are in english. In other search engines I have both. In searx I have to switch the language everytime I want the english results too.

asciimoo:
@GambaJo you can specify the language in the search query, too: https://asciimoo.github.io/searx/user/search_syntax.html .

GambaJo:
@asciimoo You did't read my text. Thx for nothing.

Pofilo:
What @GambaJo means is that if he makes a research in a language (let's say french), he will only get results from this language.
As a developer, a lot of results came in english on google/ddg .. So he doesn't get them with searx and is forced to make a new research specifying english.

GambaJo:
@Pofilo This is exactly, what I ment. Thx!

thlautenschlaeger:
same issue for me. would be cool to have saved fast switching

ilf:
I second the request for an option to specify multiple languages. Something like `https://searx.xyz/?q=test&language=de,en` would be ideal.

oyren:
This prevents me from using searx.

ghost:
Same thing, I want but can't use searx because of it.

ghost:
I also frequently just use another search engine, because searx started to filter search results to show only english results (talk about filter bubbles…). This is fine for people who speak only English, but I would like to see German and Polish, too, for instance. However, the user should not have to "unfilter" English; the default case should be that there is *no* language filter at all (with the option to exclude or prioritise languages).

In short: don't filter prematurely; let the user filter, not the server.

miicha:
For me, it would be sufficient, if I can search in english and one other selected language. Therefore I implemented kind of a quick and dirty solution: https://github.com/asciimoo/searx/pull/1419.

It can be tested at: https://finden.tk/

GambaJo:
@miicha Looks good to me (compared to duckduckgo)
![duckduckgo](https://user-images.githubusercontent.com/1378907/47215752-d5cdd800-d3a2-11e8-9ff5-78ab40867efb.jpg)
![searx](https://user-images.githubusercontent.com/1378907/47215753-d5cdd800-d3a2-11e8-89e1-652670bf9d43.jpg)

[...]

PR #1384, opened 2018-08-24
add: Search with multiple languages (solves #1344)

SilverBut:
Now users can use the following syntax to search with multiple
languages:

```
:en :ja conoha test
```

The first language will be the primary language.

However, a related change to the settings interface is NOT implemented
yet, so this feature is only for advanced users now.

Signed-off-by: SilverBut <SilverBut@users.noreply.github.com>

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -213,21 +213,21 @@ def get_search_query_from_webapp(preferences, form):
 
     # get language
     # set specific language if set on request, query or preferences
-    # TODO support search with multible languages
     if len(raw_text_query.languages):
-        query_lang = raw_text_query.languages[-1]
+        query_lang = raw_text_query.languages
     elif 'language' in form:
-        query_lang = form.get('language')
+        query_lang = [ form.get('language') ]
     else:
-        query_lang = preferences.get_value('language')
+        query_lang = [ preferences.get_value('language') ]
 
     # provides backwards compatibility for requests using old language default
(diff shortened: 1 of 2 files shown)


dalf:
It seems okay. 

Just a few notes : 
* the wikidata engine sends two requests for each language. The first request contains only the query text not the request language --> the exactly same request is sent many times. 
* about infoboxes : count(infoboxes) = count(languages)*2
* match_language (in utils.py) modification can be skipped, and replace by 
```python
current_language=match_language(search_query.lang[0],
```
(in webapp.py). It makes more sense to me : the choice of which languages is displayed belongs to webapp.py.

MarcAbonce:
> you may send about 20 requests only for one search engine.  

My only concern with this is that maybe some engines could consider 20 identical simultaneous requests as spam and then block us because of that.  
I think we should limit the number of simultaneous languages to avoid this risk.

ghost:
It sounds inefficient.to submit one query per language per search engine. The yahoo search engine  supports multiple languages in a single query. The findx search engine will support multiple languages in a single query soon.

How about letting the engines/*.py have an optional variable 'multi_language_support=True'. If it exists and true then let the engine deal with the multiple languages, otherwise submit multiple queries.

kvch:
Closing as this implementation might lead to service abuses as described in https://github.com/asciimoo/searx/issues/1030#issuecomment-665633871

PR #1385, opened 2018-08-26
Minor changes

dalf:
* manage.sh : no error when npm is not found and not required
* simple theme : remove unused files


asciimoo (approved):
Great, thanks!

PR #1386, opened 2018-08-29
Fix google engine bug with new search suggestion format

bourrel:
Fix an error raised because of the new google suggestion format on the right side of the page. The html almost look like a classic result but cause an xpath error.

Reducing the scope of the results path in the `search` column (which contains only search results) fixed this issue.


dalf:
Thank you for your PR.
Can you :
* squash your PR in one commit
* fix the case about "```city``` weather"
?

asciimoo:
already fixed

PR #1392, opened 2018-09-10
Creating setting to enforce a minimum safe search.

rolobio:
- Adding `search.minimum_safe_search` to `settings.yml` which allows the instance host to enforce a minimum safe search allowed on their instance.

This still needs support in the UI (hiding the options that would raise an Exception).  I looked around for how to get the value of a setting, but I didn't see anything.  **How do I get the value of a setting from `settings.yml` in a UI template?**

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -240,7 +240,9 @@ def get_search_query_from_webapp(preferences, form):
         query_safesearch = preferences.get_value('safesearch')
 
     # safesearch : second check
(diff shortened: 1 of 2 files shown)


return42:
FWIW: see also discussion at https://github.com/asciimoo/searx/pull/1985#pullrequestreview-428029092

asciimoo:
@rolobio could you please rebase this PR to be able to merge?

PR #1399, opened 2018-09-17
Use a more human-friendly format in searx/data/engines_languages.json

ghost:
Instead of a single line with 500000 characters use nicely formatted JSON.
Sort the lists in engine_languages.py so when updating it is possible to
more easily see the differences (search engines do change the order their
languages are listed in)


kvch (approved):

Pofilo:
Great thanks.

PR #1413, opened 2018-10-08
Support per-query time limit

ghost:
The CGI parameter 'time_limit', if specified, overrides the time_limit
specified globally and per-engine. This is mostly useful for programs that use
the JSON results and have a longer patience than interactive users.


Pofilo:
Outdated with #1640. This issue can be closed.

PR #1420, opened 2018-10-19
clearer results list

miicha:
Nicer results list "inspired" by google. I guess it would be a better user experience for someone coming from google...


Pofilo:
Can you provide a screenshot of the new look ? :)

miicha:
You can also view it live (together with my other PR (https://github.com/asciimoo/searx/pull/1419) at https://finden.tk

grafik

Pofilo (dismissed):

searx/static/themes/oscar/less/logicodev/variables.less
```diff
@@ -1,4 +1,5 @@
 @black: #29314D;
+@dim-blue: #2547bc;
```
Pofilo: That's a small thing, but can you change `2547bc` by `2547BC` ?
It's just to have all colors written in the same way, thanks !

Pofilo:
You also need to modify [this file](https://github.com/asciimoo/searx/blob/master/searx/static/themes/oscar/css/logicodev.min.css) right ?

searx/static/themes/oscar/less/logicodev/variables.less
miicha: Yes sure, no problem

miicha:
No it should be build from the .less files when running grunt in oscar folder.

Pofilo:
It is build from grunt but we used to commit it.
Look into [https://github.com/asciimoo/searx/tree/master/searx/static/themes/oscar/css](https://github.com/asciimoo/searx/tree/master/searx/static/themes/oscar/css).

It has a drawback to have the "same thing" twice on git, but the advantage that after a pull, we don't need to use `make` or anything else.

miicha:
Sorry, I didn't know...
Now I built with grunt and commited the two minified css files.

return42:
@miicha sorry for our late response .. do you have time to update to master and resolve conflicts?  I can understand if not and apologize for the inconvenience due to our late response on your PR.

Issue #1398, opened 2018-09-17
CGI parameter 'lang' vs. 'language'


ghost:
The documentation on asciimoo.github.io/searx/dev/search_api.html says that the parameter is called 'lang', but the code uses 'language' (search.py:get_search_query_from_webapp())

Because other programs already rely on the parameter being called 'language' I suggest fixing the documentation.

(sorry if this is the wrong place to report this error)

PR #1421, opened 2018-10-20
#1398 replaced parameter 'lang' with 'language'

mrozo:
this is bug in the documentation. The code uses 'langugage' parameter


Pofilo:
Great, thanks.

PR #1438, opened 2018-11-20
[add] seznam engine

Venca24:
It adds new search engine for users from the Czech Republic (and therefore it is disabled by default). Seznam.cz is the biggest competitor to Google in the Czech Republic.


Pofilo (approved):

Pofilo:
Great, thanks.

PR #1252, opened 2018-03-27
[mod] Refactor engine's search language handling

MarcAbonce:
This PR includes two things:

1. Add languages supported by either all default general engines or 10 engines.
This closes #1170 and #1248.

2. Add match_language function in utils to match any user given language code with a list of engine's supported languages.
This replaces duplicated language matching for engines and also fixes common issues such as guessing a country code and handling aliases (multiple language codes that represent equivalent languages).
Babel already handles most of this problems, actually, so many of the previously hardcoded fixes are not necessary anymore. Exceptional cases (mostly non ISO standard codes) are stored in a language_aliases dict on each engine.

PR #1446, opened 2018-11-26
[fix] Fix queries in Hebrew and Norwegian so they give results in the right language

MarcAbonce:
Right now when you query Google in Hebrew, all results are in the wrong language. E.g. `:he !google iphone`. The same bug can be reproduced in other engines when using Norwegian. This happens because some languages have more than one possible language code.

This was a mistake on my part. This was supposedly fixed in #1252, but I forgot to pass a parameter in some engines.


kvch:
Sorry for the delay. I tested your branch and I get the following error message when I use the example search query you mentioned.
```
ERROR:searx.search:engine google : exception : global name 'language_aliases' is not defined
  File "/home/n/p/searx/searx/search.py", line 104, in search_one_request_safe
    search_results = search_one_request(engine, query, request_params)
  File "/home/n/p/searx/searx/search.py", line 73, in search_one_request
    engine.request(query, request_params)
  File "/home/n/p/searx/searx/engines/google.py", line 169, in request
    language = match_language(params['language'], supported_languages, language_aliases)
NameError: global name 'language_aliases' is not defined
```

MarcAbonce:
Fixed. It should be working now.

kvch (approved):

kvch:
Thank you!

PR #1457, opened 2018-12-05
[Scripts] Force upgrade requirements.

hobbestigrou:
Force the upgrade of requirements with pip. At the moment, if the version
in the requirement file is changed, there will be no effect, because the
dependencies is already present.


bouckdarko (approved):

asciimoo (approved):
Thanks!

PR #1458, opened 2018-12-05
[Scripts] Add a function check virtualenv.

hobbestigrou:
Add a function to check the command it's running inside a virtualenv.
It's better to use a virtualenv and the update functions doesn't use
sudo.

### DIFF
--- a/manage.sh
+++ b/manage.sh
@@ -15,7 +15,23 @@ ACTION="$1"
 # Python
 #
 
+check_virtualenv() {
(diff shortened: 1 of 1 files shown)


return42:
@hobbestigrou / apologize our late response .. in the meantime we have a Makefile workflow which covers the [python environment](https://asciimoo.github.io/searx/dev/makefile.html#python-environment) .. anyway, thanks a lot for your helping hands.

PR #1464, opened 2018-12-14
restore startpage search results

### LINKED ISSUE #1454: Seems that startpage and ixquick enignes are not working anymore on Searx
Hello and thanks for the works.

I tried on my searx and on other instance like searx.me...  but It's not working.

How could we fix that?

Thanks.

miicha:
startpage works again.
resolves #1454


kvch (approved):
Thank you!

PR #1419, opened 2018-10-19
add additional search language

miicha:
Up to now it is a working, but kind of quick and dirty implementation of an additional search language. Drawbacks are: 
- additional querys
- only one additional language
- language not user configurable

PR #1465, opened 2018-12-14
Feature startpage pagination

miicha:
enable paginated searches with startpage
this PR unfortunately interferes with my dual language search approach (#1419): only every second page contains results of booth languages


miicha:
It can be tested together with my other PRs on https://finden.tk
This PR also includes the bugfix for startpage (#1464 )

searx/results.py
```diff
@@ -133,6 +133,8 @@ def __init__(self):
         self.answers = set()
         self.corrections = set()
         self._number_of_results = []
+        self._qid = []
```
kvch: Please do not add engine specific variables to general, more abstract classes. Why is this needed? If you need to query some random value on startup in order to perform a search, you should take a look at the engine Soundcloud: https://github.com/asciimoo/searx/blob/master/searx/engines/soundcloud.py#L67
If not, please share the idea behind this architecture.

miicha: The Problem with startpage pagination is the following:
- to get the first result page a normal query without any state parameters is sufficient
- to protect following result pages, startpate implemented some checks:
  - the first result page contains two additional parameters (important) parameters
  - these parameters have to be passed when trying to access following result pages
- therefore my implementation uses does the following:
  - extract these parameters from the first result page
  - insert the parameters into the renderd html page as hidden form fields
 - reads the parameters from the submitted form for subsequent page requests
- basically I was searching for a possibility to temporarly save startpage specific data on a per user and request basis

Most probably there is a better solution or implementation to that problem. I only had a quick look to the soundcloud engine, but it seems that the guest_client_id is some "semi constant" value and not changing for every next page call?

kvch: Yes. You are right, this is teh only solution to persist the data temporarily. However, I am still not happy with adding engine specific attributes to a common resource among those.

It would be nice to "generalize" the functionality, so all engines can add their own attributes. What do you think about maybe adding a new attribute to the `ResultContainer` named `engine_attributes` which is a dict containing the key value pairs by engines? Then it would be persisted in the HTML as you have implemented it.

miicha: That sounds nice.
Once I find some time, I will update the PR

kvch: Thank you.

PR #1469, opened 2018-12-19
Fix bing image thumbnail path

bourrel:
Avoid following error by updating thumbnail xpath for bing_images engine.

```
  File "/usr/local/searx/searx/search.py", line 105, in search_one_request_safe
    search_results = search_one_request(engine, query, request_params)
  File "/usr/local/searx/searx/search.py", line 88, in search_one_request
    return engine.response(response)
  File "/usr/local/searx/searx/engines/bing_images.py", line 92, in response
    thumb_json_data = loads(_quote_keys_regex.sub(r'\1"\2": \3', link.attrib.get('mad')))
TypeError: expected string or buffer
ERROR:searx.search:engine bing images : exception : expected string or buffer
```

### DIFF
--- a/searx/engines/bing_images.py
+++ b/searx/engines/bing_images.py
@@ -88,9 +88,7 @@ def response(resp):
 
         url = json_data.get('purl')
         img_src = json_data.get('murl')
-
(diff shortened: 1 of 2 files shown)


kvch (approved):
Tested locally and it works.

PR #1477, opened 2019-01-06
Release v0.15.0

### LINKED ISSUE #1442: New release?
Hi,

I'm the Debian maintainer of searx and we are still shipping version 0.14.0 of searx because there has not been a more recent release since February.

Unfortunately, we see a number of services that do not work anymore with 0.14.0 but which are fixed in git. Would it be possible to make a release soon?

The last commit to the repository was back in August. Is the project still alive? Debian plans to release its next stable version soon and we should not include searx if it's not maintained anymore.

Thanks!
cheers, josch

kvch:
Closes #1442


asciimoo (approved):
Awesome, thanks!

PR #1479, opened 2019-01-07
fix: Remove english shell dependency in manage.sh script

milouse:
The current `manage.sh` script checks its internal methods using:

```
[ $(command -V "$ACTION" | grep ' function$')" = "" ]
```

A problem occurs when your shell locale is not english. Mine is french for example and `command -V update_packages` will return:

```
update_packages est une fonction
update_packages ()
{
    pip install --upgrade pip;
    pip install --upgrade setuptools;
    pip install -r "$BASE_DIR/requirements.txt"
}
```

`fonction` being different than `function`, the script does not work at all.

This pull requests replace the `command` by a `sed` one, which expect than the internal method will all use the same syntax. That way, method check is now language independant. However, the sed check is very naive for now and would need improvement the day more methods (or "private" methods) will be added to the script.


milouse:
Rebased on master

return42:
Hi @milouse, please apologize  the late response .. thanks for pointing out this bug! A more simpler solution might be:

    LANG=C command -V ....

what do you think?

milouse:
Effectivelly, it can be a more simple solution. I discover only recently the locale switch trick, that's why I didn't choose it initially. Feel free to use it instead!

return42 (approved):
Thanks! .. and sorry for the delay :)

PR #1504, opened 2019-01-31
[WIP] Add pipenv packaging

Serkan-devel:
Pipenv is a combination of pip and virtualenv, which should make development and deployment of python app easier.

https://pipenv.readthedocs.io/


return42:
With Makefile (added PR #1756) we do no longer need to build up the  [Python environment](https://asciimoo.github.io/searx/dev/makefile.html#python-environment) manually.  Over this I can't see any benefit from pipenv and close this issue.

PR #1515, opened 2019-02-15
[engine] restore btdigg as btdig.com

volth:
It seems that searx has no working DHT search engine at the moment


kvch:
Could you please add the engine to `settings.yml` and disable it by default?

volth:
In ```settings.yml``` I just undid https://github.com/asciimoo/searx/commit/b72aec0a9b2b548d7f6a8ddecedd58f5392b8372

PR #1528, opened 2019-03-05
[fix] Escape perecent signs in translation file (fixes #1523)

### LINKED ISSUE #1523: Language incompatibility
Hi! there is something wrong with Chinese.
When i set the Chinese as the interface language,  the preferences page return Internal Server Error.

MarcAbonce:
I'm sending the PR here because this causes a bug (#1523) that makes the preferences page unusable in Chinese.
I also tried to do the same change directly in Transifex but for some reason I wasn't able to.

Apparently the bug is due to the url in the translated string which contains percent signs, so when the template is rendered it thinks that the string needs to be formated.


jsthon:
nice work!

dalf:
Thank you !

PR #1542, opened 2019-03-26
archive.is: fix search_url

volth:


dalf:
If I search for ```bbc.co.uk``` (a DNS name), then
```/search/?q=bbc.co.uk``` is redirected to ```/bbc.co.uk``` (so no need to fetch the first URL ```/search....```)

If I search for ```bbc``` provides results, but the JSON expressions in settings.yml are not setup correctly.

volth:
> (so no need to fetch the first URL /search....)

Unless your search contains  ```#``` or ```?``` or another symbol which requires escaping.

> If I search for bbc provides results, but the JSON expressions in settings.yml are not setup correctly.

Yes, it would be nice to fix the parser to handle that page as well

dalf:
Sorry for the delay.
Now, archive.is uses Google Custom Search.

PR #1544, opened 2019-03-27
[fix] get YouTube results

MarcAbonce:
Results are not appearing in the html document anymore, instead they are found inside an object embedded in a script.

### DIFF
--- a/searx/engines/youtube_noapi.py
+++ b/searx/engines/youtube_noapi.py
@@ -8,7 +8,8 @@
 # @stable      no
 # @parse       url, title, content, publishedDate, thumbnail, embedded
 
-from lxml import html
+from functools import reduce
+from json import loads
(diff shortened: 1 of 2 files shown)


dalf:
Not sure if it is my setup, but I have two requests sent to youtube for one searx request : 
```
DEBUG:urllib3.connectionpool:https://www.youtube.com:443 "GET /results?search_query=linux&page=1 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:https://www.youtube.com:443 "GET /results?search_query=linux&page=1 HTTP/1.1" 200 None
```

MarcAbonce:
You're right. I just created #1545 as a separate PR because the issue is not really about the youtube engine, but rather with the way bangs are parsed.

dalf (approved):
Thank you !

PR #1545, opened 2019-03-29
[fix] Stop engine from sending duplicated requests

MarcAbonce:
If you search `!youtube example` two requests are being sent to YouTube.
This happens every time you use a bang to call an engine that supports more than one category.

That was originally added in #1136 in order to support Findx, but now that engine is gone and as far as I know no other engine requires this.

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -113,19 +113,16 @@ class RawTextQuery(object):
                     parse_next = True
                     engine_name = engine_shortcuts[prefix]
                     if engine_name in engines:
-                        for engine_category in engines[engine_name].categories:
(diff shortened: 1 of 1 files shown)


dalf (approved):
This PR, the previous behavior makes each engine deal with the 'none' category. 
It is the case :
- piratebay engine ```search_types.get(params['category'], '0')```
- swisscows (doesn't work) : ```if params['category'] == 'images'```
- faroo : ```categorie = search_category.get(params['category'], 'web')```
- searx_engine: ```'category': params['category'],```
- qwant : same code but one engine per category declared in settings.yml

PR #1562, opened 2019-04-12
Deactivate some engines : subtitleseeker, seedpeer, swisscows

dalf:
http://www.subtitleseeker.com and http://www.seedpeer.eu don't exist anymore.
https://swisscows.ch/ has changed : the engine needs to be updated


dalf:
About swisscows some WIP : https://gist.github.com/dalf/96f62e7c0ff5aac39f11fd50882b4ec9
I prefer to clearly deactivate the engine for now since it doesn't work.

IMHO ```disable: True``` is not enough (for example ddg definitions engine is disabled but it is working)

kvch:
Why not delete the engines which does not exist anymore of deactivating them? If they become live again, we can restore the code from git. But this way, it only pollutes our configuration file with dead engines.

PR #1566, opened 2019-04-14
(#1496 - partly) Fix google_images search

Yetangitu:
Fixes Google Images search in two steps:

 - remove (deprecated?) `asearch` parameter - Google returns `400` error when it is used
 - Google returns JSON embedded in HTML, engine expected pure JSON

PR #1567, opened 2019-04-14
Add image format and source information to display - needs changes to…

Yetangitu:
… engines to actually display something. Displays result.source (website from which the image was taken) and result.format (image type and size). Result is styled with result-format and result-source classes. See PR #1566 for an example of an engine which has the necessary changes.


Yetangitu:
source data bottom-left under image, format data bottom-right

image

Yetangitu:
I can add the changes from commit c14ef68e (Tags are shown as text in the template so strip them.) to this PR to make sure those tags *are* stripped while a better solution (i.e. not adding them in the first place) is being worked on. Once such a solution is ready the change can be undone. Yes?

dalf:
Yes I agree, the commit c14ef68 should be mobed to this PR.

In this PR, since some .less files are modified, the oscar must be recompiled.

See https://asciimoo.github.io/searx/dev/quickstart.html#how-to-build-styles

dalf:
I'm not sure to understand why, but the contents of the css files are replaced by 
```
[object Object]
```

To compile the css, the steps should be : 
```sh
./manage.sh npm_packages
./manage.sh grunt_build
```

Yetangitu:
No such css-related problems here on my local branch, using `node v8.11.2` with `npm 5.8.0` and the latest crop of whatever `./manage.sh npm_packages` dragged in while producing loads of deprecation and security warnings which npm users are more or less taught to ignore...

dalf:
I'm talking about this file for example : https://github.com/asciimoo/searx/blob/0c5c4ca50900f44adce09b48779253d55857a88a/searx/static/css/bootstrap.min.css

Yetangitu:
OK, that one... yes, that contains `[object Object]` here as well, I just checked the theme-related css files which look like the way they should. 

It seems to work when you run `./manage.sh styles` before running the `grunt_build` task - at least it works here:

```
$ ls -l searx/static/css/bootstrap.min.css 
-rw-r--r-- 1 frank src 110477 May 28 15:13 searx/static/css/bootstrap.min.css
```

dalf:
But I would expect to have these files in the PR :  
* searx/static/themes/oscar/css/logicodev.min.css
* searx/static/themes/oscar/css/logicodev-dark.min.css

Yetangitu:
Added

PR #1568, opened 2019-04-15
bing_images: use data from embedded JSON to improve results (e.g. real page title)…

Yetangitu:
…, add image format and source info (see PR #1567), improve paging logic (it now works)


Yetangitu:
Latest build failure is caused by outdated unit test (which does not represent actual content returned by Bing), I'll add an updated version to the PR

dalf:
Using this PR, I have different result counts per page : 34, 27, 28 (paging over ```Paris``` query, first page 34 results, and so on).

I've tried without success this URL (same behavior) :
```python
search_string = 'images/async?{query}&first={first}&count={count}&scenario=ImageBasicHover&datsrc=N_I&layout=RowBased&mmasync=1&SFX=4'
```

I've seen once some requests to an API which returns an JSON output (but not anymore) : 
```
https://www.bing.com/images/api/custom/search?{query}&preserveIdOrder=1&count={count}&offset={first}&skey=5WFJxUWFcRZ8T3vLalCpMXVEhNCLEk-ADbvUNSIIlBo&safeSearch=Moderate&mkt=ja-jp&setLang=fr-fr&IID=idpfs&SFX=1
```

The next problem with this URL is to get ```skey``` (I removed to parameter IG : it doesn't seems to create a problem, the query can change).

Same requests about the commits :
* can you squash the commits into one ?
* can you rename the ```format``` variable into ```img_format``` (and for ```source```) ?

Yetangitu:
The API queries require a key and are limited to 3000/month, 3/second for the free tier, above that payment is required. Both the key requirement (which goes against the objective of increasing search privacy) as well as the query limit make it unusable in practice.

https://azure.microsoft.com/en-us/pricing/details/cognitive-services/search-api/

The strange paging behaviour seems to be an oddity of the way Bing seems to return results, the first search (with `count=30&first=1`) returns *35* hits, subsequent pages return 30 hits with `count=30` The engine uses `count=28` as that value is used by Microsoft for non-JS browsers using the `IBASEP` form - might as well hide in the masses...

```
$ curl -s -b '_EDGE_S=mkt=sv-se&ui=en-us&F=1; SRCHHPGUSR=ADLT=DEMOTE' --user-agent 'Mozilla/5.0 (X11; Linux i686; rv:60.0) Gecko/20100101 Firefox/60.0' 'https://www.bing.com/images/search?q=irus&c^Cnt=30&first=91&FORM=IBASEP'|xidel - --extract '//div[@class="imgpt"]/a/@m'|grep shkey|wc
frank@sloth:~$ for n in 1 31 61 91; do (curl -s -b '_EDGE_S=mkt=sv-se&ui=en-us&F=1; SRCHHPGUSR=ADLT=OFF' --user-agent 'Mozilla/5.0 (X11; Linux i686; rv:60.0) Gecko/20100101 Firefox/60.0' 'https://www.bing.com/images/search?q=paris&count=30&first='$n'&FORM=IBASEP'|xidel - --extract '//div[@class="imgpt"]/a/@m'|grep shkey|wc);done
**** Processing: stdin:/// ****
     35     241   14840
**** Processing: stdin:/// ****
     30     187   12648
**** Processing: stdin:/// ****
     30     150   12273
**** Processing: stdin:/// ****
     30     212   12936
```

The number of results displayed in the browser does not seem to equate the number returned by the engine as can be seen by instrumenting it:

```
...8<...
    print(len(results))
    return results
...8<...

First page: 35 - 34 shown in browser
2nd page: 28 - 28 shown in browser
3d page: 28 - 27 shown in browser
4th page: 28 - 27 shown in browser
5th page: 28 - 28 shown in browser
etc.
```

Something in the way between bing_images.py and the browser is eating those missing images. The previous iteration of the bing_images engine made mention of problems with paging, in that case it was caused by the use of wrong parameters (`offset` instead of `first`, may have worked once upon a time but certainly does not work now, missing `FORM=IBASEP` without which paging doesn't work at all)


I'll change `format`to `img_format`. The `source` parameter indicates the site from which the images were gathered by the scraped site, I could rename it to `img_source` but the question is whether that name is more applicable than plain `source` - not to mention the confusion it would cause with the `img_src` parameter.

dalf:
About the API : using bing.com/images I had some requests to https://www.bing.com/images/api 
Not anymore. Perhaps Bing has done some A/B testing ?

About the changing number of result : I think Bing (or at least the engine) gives twice the same results in some cases. Searx merges with the results [with the same URL and template](https://github.com/asciimoo/searx/blob/master/searx/results.py#L215) : 
* https://github.com/asciimoo/searx/blob/master/searx/results.py#L233
* looking at the JSON output of ```!bing_image paris``` : some results have two positions, which is the result of a merge.

About ```source``` : yes good point.

dalf:
After a long time, the image eater is here : 
https://github.com/asciimoo/searx/blob/554a21e1d07f3b434b5097b4e3d49e1403be7527/searx/results.py#L216

Results are merged if the ```url``` is the same without taking care of ```img_src```, so two images from the same web page are merged.

Thank you for your PR

PR #1571, opened 2019-04-16
flickr_noapi: use complete JSON data block, add 'content', 'img_format', 'source', etc.

Yetangitu:
fetch complete JSON data block, use legend to extract images. Unquote urlencoded strings. Add image description as 'content'. Add 'img_format' and 'source' data (needs PR #1567 to enable this data to be displayed). Show images which lack ownerid instead of discarding them.


Yetangitu:
This is what the result will look like (PR #1567 applied)

image

dalf:
Thank you !
Sorry for the long delay.

For information, in modelExport variable, realname, username, title, description fields are actually encoded using the ```escape``` javascript function.
* "%u5409" becomes "吉"
* "%20" becomes " ", "%F3" becomes "ó"

For reference:
* https://www.ecma-international.org/ecma-262/6.0/#sec-unescape-string
* https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/unescape

PR #1574, opened 2019-04-18
devianart: use RSS API, etc.

Yetangitu:
use RSS API, add 'content', 'img_format' and 'source' (these two need PR #1567 to be displayed), 'author' fields, use biggest thumbnail image available

### DIFF
--- a/searx/engines/deviantart.py
+++ b/searx/engines/deviantart.py
@@ -8,8 +8,6 @@
  @results     RSS
  @stable      yes
  @parse       url, title, thumbnail_src, img_src
-
- @todo        rewrite to api
 """
 
 from lxml import etree
@@ -60,7 +58,6 @@ def response(resp):
     rss = etree.fromstring(resp.content)
     ns = rss.nsmap
 
-    # parse results
     for item in rss.xpath('./channel/item'):
 
         # find largest thumbnail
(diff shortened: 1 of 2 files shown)


Yetangitu:
BTW, pep8 might choke on the unit test which contains long lines. May I suggest turning off that check, at least for unit tests? As it is now it just creates extra work for something which many already see as extra work, i.e. extra extra work. In this case the unit test contains long RSS lines as that is what the Deviantart backend returns and as such is what the code should be tested with.

```
$ pip install pycodestyle
$ pycodestyle ...

  '\n\n'
/srv/searx/tests/unit/engines/test_deviantart.py:38:121: E501 line too long (181 > 120 characters)
/srv/searx/tests/unit/engines/test_deviantart.py:49:121: E501 line too long (138 > 120 characters)
/srv/searx/tests/unit/engines/test_deviantart.py:50:121: E501 line too long (125 > 120 characters)
```

Yetangitu:
This is what the result should look like (with PR #1567 applied)

image

kvch:
@Yetangitu To solve the issue of too long lines put `# noqa` to the end of the line. This is what we do in case of tests with long lines.

dalf:
There are encoding problem with the DeviantArt API, see : 
https://github.com/DeviantArt/DeviantArt-API/issues/157

return42:
@Yetangitu  @dalf do you know the state of this PR .. should we merge? If so, @Yetangitu would you please update to current branch master and solve conflicts? / Thanks a lot!

Yetangitu:
I'll update if it is to be merged, using the local version without problems here.

return42:
> I'll update if it is to be merged, using the local version without problems here.

There are conflicts which needs to be solved before one can test this PR.

dalf:
See https://github.com/searx/searx/pull/2286

PR #1579, opened 2019-04-29
bing_videos: fix results, use JSON where possible

Yetangitu:
Fix results, use JSON data block for most data, compose 'content' using all available data, use correct 'url' (direct to source instead of redirect through bing)


Yetangitu:
This is what the result should look like, more or less

image

dalf:
The results which are not videos are skipped, not a big deal (channels and playlists).

Thank you !

PR #1588, opened 2019-05-16
fix: dev: nose package

rachmadaniHaryono:
based on this https://github.com/nose-devs/nose2/blob/master/setup.py nose coverage extra use `coverage_plugin` instead of `coverage-plugin`


kvch (approved):

kvch:
Thank you.

Issue #1596, opened 2019-05-22
Google results not working everytime


unixfox:
I just experienced a recent bug where Google sometimes is trying to load his new UI even if JavaScript is not activated. This make Searx not finding any results because it can't parse the new UI:
![](https://i.imgur.com/7smUpFT.png)
Here is the saved HTML page generated by Google when it tries reply with his new UI:
![](https://i.imgur.com/kC5zKr6.png)
And the old UI that Searx can parse without any issue:
![](https://i.imgur.com/n6K9sk9.png)

To resolve this bug I had the idea to force the user agent to Internet Explorer 12 by adding:
````python
params['headers']['user-agent'] = "Mozilla / 5.0(MSIE 12.0; Trident / 7.0; rv: 11.0) like Gecko"
````
I tried that trick and it worked everytime because Google know by default that it can't load his new UI on IE.

**Comment by rachmadaniHaryono** [comment]
is there a way to check if they use a new one? can you share the html file?

**Comment by unixfox** [comment]
I had to write the HTML code into a file with python like this:
````python
file = open('google.html', 'w')
file.write(resp.text)
file.close()
````
To check if Searx received the new one or the older one.

Here is the .zip of the new UI and old UI HTML files. Please disable Javascript on your browser before because Google automatically redirect to a page.
[GoogleUI.zip](https://github.com/asciimoo/searx/files/3209417/GoogleUI.zip)

**Comment by rachmadaniHaryono** [comment]
```html
 <div class="ZINbbc xpd O9g5cc uUPGi">
  <div>
   <div class="jfp3ef">
    <a href="/url?q=https://www.eonline.com/&amp;sa=U&amp;ved=...&amp;usg=...">
     <div class="BNeawe vvjwJb AP7Wnd"> E! News </div>
     <div class="BNeawe UPmit AP7Wnd"> https://www.eonline.com </div>
    </a>
   </div>
   <!-- ... -->
  </div>
 </div>
```
it seem the content is actually still exist on new google ui, but the class name is obfuscated

i will try if it is possible to parse starting from the url `/url?q=https://www.eonline.com/&amp;sa=U&amp;ved=...&amp;usg=...` instead of class name

immanuelfodor:
Although I tried to rebuild the docker image from latest master, I can't get any results from Google, it just doesn't work, the blue warning is displayed. Had to turn on alternative engines to get back search results.

unixfox:
My PR isn't merged yet you will need to apply my PR as a patch.

[...]

PR #1597, opened 2019-05-22
[fix] Force Google old UI

unixfox:
More details about this PR here: #1596.

In summary: Google sometimes tries to load his new UI that Searx can't parse so by defining the user agent of Internet Explorer 12 it will by default respond with the old UI because it knows that IE doesn't support its new UI.


immanuelfodor (changes requested):
I tested your PR by these commands and it did not solve the original problem, still getting the message "Sorry! we didn't find any results. Please use another query or search in more categories." :S 

```bash
git clone https://github.com/asciimoo/searx.git ~/docker/searx

# enable checkout to "hidden" PR branch
cat ~/docker/searx/.git/config
grep "origin/pr" ~/docker/searx/.git/config
sed -i -E 's#(asciimoo\/searx\.git)#\1\n        fetch = +refs/pull/*/head:refs/remotes/origin/pr/*#' ~/docker/searx/.git/config
cat ~/docker/searx/.git/config

# checkout to https://github.com/asciimoo/searx/pull/1597
cd ~/docker/searx
git fetch origin
git checkout pr/1597
git config user.email randomuser@gmail.com
git config user.name randomuser
git merge master
git status

# start testing
docker build -t searx-f ~/docker/searx/Dockerfile ~/docker/searx/
docker run -d --name searx -p 8888:8888 -e IMAGE_PROXY=False -e BASE_URL=https://domain.tld -e TINI_SUBREAPER=True searx
```

In the docker logs, there are only messages from werkzeug, if you have any idea how to debug this with verbose information produced, I can try.

immanuelfodor:
I tried to debug with these steps:

```bash
docker exec -it searx sh
~ $ vi searx/engines/google.py
```

Then add to line ~213:

```python
    file = open('google.html', 'w')          
    file.write(resp.text.encode('utf-8'))                               
    print(resp.text.encode('utf-8'))                           
    file.close()
```

```bash
docker restart searx
docker logs -f searx
```

Made a search and the classes are still obfuscated:

```html
<div class="ZINbbc xpd O9g5cc uUPGi"><div><div class="jfp3ef"><a href="/url?q=https://www.linkedin.com/company/asdasdasdasds&amp;sa=U&amp;ved=2ahUKEwjq0frGzLbiAhXHLlAKHSTGBBsQFjAEegQIBxAB&amp;usg=AOvVaw3q4RreQiKUrUXDYy9QFB4G"><div class="BNeawe vvjwJb AP7Wnd">asdasd | LinkedIn</div><div class="BNeawe UPmit AP7Wnd">https://www.linkedin.com › company › asdasdasdasds</div></a></div><div class="NJM3tb"></div><div class="jfp3ef"><div><div class="BNeawe s3v9rd AP7Wnd"><div><div><div class="BNeawe s3v9rd AP7Wnd">Learn about working at asdasd. Join LinkedIn today for free. See who you know at asdasd, leverage your professional network, and get hired.</div></div></div></div></div></div></div></div>
```

unixfox:
Can you check if in your `searx/engines/google.py` file at the 203th line you have the same exact line as my PR: https://github.com/asciimoo/searx/pull/1597/files#diff-ed0043204dab3ab8a98a1e916146b068R203?

immanuelfodor:
Yes, it's the same on the pr/1597 branch.

I also tried out many other user agent strings there starting from IE9 but the responses are always the same.

unixfox:
That's strange... you have the error message `Sorry! we didn't find any results. Please use another query or search in more categories.` at every search? I had before my patch sometimes that error but not at every search.
Where is your instance located (country)?

immanuelfodor:
Yes, each and every search. I recognized this behavior this morning, this is why I came here right after it. Previously, a few tap on the search button solved it, but starting from today I could not get a response anymore. It's in Budapest, Hungary.

searx/engines/google.py
```diff
@@ -199,6 +199,9 @@ def request(query, params):
     params['headers']['Accept-Language'] = language + ',' + language + '-' + country
     params['headers']['Accept'] = 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'
 
+    # Force Internet Explorer 12 user agent to avoid loading the new UI that Searx can't parse
+    params['headers']['user-agent'] = "Mozilla / 5.0(MSIE 12.0; Trident / 7.0; rv: 11.0) like Gecko"
```
immanuelfodor: I think I've found what's the problem here:
- Your user agent string is not formatted well, extra and/or missing spaces, etc.
- The header's name should be written as "User-Agent" to be valid.

I tried with this line and now it works! (IE9 on Win7: https://en.wikipedia.org/wiki/Internet_Explorer_9#User_agent_string)

```python
params['headers']['User-Agent'] = "Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
```

Please update the PR and I'll mark it as done :)

immanuelfodor (changes requested):

searx/engines/google.py
unixfox: Oh! Nice finding! I used the user agent that a website gave me in the first google results and it worked for me that's why I didn't bother about the formatting.

immanuelfodor (approved):
Thanks for the edit, I approve the changes although I don't have write access to the repo, so it just means it finally works for me :)

@kvch If you would be so kind to have a look on this PR and maybe release a minor version, it fixes the broken Google search engine, which is a huge win.

Pofilo (approved):
We don't know how long it will works, but let's enjoy this time !
@kvch, I let you merge if you agree.

PR #1602, opened 2019-05-24
Feature/fix new hmac

### LINKED ISSUE #1600: Searchengine Wikipedia causes Internal Error 500 when searching for placenames
When the searchengine Wikipedia is enabled; I get Internal server errors ("TypeError: encoding without a string argument") when searching for placenames (tested with searchterms 'Amsterdam', 'Singapore', 'Eindhoven' and others).

Disabling Wikipedia completely avoids these errors.

rachmadaniHaryono:
fix #1600


searx/utils.py
```diff
@@ -384,10 +384,14 @@ def load_module(filename, module_dir):
 
 
 def new_hmac(secret_key, url):
+    try:
+        secret_key_bytes = bytes(secret_key, 'utf-8')
+    except TypeError:
```
kvch: Do you mind adding additional check to see if in case of `TypeError` `secret_key` is indeed `bytes` type?

rachmadaniHaryono: added commit for test case and a check for `bytes` type but this will not for bytearray

https://stackoverflow.com/questions/34869889/what-is-the-proper-way-to-determine-if-an-object-is-a-bytes-like-object-in-pytho

PR #1606, opened 2019-05-27
new: test: engine

rachmadaniHaryono:
a simple test, which collect any html (and yaml result) on `tests/html` folder

html filename must have following format `{engine_name}_{date}_{time}.html` as example `google_20190523_060535.html`

this result will not run if any of required import is failed (it will not run on python2.7)

### DIFF
--- a/tests/unit/test_engines.py
+++ b/tests/unit/test_engines.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+
+import os
+import mock
+try:
+    import importlib
+except ImportError:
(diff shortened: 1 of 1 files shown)


rachmadaniHaryono:
i will give up on this pr and maybe create another pr for this

some idea for this is to use vcrpy [^1] and maybe create something similar to pytest-golden [^2] but for nose

[^1]: https://vcrpy.readthedocs.io/
[^2]: https://pypi.org/project/pytest-golden/

PR #1612, opened 2019-06-03
[fix] google play

Venca24:
fixes google play engines and adds thumbnails to their results


kvch (changes requested):
I would rather extend the XPATH engine with thumbnail than to duplicate its code in a new engine.

Venca24:
@kvch Is the test stuck?

unixfox:
probably because you need to rebase your branch with the master branch of searx.

kvch (approved):

PR #1615, opened 2019-06-04
Fix not jumping to results loaded by infinite scroll

Nachtalb:
Infinite scroll adds a `hr` tag to split up the sections loaded by it.

https://github.com/asciimoo/searx/blob/e8319fa2cbd388712a39202d44c1b6a3a520b048/searx/static/plugins/js/infinite_scroll.js#L11

The vim bindings `j` and `k`, which jump to the next and previous result
respectively, search for a **direct** sibling with the class `result`.
With the `hr` between results, a direct sibling cannot be found. To fix
this we remove the restriction of it having to be a direct sibling.


kvch:
I've just tested your fix, but it does not seem to work for me. What I did was checkout your branch, made sure that no cookie was saved, turned on vim-like hotkeys and infinite scroll plugins, searched for a string and started moving down the result list using the key "j". When I reached the last result on the page, it jumped to the first one again without loading more results.

Nachtalb:
Hey @kvch,

The problem you are describing is fixed with #1614. This PR only ensures that it the shortcuts jump to **already loaded** results.

kvch:
Great. Thank you for the clarification.

bouckdarko (dismissed):
Ok (with clarifications) on my server.

Nachtalb:
Rebased onto latest master.

return42 (approved):
Thanks!

PR #1621, opened 2019-06-17
Fix regional settings #844 

### LINKED ISSUE #844: Use browser's regional settings to show homepage and results
By default Searx loads the american version of its homepage, with all the text in english, and also the search results are from pages in english and the interface is in english as well, even if the browser's user agent shows other regional settings.
This, by one side makes the results poorly accurate, for most users expect results in their own language and related to their region (e. g. if I search for some gadget to buy, or some news regarding ome person I surely prefer stores in my country or that at least speak my language, and digital newspapers written in my lenguage and preferably from my region), and by other side, most people just don't understand english well enough for a good Internet browsing or don't feel comfortable being forced to read in a language that isn't their native ones.
So, I suggest to use the browser's user agent and the IP to improve the accuracy and utility of the results (the "default language shouldn't be english but the browser's language declared in the UA, for instance) and also the understanding. Sure the user can change all thi by hand, but I think it's an unnecessary inconvenience; other search providers work as I propose by default; IMHO it's the correct approach.

nfk:
Report locale fix done in [spot](https://gitlab.e.foundation/e/cloud/my-spot) project. This one fixes the website language from browser settings `Accept-Language` header. Then accordingly to that setting the right language is use for search. 

Fix #844


return42:
FYI: merged your PR into my https://github.com/return42/searx/tree/dm-cloud instance [searX@darmarIT](https://darmarit.cloud/searx/)

Works! / Thanks!

return42 (approved):
Thanks! .. for being patience :)

return42:
Side-note: Very strange, I merged it locally and pushed direct to nfk:locale branch .. locally this results 
as expected in three patches .. github's PR made 250+ commit from .. very noisy, very strange .. grumpy tool.

PR #1630, opened 2019-07-02
[fix] Small fixes with bangs in queries

MarcAbonce:
This provides two small fixes related to bangs in queries.

1. Never pass bangs to autocomplete suggestions: 
 For example, typing `!go ka` should not autocomplete to "go karts".

2. Preserve bangs in suggestions and corrections:
 For example, if my first query was `!yahoo stack` and then I click on the "stack overflow" suggestion, then the query should be run with the same bangs as the original query, so in this case `!yahoo stack overflow`.


dalf:
Thank you !

PR #1633, opened 2019-07-06
Fix ./manage.sh robot_tests

dalf:


dalf:
doi_resolvers / default_doi_resolver were missing in the settings_robots.yml file, so the test server was not able to start (crash). Since the output wasn't displayed, it was not obvious why the Selenium couldn't connect to searx.

Issue #1077, opened 2017-11-02
Faster Searx (suggestion)


jugi1:
I would like to see option where users can quickly select several speed options,
so that Searx can search:

- **FAST** (minimal search engines selected for speed)
- **NORMAL** (standard default search engines selected)
- **TOTAL** (maximum all search engines selected - slower speed)

jibe-b:
I really like this idea!

Please ping me if nobody advances on this within three weeks please.

hook321:
@jibe-b

CaffeinatedTech:
Yeah, I just started using searx, self-hosted.  It's taking 4 to 6 seconds of blank screen before the results appear.  I'd like a faster option also.  I assume that it is waiting for all engines to either return a result, or timeout.  I suppose the 'Fast' option could just set much lower timeouts, then on the results page give us a 'Search Harder' type button to either try a more relaxed timeout, or just re-sort the results including the results that came back late.

I haven't looked at the code, but here are two theoretical options:
- Fast search sets short timeouts, then on the results gives us a chance to do the full search
- Fast search just shows results sooner, then tells us when the rest of the results come in with a button to include them. This will no doubt re-order the page, so we don't want it to happen automatically.

Pofilo:
Well, taking 1 or 4 or 6 seconds depends on the machine you use and your internet speed. Having a searx instance on a Raspberry Pi on a small home internet connection doesn't seem to be a good idea for example. 

What is your current configuration ?

CaffeinatedTech:
My Searx instance is running in docker on my FX6300 machine behind 100Mbit down 40Mbit up connection.  The issue is more that I'm in AU, so it takes longer to talk to most of the search engines.

Turning off certain engines obviously makes it go fast, because its not waiting for them.  It'd be nice if I could make the default search be fast, then have it take a longer look if I need it, without going in and changing the selected engines all the time.

dalf:
The speed of one engine may change over time (wikidata seems slower than before, for example).

I've implemented a timeout limit per request. See #1640 for the syntax & parameter. Feedback is welcome.

CaffeinatedTech:
That looks cool @dalf 
A setting in settings.yml to set the default timeout_limit might be handy.  I could set the default to 2 seconds for instance, then if I want to do a more thorough search I can use the URL parameter. "<2000" for example ( I think that is 20 seconds looking at your code. )

dalf:
[...]

PR #1640, opened 2019-07-18
[enh] Add timeout limit per request

dalf:
The new url parameter "timeout_limit" set timeout limit defined in second.
Example ```"timeout_limit=1.5"``` means the timeout limit is 1.5 seconds

In addition, the query can start with <[number] to set the timeout limit.

For number between 0 and 9, the number unit is the second :
Example: ```"<3 searx"``` means the timeout limit is 3 seconds

For number above 9, the number unit is the centisecond:
Example: ```"<150 searx"``` means the timeout is 1.5 seconds

Related to #1077
Updated version of PR #1413 from @isj-privacore


Pofilo:
@dalf is it still in WIP ? Can I read and test it ?

dalf:
@Pofilo : you can read / test it ! Thank you !

searx/query.py
```diff
@@ -69,6 +70,19 @@ def parse_query(self):
                 self.query_parts.append(query_part)
                 continue
 
+            # this force the timeout
+            if query_part[0] == '<':
+                try:
+                    raw_timeout_limit = int(query_part[1:])
+                    if raw_timeout_limit < 10:
```
Pofilo: Maybe adding a comment about seconds and centiseconds could help for the future.

Pofilo:
I don't have enough time these days to deploy a platform for testing, however, I don't have anything to say about the core of the feature. Feel free to add the comments if you want (only if you have time for this). Do I need to test it or your test is enough according to you ?

dalf:
@Pofilo no problem. 

Update:
* bug fix: get_search_query_from_webapp ignored "<[num]" if the URL query parameter "timeout_limit" was not specified
* the real timeout limit is minimum between the one ask by the user and the normal timeout. Purpose: avoid possible deny of service ask for a lot of queries with a long timeout limit.
* ```<[anything else than a number]``` will be processed as part of the search query (similar behavior to other special characters processing ```!```, ```?```, ```:```)

dalf:
I changed the bang ```<[num]``` to something perhaps more intuitive:
* below 100, the unit is the second: ```<30``` = 30 seconds timeout.  
* 100 or above, the unit is the millisecond: ```<850``` = 850 milliseconds timeout.

More over, there is a new optional value in settings.yml: ```outgoing.max_request_timeout```.

engine timeout | timeout_limit query param(*) | settings.yml... max_request_timeout | Actual timeout | Comment
--: | --: | --: | --: | :--
3 | None | None | 3 |  
3 | 2 | None | 2 |  
3 | 5 | None | 3 | timeout_limit is ignored
3 | None | 30 | 3 |  
3 | 2 | 30 | 2 |  
3 | 5 | 30 | 5 | timeout_limit is NOT ignored

(*) either the timeout_limit query parameter or the bang ```<[num]``` (if both are specified, priority for the latter one)

Why this complicated thing:
* for people who don't set max_request_timeout, no surprise, the timeout won't go above the previous values (engine timeout).
* for people who set a max_request_timeout, timeout_limit can use the full range from 0 to max_request_timeout.

[EDIT] explain priority between the timeout_limit query parameter and ```<[num]```

PR #1642, opened 2019-07-19
Speed up start time and decrease CPU load

dalf:
@MarcAbonce > can you check this PR ?
Fix the two major bottlenecks : 
```python
    for lc in lang_list:
        if lang_code == lc.split('-')[0]:
            return lc
```
and (in soundcloud.py):
```
response.text
```
which try to detect the charset encoding.

### DIFF
--- a/searx/engines/soundcloud.py
+++ b/searx/engines/soundcloud.py
@@ -28,8 +28,10 @@ categories = ['music']
 paging = True
 
 # search-url
-url = 'https://api.soundcloud.com/'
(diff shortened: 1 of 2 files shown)


MarcAbonce:
Yup, I can confirm that the match_language function still behaves exactly the same way as before and that start up time is now significantly faster.

The SoundCloud engine is also fixed now.

dalf:
Thank you MarcAbonce

Issue #1575, opened 2019-04-22
replace fromstring(html) by fromstring(html, parser=newparser)


dalf:
It is a **possible** speed improvement : lxml.html.fromstring without a specific parser doesn't benefit from multithread. See :
https://stackoverflow.com/questions/32285453/why-does-multithreading-do-not-speed-up-parsing-html-with-lxml

I confirm the benchmark, but the result may be different with searx since fromhtml function is called at different time. 

So in duckduckgo.py for example :
```python
def response(resp):
    results = []
    doc = fromstring(resp.text)
```

should be replace by 

```python
def response(resp):
    results = []
    parser = HTMLParser()
    doc = fromstring(resp.text, parser=parser)
```

Most probably it would convenient to add utility function doing that in searx.utils module : 
```python
from lxml.html import fromstring, HTMLParser

def htmlfromstring(str, **kwargs):
    parser = HTMLParser()
    return fromstring(resp.text, parser=parser, **kwargs)
```

----

Note : [selectolax](https://github.com/rushter/selectolax) is faster than lxml.

PR #1650, opened 2019-07-25
[fix] wikidata engine: faster processing, remove one HTTP redirection.

dalf:
* Search URL is https://www.wikidata.org/w/index.php?{query}&ns0=1 (with ns0=1 at the end to avoid an HTTP redirection)
* Add call_xpath function: compile once for all xpath.
* Add get_id_cache: retrieve all HTML with an id, avoid the slow to procress dynamic xpath '//div[@id="{propertyid}"]'.replace('{propertyid}')
* Create an etree.HTMLParser() instead of using the global one (see #1575)

### DIFF
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -20,6 +20,7 @@ from searx.utils import match_language
 
 from json import loads
 from lxml.html import fromstring
+from lxml import etree
 
 logger = logger.getChild('wikidata')
 result_count = 1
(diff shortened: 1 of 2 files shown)


dalf:
The current search URL is ```https://www.wikidata.org/w/index.php?{query}&ns0=1```
It returns an HTML result.

This one ```https://www.wikidata.org/w/api.php?action=wbsearchentities&format=json&uselang=en&errorformat=plaintext&errorlang=en&search={query}&language=en&type=item&limit=1&props=&utf8=1&formatversion=2```
is used by the autocomplete feature in the wikidata website.

The output is JSON, and it is possible set the search language.

Not sure if there is a drawback ?

Issue #1532, opened 2019-03-10
Interface language set zh_TW but return zh_Hant_TW


bestpika:
1. Goto `/preferences`
1. Set interface language to 國語
![image](https://user-images.githubusercontent.com/5698157/54087987-d851be00-4393-11e9-927c-ba2080691046.png)
1. Save
1. Goto `/preferences`
![image](https://user-images.githubusercontent.com/5698157/54088008-12bb5b00-4394-11e9-82d4-08a277d4ebfe.png)

searx | 0.15.0

----

This return `zh_Hant_TW` but list only have `zh_TW`

https://github.com/asciimoo/searx/blob/master/searx/webapp.py#L169
https://github.com/asciimoo/searx/blob/master/searx/settings.yml#L787

dalf:
Most probably related to https://github.com/flask-admin/flask-admin/issues/403

dalf:
This issue has been fixed by the PR #1654.

PR #1654, opened 2019-07-28
[fix] Small fixes in Preferences view's text

MarcAbonce:
This fixes the following bugs:

1. With Python 3 the random answerer's examples are being rendered as ``random b'string', random b'int', random b'float', random b'sha256', random b'uuid'``.

2. In the interface language dropdown, the Vietnamese option has the Chinese name of the language inside the parenthesis rather than the English version like all the other options.

3. When the interface language is set to Taiwanese Chinese and you return to the Preferences view, the interface language dropdown shows Arabic as the selected language (reported in #1532).

About the last one: The ``locale = 'zh_Hant_TW'`` line is not needed anyway because Flask Babel already converts zh_TW to zh_Hant_TW by itself.

### DIFF
--- a/searx/answerers/random/answerer.py
+++ b/searx/answerers/random/answerer.py
@@ -70,4 +70,4 @@ def answer(query):
 def self_info():
     return {'name': gettext('Random value generator'),
             'description': gettext('Generate different random values'),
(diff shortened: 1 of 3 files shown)


dalf:
Thank you !

PR #1655, opened 2019-07-28
[mod] add searx.utils.ecma_unescape function

dalf:
fix flickr_noapi engine


dalf:
I need some help to fix this one. Especially on the encoding in the test_utils.py file.

PR #1656, opened 2019-07-29
invidious engine. closes #1372

### LINKED ISSUE #1372: [engine] Invidio.us
Invidious is similar to what [HookTube](https://hooktube.com/) used to be, but doesn't use the YouTube API. It's also open source (AGPLv3).

https://www.invidio.us/

cy8aer:
Fast implementation of invidious. May be tuneble with more parameters...
Can override `base_url` for using your own invidious server.


cy8aer:
nope - changes in master...

Issue #652, opened 2016-08-01
Text Wrap Theme Issue on Mobile


ibrokemypie:
Issue on Oscar-Logicodev theme (not sure about others)  with text not being consistently wrapped and sometimes extending out to the side,  adding a horizontal scroll.

ibrokemypie:
![ ](http://picpaste.com/pics/Screenshot_20160801-155725.1470651783.png)

asciimoo:
@ibrokemypie could you send a screenshot or example query where the problem occurs?

dalf:
https://searx.me/?q=Go+vs+python&category_general=on&time_range=

The layout has an problem with this link :
https://www.reddit.com/r/golang/comments/2aup1g/why_are_people_ditching_python_for_go/

The pretty url isn't cut.

asciimoo:
@dalf thanks!

PR #1657, opened 2019-07-29
invidious engine. closes #1372

### LINKED ISSUE #1372: [engine] Invidio.us

cy8aer:
Just a quick implementation of invidious. Able to override `base_url` for using your own invidious instance.


dalf:
Thank you !!

I see that there are other query parameters: 
```
q: String
page: Int32
sort_by: "relevance", "rating", "upload_date", "view_count"
date: "hour", "today", "week", "month", "year"
duration: "short", "long"
type: "video", "playlist", "channel", "all", (default: video)
features: "hd", "subtitles", "creative_commons", "3d", "live", "purchased", "4k", "360", "location", "hdr" (comma separated: e.g. "&features=hd,subtitles,3d,live")
region: ISO 3166 country code (default: "US")
```

* region support would be great.( not sure, something like ```params['language'].split('-')[1]``` )
* date support, something like ? : 
```python
time_range_dict = {'day': 'today',
                   'week': 'week',
                   'month': 'month',
                   'year': 'year'}
...
    if params['time_range'] in time_range_dict:
        params['url'] += "&date={range}".format(range=time_range_dict[params['time_range']])
```
* support for paging ?

cy8aer:
Yep stay tuned. I did not find real paging but by time is an idea. 

But do not expect to much: try the patch and look at the dates and compare these with the dates shown on the invidious pages. It does not matter which instance you use. The `published` fields are somehow broken on the api side (tried the times with a calculator). I need to bugreport it on the upstream project...

Or I am just stupid and cannot find the problem on my side.

cy8aer:
hm, I am very unhappy with the region support. That needs time...

cy8aer:
Ok, do not expect too much from region: It only works on `params{'language'].split('-')[1]` and it does not get videos with the language you select but the result from  a region (aka geo-ip). Select us in germany and the results may not work on a german invidious server.

cy8aer:
@dalf just a question: on which wiki page did you find the parameter set? ;-)

dalf:
@cy8aer Thanks, I will look at it tomorrow morning.

I found information on this page: https://github.com/omarroth/invidious/wiki/API#get-apiv1search
and https://github.com/asciimoo/searx/blob/master/searx/engines/google.py#L97

cy8aer:
https://invidio.us/watch?v=jjaqrPpdQYc 

I learned: read the whole page...

dalf:
lol :-)

dalf:
It is working, except there is an encoding problem, see https://github.com/omarroth/invidious/issues/652

The encoding of the responses seem mysterious.

cy8aer:
Sorry for the rebasing force push (I pressed this useless merge button). 

I now try to detect the encoding of the title and description but I fail. The strings of `resp` are unicode types, `chardet` fails. Does `resp` itself know anything about it's upstream data encoding (though `resp.text` is `unicode`? The Flask documentation is not very informative about that...

But for me it is one special search which creates Windows-encodings in the strings... Looks similar to  invidious closed bug #652 ('rether')

dalf:
This part of the code is not related to Flask.
```resp``` comes from the [requests](https://pypi.org/project/requests/) package.

In the documentation: https://2.python-requests.org/en/master/user/quickstart/#response-content
 > When you make a request, Requests makes educated guesses about the encoding of the response based on the HTTP headers. The text encoding guessed by Requests is used when you access r.text. You can find out what encoding Requests is using, and change it, using the r.encoding property

But ```resp.encoding``` is ```None``` 

That's why I suggested ```resp.content.decode("utf-8")``` (UTF-8 is the default encoding according to JSON RFC).

However, [there is ```resp.json()``` too](https://2.python-requests.org/en/master/user/quickstart/#json-response-content).

So line 56 can be as simple as ```search_results = resp.json()``` (no need to import ```loads```)

cy8aer:
`resp.json()` does some other stuff with the encoding. My problem is fixed with this!

cy8aer:
there are many engines which do loads (that is why I used it), like wolfram, spotify, vimeo...

Are there coding probs too?

dalf:
@cy8aer looking at the requests [source code](https://github.com/psf/requests/blob/master/requests/models.py#L873), ```resp.json()``` basically does ```loads(resp.text)``` except when the ```resp.encoding``` is None ; in this case it uses [```guess_json_utf```](https://github.com/psf/requests/blob/a4c18cd733f97b5659a29589432d8a39e7a0de87/requests/utils.py#L857).

If ```resp.encoding``` is None for the other engines, then yes it would be better to use ```resp.json()```

PR #1661, opened 2019-07-31
Fix engine spotify

0xhtml:
The spotify API requires authentication with API credentials.


return42:
Hi @finn0 .. thanks for your PR and sorry for the late response .. we (maintainers) are all busy :o 

> The spotify API requires authentication with API credentials.

Does the old API without authentication no longer exists or is there no longer a way to crawl spotify without account?

I rellay like to test your PR .. Is there any chance to test spotify engine without account?

0xhtml:
As you can read here https://developer.spotify.com/documentation/web-api/#authentication all requests to the spotify api require authentication. You can not test the api without credentials.

return42:
> all requests to the spotify api require authentication. You can not test the api without credentials.

Thanks for clarifying .. with this information we do not have any regression (It can only get better ) and I merge PR even without explicit testing the spotify engine.

return42 (approved):
Thanks!

HLFH:
I am getting the error: ERROR:searx.engines:Missing engine config attribute: "spotify.api_client_id"
searx is no longer starting correctly.

unixfox:
Just comment the engine like in the default settings.yml and everything will be fine.

return42:
Spotify needs API credentials / see 

- https://github.com/asciimoo/searx/pull/1661#issuecomment-569436131
- https://github.com/asciimoo/searx/blob/master/searx/settings.yml#L603

HLFH:
You can create a client_id from here: https://developer.spotify.com/dashboard/login

In /etc/searx.conf, adding the related `api_client_id` & `api_client_secret` fixed it.

PR #1671, opened 2019-08-06
[mod] tidy oscar HTML output

dalf:
Oscar theme: an uncompressed result page is about 50kb, with this PR it is 40kb (general category).

A minor HTML error is also fixed.

### DIFF
--- a/searx/templates/oscar/advanced.html
+++ b/searx/templates/oscar/advanced.html
@@ -1,16 +1,17 @@
 <input type="checkbox" name="advanced_search" id="check-advanced" {% if advanced_search %} checked="checked"{% endif %}>
-<label for="check-advanced">
+<label for="check-advanced">{{- "" -}}
     <span class="glyphicon glyphicon-cog"></span>
-    {{ _('Advanced settings') }}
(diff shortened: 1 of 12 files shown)


return42:
@dalf can please solve the conflict with a appropriate patch / thanks!

dalf:
In ```searx/templates/oscar/results.html```, line 13 there is :
```python
{% block meta %}{{"    "}}<link rel="alternate" type="application/rss+xml" title="Searx search: {{ q|e }}" href="{{ search_url() }}&amp;format=rss">{% endblock %}
```

The ```{{"    "}}``` is here to have consistent HTML indentation. 

Same in  searx/templates/oscar/base.html : 
```python
    {% for script in scripts %}
    {{""}}<script src="{{ url_for('static', filename=script) }}"></script>
    {% endfor %}
```

It's matter of trade-off between the jinja2 templates and the HTML outputs.

I can remove these "fixes".

return42 (approved):
Tested localy .. works / thanks!

> I can remove these "fixes".

I really like it.  If it is about those 5 bytes more .. this could be handled in with gz in HTTP

Issue #1674, opened 2019-08-09
backport from my-spot


dalf:
May be these commits are interesting to backport into searx:
* https://gitlab.e.foundation/e/cloud/my-spot/commit/812bac9d9649b5ba53a97a43c15b739692706afa
* https://gitlab.e.foundation/e/cloud/my-spot/commit/a9e2a3a01c19dfcb340c211a6590fe4948301803
* https://gitlab.e.foundation/e/cloud/my-spot/commit/6e676cc0d25c38106e0256e406d9b3e9041f6dfd
* https://gitlab.e.foundation/e/cloud/my-spot/commit/2cc736bdea21aab0c7b1680622df9c1496526411

rachmadaniHaryono:
https://gitlab.e.foundation/e/cloud/my-spot/commit/6e676cc0d25c38106e0256e406d9b3e9041f6dfd

i am interested on that commit will create pr with minimal change later

dalf:
* commit https://gitlab.e.foundation/e/cloud/my-spot/commit/812bac9d9649b5ba53a97a43c15b739692706afa
fixed by https://github.com/searx/searx/commit/678699beaf97184a9059b0c6514b1a51c7e6dfd0
* commit https://gitlab.e.foundation/e/cloud/my-spot/commit/a9e2a3a01c19dfcb340c211a6590fe4948301803 fixed by https://github.com/searx/searx/commit/1022228d950c2a809ed613df1a515d9a6cafda7c

dalf:
* commit https://gitlab.e.foundation/e/cloud/my-spot/commit/6e676cc0d25c38106e0256e406d9b3e9041f6dfd : see https://github.com/searx/searx/pull/1675#issuecomment-621165598
* commit https://gitlab.e.foundation/e/cloud/my-spot/commit/2cc736bdea21aab0c7b1680622df9c1496526411 : see PR #2244 and PR #2246

PR #1675, opened 2019-08-09
flake8

rachmadaniHaryono:
- all errors are ignored. future pr can try to fix it with pep8 error
- because requirements-dev.txt is not sorted, the flake8 package is put on top
- ~only `searx/static/themes` is ignored, searx and example folder should be checked (unlike pep8 check)~
- test is limited to some folder and setup file, also`searx/static/themes` is ignored because generated python file

related #1674

e: update test target


dalf:
It would be better to only include the directories to check, otherwise the exclude list might be long.
Example of directory that doom flake8: node_modules, the virtual env directory.

Note sure if 
```
flake8 --ignore "E117 E211 E231 E302 E305 E402 E501 E722 F401 F632 F811 F821 F841 W291 W391 W503 W504 W605" --exclude "$SEARX_DIR/static/themes" searx tests
```
is okay ?

rachmadaniHaryono:
i just made change as your suggestion

as note here are failed file/folder:
```
searx/__init__.py
searx/answerers/statistics/answerer.py
searx/engines
searx/plugins/__init__.py
searx/plugins/https_rewrite.py
searx/query.py
searx/results.py
searx/search.py
searx/utils.py
searx/webapp.py
setup.py
tests/test_robot.py
tests/unit/engines/test_btdigg.py
tests/unit/engines/test_currency_convert.py
tests/unit/engines/test_duden.py
tests/unit/engines/test_google_images.py
tests/unit/engines/test_seedpeer.py
tests/unit/test_preferences.py
utils/fetch_currencies.py
utils/fetch_firefox_version.py
utils/fetch_languages.py
utils/google_search.py
utils/standalone_searx.py
```
e: as for the long line of the ignored error, it is unavoidable for now and because flake8 is stricter than pep8

return42:
FWIW: pylint is a part of the boilerplate see (don't bookmark) https://return42.github.io/searx/dev/makefile.html#make-pylint

I really prefer to conditioning the [.pylintrc](https://github.com/asciimoo/searx/blob/master/.pylintrc) file so that it becomes a quality gate one day .. and replaces the pep8.

rachmadaniHaryono:
if it is agreed and pylint will replace this pr i'm okay with that and will close this pr 

i actually wait months for response

return42:
> if it is agreed and pylint will replace this pr i'm okay with that and will close this pr

no, it is not agreed .. pylint its just a alternative I do prefer (help me to detect errors where code coverage fails).  Please let this PR open.

> i actually wait months for response

Thats true and I am very sorry about .. please take into mind, that we are all have a payed job .. we invest a lot of time on searx .. I am new here at searx, other do this since years .. we have so many PRs open ... that was the reason why I asked to help on searx .. now I am here and I will do my best .. but we need more much resources anyway.

/ thanks for your patience.

rachmadaniHaryono:
actually i'm glad it is active now, even if i didn't use it as often as before.

just ping me if something need to be done with my pr or issue

return42:
We are gonig to check more and more python modules with pylint / https://github.com/searx/searx/blob/master/Makefile#L216

I close this PR.

PR #1680, opened 2019-08-18
Add --no-cache to tell apk to not cache packages

robbyoconnor:
1. Add `--no-cache` to `apk upgrade` and `apk add` calls. This eliminates the need to delete the package cache. 

2. Some other minor optimizations mostly around metadata where order does not matter, but you want to delay cache invalidation.


asciimoo (approved):
Thanks!

Issue #1666, opened 2019-08-02
Chromium-based browsers (Android) don't detect searx


smnthermes:

dalf:
It requires more information : 
* Chromium-based browsers: which one ? is Chrome or another browser ?
* what do you mean by "detect" ?

smnthermes:
1) I tested in Chrome and Bromite
2) I meant by "detect" being able to set as the default engine: https://coinsh.red/u/Screenshot_20190803-171609_Bromite.png

smnthermes:
Should appear in the "recently visited"

unixfox:
Did you do a search on the Searx instance? According to [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=348360#c53) on the chromium bug tracker you need to do a search on the search engine before being able to add it as a default search engine.

smnthermes:
Yes, I did a search in the searx instance.

asciimoo:
I've tested it with the latest chrome with multiple opensearch setups - even with 1.0 and 1.1 opensearch specifications and it doesn't work. I also tried to research on the internet, but I couldn't find any useful info about mobile chromes' opensearch support. Do you have any information about this topic?
It seems to me that mobile chrome doesn't handle opensearch standards.

bat999:
Well, bugs.chromium.org have told me that there are "old internal bugs mentioned in the code" but they "WontFix".

The OpenSearch method with ANDROID chromium-based browsers doesn't work for searx.be and startpage.com sites.

But it works OK for metager.org and qwant.com sites.
So they must have figured out how to work around those "old internal bugs" in the chromium code.

dalf:
@bat999 can you provide the references to these bugs ?

bat999:
Hi **dalf**
Look in this post...
https://bugs.chromium.org/p/chromium/issues/detail?id=1097707&q=os%3DAndroid

return42:
> I've tested it with the latest chrome with multiple opensearch setups 

@asciimoo I can't speak for android, but chrome(ium) has this limitation ...

> Chromium only adds sites that the user navigated to without a path.

You can find this under the link @bat999 and we have it also named in our 'about' page / added with commit e7788980 (https://github.com/asciimoo/searx/commit/e7788980ad1bb82e9ae084428d05a79b235c17ff#diff-a670502bb265cc5b6631bec680aaabccR70). 

[...]

Issue #768, opened 2016-11-15
POST requests (searches) have stopped working in Chrome


negroblack:
Hello

This is not a problem in Firefox or Safari.

The latest version of Chrome on Mac (Version 54.0.2840.98 (64-bit)) can no longer do POST searches. If you do a search using POST, it just reloads the search index page.

How can I debug this to figure out what's causing it?

Note GET works fine.

Thanks.

asciimoo:
@negroblack chrome cannot send POST on the URL which is set to the default search URL. Is searx the default search engine in your case? If so, try to rewrite your default search URL from `/?q=%s` to `/search?q=%s`

nixoye:
@asciimoo how do we make it by default use GET so users coming to our site on chrome can browse without a problem?

kvch:
@NIXOYE It is already handled in `opensearch.xml`. If the user agent is Chrome, it uses `/search` as the default search url. So `POST` can work at `/`. If searx is added to Chrome as it supposed to be done, it should work without any problems.

thearkadia:
@kvch @asciimoo POST isn't working on chromium and it doesnt change to Get by default

asciimoo:
@arkgit it works for me with the latest chromium. Do you have any custom browser settings/plugins?

unixfox:
Works for me. I cannot replicate anymore.

PR #1681, opened 2019-08-20
[mod] Separate index and search routes

MarcAbonce:
This should make it easier to separately handle search and index requests from a web server or from a reverse proxy. Specifically, the problem right now is that both index and search requests look exactly the same to anything before searx itself, so it's hard to filter or rate-limit search queries without also targeting requests to index.

If a request to index contains a query, a permanent redirect HTTP response is returned. This should give some level of backwards compatibility for users that have set a searx instance in their browser's search bar. I would still recommend that users remove and add again searx in their browsers to avoid making double requests every time.

The search path in the URL is ``/search`` by default, but it can be customized in settings (this featured is inspired by [this comment](https://github.com/asciimoo/searx/issues/1584#issuecomment-493596451)). The search_path setting can even be empty. In that case, queries are run on index, just like they did before this pull request.

### DIFF
--- a/searx/static/plugins/js/infinite_scroll.js
+++ b/searx/static/plugins/js/infinite_scroll.js
@@ -9,7 +9,7 @@ function loadNextPage() {
         $('#pagination').html('<div class="loading-spinner"></div>');
         $.ajax({
             type: "POST",
-            url: './',
+            url: $('#search_form').prop('action'),
             data: formData,
             dataType: 'html',
             success: function(data) {
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -6,9 +6,9 @@
   <Image>{{ urljoin(host, url_for('static', filename='img/favicon.png')) }}</Image>
   <LongName>searx metasearch</LongName>
   {% if opensearch_method == 'get' %}
-    <Url rel="results" type="text/html" method="get" template="{{ host }}?q={searchTerms}"/>
+    <Url rel="results" type="text/html" method="get" template="{{ url_for('search', _external=True) }}?q={searchTerms}"/>
   {% else %}
-    <Url rel="results" type="text/html" method="post" template="{{ host }}">
+    <Url rel="results" type="text/html" method="post" template="{{ url_for('search', _external=True) }}">
         <Param name="q" value="{searchTerms}" />
     </Url>
   {% endif %}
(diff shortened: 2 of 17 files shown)


dalf:
Related to https://github.com/asciimoo/searx/issues/1666#issuecomment-657028715

asciimoo:
> This should make it easier to separately handle search and index requests from a web server or from a reverse proxy. Specifically, the problem right now is that both index and search requests look exactly the same to anything before searx itself, so it's hard to filter or rate-limit search queries without also targeting requests to index.

This workaround is introduced because of chrome/chromium and how they handle GET/POST search requests. What other solution do you suggest to solve both your and chrome's issues?

MarcAbonce:
I just tested my branch with desktop Chromium and I also added my local searx as the browser's default search engine, which indeed defaults to GET, just like before. I tried reproducing #1666 (only with the desktop Chromium though) and #768 with my branch and I couldn't find anything broken, but maybe I wasn't looking at the right place.

Besides the Android issue, am I missing some other thing I should be testing with Chrome?

asciimoo:
>  I couldn't find anything broken, but maybe I wasn't looking at the right place

Can you do POST requests from the UI in this case?

MarcAbonce:
Yeah, I just tried and I can do POST requests with Chromium from either the index page or from the search results page.

asciimoo:
Looks good to me. My only concern is about making the search URL path configurable. It would make search instances incompatible with each other. If anyone really wants to configure a different search URL they can do it in their production webserver.
What do you think?

MarcAbonce:
Makes sense.

I removed the configurable URL path, but now I got a conflict with a5e6423c3926a6c9811f11cd3e2b10fa1bbde012 in the `opensearch.xml` file. Unfortunately, I don't have an Android with me, so I can't test if I broke something in there or not.

unixfox:
I tested your PR on multiple browsers:

- :heavy_check_mark: Google Chrome 86 on Windows: https://user-images.githubusercontent.com/4016501/97990702-40aaaa00-1dd8-11eb-83a3-aaa53afb07ed.png
- :heavy_check_mark: Firefox 82 on Windows: https://user-images.githubusercontent.com/4016501/97990867-78b1ed00-1dd8-11eb-966c-991c1c28148e.png
- :heavy_check_mark: Google Chrome on Android: https://user-images.githubusercontent.com/4016501/97991005-b1ea5d00-1dd8-11eb-9653-70eb88e14ac2.png
- :heavy_check_mark: Safari 13.1 on macOS: https://user-images.githubusercontent.com/4016501/97991418-5a98bc80-1dd9-11eb-95ba-b7581c025b1d.png

asciimoo (approved):
Awesome, thanks a lot @MarcAbonce @unixfox !

Issue #1685, opened 2019-08-25
Fix info box: Data fetch from Wikipedia page isn't showing properly inside the info-box


etanot:
Data fetch from Wikipedia page isn't showing properly inside the info-box.

* **Ambiguous Pages:** For some search results it only show title matched to searched keyword nothing else (try these keywords: [ip](http://search.disroot.org/?q=ip&categories=general&language=en-US), [tcp](http://search.disroot.org/?q=tcp&categories=general&language=en-US) etc).

![ip](https://user-images.githubusercontent.com/28824347/63648783-211fe800-c724-11e9-9073-be69ef7ebee4.png)

see there are two info-boxes one with empty content other with only one line in it.

* **Pronunciation:** If fetched content have pronunciation within it, either show it inside parenthesis or drop it like DDG's [info-box](https://duckduckgo.com/?q=git+wiki&t=ffab&ia=web).

![git](https://user-images.githubusercontent.com/28824347/63648824-b9b66800-c724-11e9-9fdf-4d1c53f48420.png)

see there is an empty parenthesis inside the info-box but, on wiki page it contents pronunciation of git. (searched keyword: [git](http://search.disroot.org/?q=git&categories=general&language=en-US))

return42:
fixed by #1686

PR #1686, opened 2019-08-26
Infobox fixes

MarcAbonce:
Fixes issues in #1685 (or at least provides some workarounds) and also fixes another bug I found with the suggestions box.

1. The first issue happens because Wikipedia returns a disambiguation page. Since disambiguation pages do not return any meaningful information for now, I think it would be better to just remove them from the results.
Alternatively, we could later add a special case where disambiguation pages return a list of possible articles in an infobox or somewhere else.

2. About the second issue, it looks like Wikipedia's API [removes](https://www.mediawiki.org/wiki/Extension:TextExtracts#How_can_I_remove_content_from_a_page_preview/extract?) the IPA pronunciation for some reason. E.g. https://en.wikipedia.org/w/api.php?action=query&format=json&titles=git&prop=extracts|pageimages&explaintext&exintro
So as long as we can't retrieve the pronunciation, it would be better to at least remove the empty parenthesis as well.
There is a small risk that this removes a set of empty parenthesis that are actually meant to be in the article, but I haven't found any case yet. Luckily, this fix even works with the [( ) (album)](https://en.wikipedia.org/wiki/(_)_(album)) article.

3. Finally, there's also a python3 specific bug where the suggestion box is always shown even when it's empty. That's because in python3 map returns an iterator so we don't know whether it's empty or not until we iterate through it.

### DIFF
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -21,7 +21,8 @@ search_url = base_url + u'w/api.php?'\
     'action=query'\
     '&format=json'\
     '&{query}'\
-    '&prop=extracts|pageimages'\
+    '&prop=extracts|pageimages|pageprops'\
+    '&ppprop=disambiguation'\
     '&exintro'\
     '&explaintext'\
     '&pithumbsize=300'\
@@ -87,7 +88,7 @@ def response(resp):
         if int(article_id) > 0:
             break
 
-    if int(article_id) < 0:
+    if int(article_id) < 0 or 'disambiguation' in page.get('pageprops', {}):
         return []
 
     title = page.get('title')
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -100,6 +100,7 @@ def response(resp):
     extract = page.get('extract')
 
     summary = extract_first_paragraph(extract, title, image)
+    summary = summary.replace('() ', '')
 
     # link to wikipedia article
     wikipedia_link = base_url.format(language=url_lang(resp.search_params['language'])) \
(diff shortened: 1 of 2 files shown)


etanot:
@MarcAbonce,
 
> Alternatively, we could later add a special case where disambiguation pages return a list of possible articles in an infobox or somewhere else.

We can show it in **Suggestions** box.

I noticed one thing about info-box which contains only one line of information in it, they also have link to their **Wikipedia** page why not fetch some content from their Wikipedia page and add it to the info-box. Isn't that would be nice? Or, this merge request fixes that also.

MarcAbonce:
The main problem I see right now with adding the disambiguation pages in the suggestions box or anywhere else is that the list can get huge in some cases. For example, [IP would return over 40 possible pages](https://en.wikipedia.org/wiki/IP).

About the one line descriptions, those are returned by Wikidata. So in those cases, the only way to show Wikipedia's first paragraph would be to send an additional request to Wikipedia. But that would likely take too long and the user would get an engine timeout instead.

So, essentially, we could solve those two things, but right now I'm not sure on what would be the best way to do so. For now, I'd prefer to keep the PR as it is and keep improving the infoboxes later.

etanot:
> list can get huge in some cases. 

We can set a limit for no. of entries to be shown to 2 or 3 and link `...` to wikipedia disambiguation page. For example, for `IP`:

* [Informatics Practice](https://en.wikibooks.org/wiki/Informatics_Practices_for_Class_XI_(CBSE))
* [Intellectual Property](https://en.wikipedia.org/wiki/Intellectual_property)
* [...](https://en.wikipedia.org/wiki/IP)

or, something similar.

return42 (approved):
Thanks!

PR #1689, opened 2019-08-27
[fix] Google Images

### LINKED ISSUE #1687: Google Images & DeviantArt don't work anymore
From one day to another, Google Images and DeviantArt stopped to show me anything, even with simple searches.
Screenshot_1
Screenshot_2
They are of course activated in the engines. It has been a few days, with a restart every day (local instance using Docker), without modifying engines nor any other setting. Using searx 0.15.0

MarcAbonce:
Sometimes the JSONs in the response are missing some values. As long as those values are not essential the result should still be returned.

This should partly fix #1687.

### DIFF
--- a/searx/engines/google_images.py
+++ b/searx/engines/google_images.py
@@ -70,11 +70,21 @@ def response(resp):
 
         try:
             metadata = loads(result)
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Awesome, thanks!

PR #1694, opened 2019-09-14
Fix broken Library Gensis Engine

### LINKED ISSUE #1693: Broken Library Gensis Engine
When I searched [linux programming](https://searx.me/?q=%21lg%20linux%20programming&categories=none&language=en-US) keyword with `!lg` prefix, I get this error message 
```
Error! Engines cannot retrieve results.
 library genesis (request exception)
Please, try again later or find another searx instance.
```

I think, it's due to `Library Gensis` is switched to new domain URL (https://libgen.is)<sup>[[1](https://en.wikipedia.org/wiki/Library_Genesis)]</sup> with TLS/SSL support and older URL (http://libgen.io) used in setting.yml<sup>[[2](https://github.com/asciimoo/searx/blob/master/searx/settings.yml#L410)]</sup> file is not reachable and also it doesn't support TLS/SSL. So, we should switch to https://libgen.is URL.

etanot:
Since, Libgen has switched to new domain (i.e. https://libgen.is) and in `settings.yml`, we're still using old domain, so I've replaced http://libgen.io with https://libgen.is and also some `http` in URLs with `https`.

This merge request will fix #1693 issue.


return42:
Merged into my [dm-cloud](https://github.com/return42/searx/tree/dm-cloud) branch and tested with [!lg Van-der-Waals](https://darmarit.cloud/searx/?q=%21lg%20Van-der-Waals) ... works! / thanks a lot!

etanot:
Seems like more people is affected by #1693 issue. Can you review and merge this MR?

/cc @asciimoo @dalf

etanot:
Since this MR isn't merged, yet, so I'm thinking about changing commit message title "fix issue#1693" to something more appropriate.

@return42, Can I force push commits to this branch?

unixfox:
You can do whatever you want in your pull request, for example force push.
I occasionally do it for some PR in the searx repository.

searx/settings.yml
```diff
@@ -407,7 +407,7 @@ engines:
 
   - name : library genesis
     engine : xpath
-    search_url : http://libgen.io/search.php?req={query}
+    search_url : https://libgen.is/search.php?req={query}
```
return42, line 411: With httpS I got exception bellow:

    $ curl https://libgen.is/search.php
    curl: (60) SSL certificate problem: certificate has expired
    
````
...  File "~/searx/local/py3/lib/python3.7/site-packages/requests/adapters.py", line 516, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='libgen.io', port=80): Max retries exceeded with url: /search.php?req=foo (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fb0f43f0320>: Failed to establish a new connection: [Errno -2] Name or service not known'))
````

The only adress that worked for me is http://libgen.is/search.php

return42 (approved):
Successfully tested all commits, except the one issue with http(s)://libgen.is/search.php 

Can you please fix it / Thanks!

searx/settings.yml
etanot: That's expected, their certificate has been expired on 23th december and server isn't configured properly to auto-renew Lets Encrypt certification in every 3 months.

> The only adress that worked for me is http://libgen.is/search.php

In my humble opinion, we shouldn't use non-https version (i.e. http://libgen.is) if https version is available, we should wait until they don't generate a new certificate. @return42 what do you think?

etanot: Is it possible to add two `search_url` addresses? Such that if first one didn't work, fallback to second `search_url` address. If so, we can add both https and non-https URL.

unixfox: It's always better to use HTTPS instead of HTTP even if the certificate is expired/invalid.

return42: > In my humble opinion, we shouldn't use non-https version .. wait until they don't generate a new certificate. @return42 what do you think?

think same, please change to http://libgen.is and we can merge this PR

> Is it possible to add two search_url addresses? 

> It's always better to use HTTPS instead of HTTP even if the certificate is expired/invalid.

Both are new features / if you think any of has its value send PR or add issue for / thanks!

etanot: > please change to http://libgen.is   

This is would be very very bad idea, anyone can clearly see all traffic between Searx instance and Libgen, can perfrom man-in-the-middle attack and all other disadvantages we've due using http instead of https. In my opinion, having a broken engine is much better solution than leaving all connection insecure because in few days admin will/must fix the problem, but once we added non-https URL here, we put every user in risk. If you still insist, I can change that to non-https, but I strongly oppose your of idea. Please think about it once.

PR #1700, opened 2019-09-30
Initialize support for offline engines

kvch:
This PR adds initial support for offline engines to searx core. It introduces changes to parsing and merging results. All engines have a new attribute named `offline`. If an engine is offline, it does not have `page_load_time` and `page_load_count` stat, as it does not make sense.

Offline engines might return results without an URL. So it's not required to have an `url` key in the result anymore. Thus, `parsed_url` is not always present in results. This has the following consequences:
* processors operating on results with `parsed_url` key (e.g. HTTS rewrite) skip such results without the key
* all results without a URL are considered unique, thus are always added to the result container
* existing templates might not be suitable for displaying the result

To mitigate the presentation issue, I added a new result template to show results with arbitrary keys named `key-value.html`. The example below shows results with the keys `filename`, `permission` and `size` from the engine named `ls searx`:

image

It has been tested with a dummy offline engine with a configurable command. The engine still needs a bit more work to support queries, so I am not submitting it in this PR.

The PR also add support for passing engine settings to the `init` functions of engines. I added a new utility function `get_engine_from_settings`. It retrieves the settings from the configuration file of a given eng


asciimoo (approved):
Awesome, thanks.

PR #1716, opened 2019-10-12
Add a PWA plugin

madmath03:
:sparkles: Add a plugin to turn searx into a PWA to close #1713

PR #1720, opened 2019-10-14
Config for instance name and autocomplete

madmath03:
:whale: Config for instance name and autocomplete


kvch (approved):

madmath03:
For info, if #1716 gets accepted, I have another PR based on this one to also update the Web App Manifest with the instance name set here.

This way, the web interface and PWA will display the same name.

asciimoo (approved):
Great, thanks.

Issue #934, opened 2017-06-02
Improve searx image proxy ?


dalf:
Privacy :
- Remove the ```If-Modified-Since``` and ```If-None-Match``` sent headers.
- Remove the ```Last-Modified```, ```Expires```, ```Etag``` received headers.

Ressource usage improvement:
- http://flask.pocoo.org/snippets/118/

dalf:
The snippets doesn't exist anymore, see : https://gist.github.com/fanzeyi/7490019

```
# -*- coding: utf-8 -*-

from flask import Flask
from flask import Response
from flask import stream_with_context

import requests

app = Flask(__name__)

@app.route('/<path:url>')
def home(url):
    req = requests.get(url, stream = True)
    return Response(stream_with_context(req.iter_content(chunk_size=1024)), content_type = req.headers['content-type'])

if __name__ == '__main__':
    app.run()
```

Issue #792, opened 2016-12-27
P2P idea


dalf:
For one user request there are most of the time different engine requests. among other searx instances (in the general category : wikipedia, wikidata, qwant, google, etc...)

### Purpose

For one user request, the purpose is to spread the individual engine requests to different searx instances. Of course, It's required trust in other searx instances from the searx administrator and from the users. 

### How 

Let's call :
* the searx instance receiving the user request, main instance. 
* the other instances, peer instances.

The first step, searx should improve response time measures :
* the main searx should measure the median response time, and the 95th percentile response time (95% of the time the response time is bellow that measure) :
** for each engine of the main instance
** for each (peer instance, engine) couple.

For example (YMMV) : There are slow and fast engines : wikipedia is fast, google is slow (double). So for user request using google and wikipedia, the main instance can send the request for the google engine, and use a peer instance for the wikipedia. The user response time won't change, since the google response time set the upper limit.

So one way to spread the requests is :
* for slow engines : do it in the normal way on the main instance.
* for fast engines : use a peer instance to proxy the request.

The global response time should not change too much.

So when there is user request on the main instance :
* the main instance guesses the global response time without peers, using the maximum the median of each engine of the user request (not sure if the median is the best choice).
* the main instance selects a peer instance for each engine according to the guessed p95 response time of the couple (peer instance, engine). Using the p95 gives the worst case scenario (or nearly).
* the main instance spreads the request to the different peers.
* the response time of each (peer instance, engine) improves the statistics.

### Problems not solved
* How to select a (peer instance, engine) if no request has been sent to it ? Perhaps the main instance should bootstrap by sending one or two requests for each (peer instance, engine). And here, starts the problem of flood.
* How the main instance and the peer instances should communicate ? Using the rss API ? Using something similar to morty with a Hash ?

[EDIT] clarification.

PR #1724, opened 2019-10-16
Proof of concept: Quart

dalf:
asyncio demo / proof of concept:
* drop python 2 support (see https://github.com/dalf/searx/tree/python3 )
* use Quart ( https://pgjones.gitlab.io/quart/index.html ).
* requires Python 3.6, see https://pgjones.gitlab.io/quart/python_versions.html (aiohttp supports Python 3.5.2, but requires more works).
* use httpx ( https://github.com/encode/httpx ). 
* image proxy seems faster (but can be done with Flask, see #934).
* only one thread except to parse html document (lxml).

what is not working: 
- [ ] tests
- [ ] queries are bytes, could be string
- [ ] some engines (minor adjustments to do)
- [ ] simple theme: bug with one jinja2 filter ( https://github.com/asciimoo/searx/blob/master/searx/templates/simple/results.html#L9 ).
- [ ] HTTP2: https://github.com/encode/httpx/issues/381

related to https://github.com/asciimoo/searx/issues/904#issuecomment-305446256
whatever benchmark results are: 
* #792 (or a similar idea) requires most probably persistent connections. asyncio would avoid numerous threads.
* (compare to uwsgi) one process would help to have more accurate statistics (instead of getting the stats for one uwsgi process).
* (compare to uwsgi) one process makes embed searx-checker implementation easier (avoid file / whatever synchronization to avoid multiple checks, one per process).
* if the image proxy is fast enough, morty could be used only when the admin wants to provide a html proxy. Not sure if httpx (or aiohttp) could be as fast as valyala/fasthttp


dalf:
See https://github.com/dalf/searx/tree/quart_aiohttp for a try of quart + aiohttp as a client.
It seems slower that quart + httpx.

## Quart

Notes about quart:
* The version 0.6.15 is the last one to support Python 3.6
*  https://github.com/pgjones/quart/blob/master/src/quart/flask_patch/_patch.py#L11 : 
   > it requires the python Task and Future implementations, which invokes some performance cost.
* https://jinja.palletsprojects.com/en/2.9.x/api/#async-support
  > all template generation code internally runs async generators which means that you will pay a performance penalty even if the non sync methods are used!

Dirty workaround idea about the last point: monkey patch ```quart.templating._render``` to render the template in a ThreadPool using [asyncio.loop.run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)

## httpx
httpx is based on [h11](https://github.com/python-hyper/h11) and [h2](https://github.com/python-hyper/hyper-h2). Both use the bring-your-own-I/O principle.
README.md says: 
> This project should be considered as an "alpha" release. It is substantially API complete, but there are still some areas that need more work.

### HTTP/1.1 and HTTP/2
I agree, after doing some requests:
* HTTP/2 supports seems cause issue here and there (it is possible to disable it)
* HTTP/1 has also issues too to deal with broken servers (requests, pycurl works not httpx on some requests).
* But to send requests to wikipedia, duckduckgo, etc... it is okay (without HTTP/2).

### Proxy and connection reuse

* https://github.com/encode/httpx/blob/1a32cf036a825f6eb35395af5388a3b23180a82e/httpx/dispatch/proxy_http.py#L108
> After we receive the 2XX response from the proxy that our tunnel is open we switch the connection's origin to the original so the tunnel can be re-used.

## aiohttp

aiohttp uses C to provide an server and a client. 
See https://github.com/aio-libs/aiohttp/tree/master/aiohttp
The library declares itself as "5 - Production/Stable".

### HTTP/2
The HTTP client doesn't support HTTP/2 : https://github.com/aio-libs/aiohttp/issues/863#issuecomment-221802223

Anyway it seems as fast as pycurl, see https://github.com/dalf/searx/blob/pycurl-2nd-edition/misc/httpclient_bench.py

### Proxy and connection reuse
* https://github.com/aio-libs/aiohttp/blob/48bf8c356c7c3bac5d520484ede5c35bd1795afb/aiohttp/connector.py#L1008
> Many HTTP proxies has buggy keepalive support.  Let's not reuse connection but close it after processing every response.

## Which path ?

I think a full aiohttp searx version should be tried:
* [aiohttp server](https://aiohttp.readthedocs.io/en/stable/web.html)
* [aiohttp client](https://aiohttp.readthedocs.io/en/stable/client.html)
* [aiohttp-jinja2](https://github.com/aio-libs/aiohttp-jinja2) (referenced by the [documentation](https://docs.aiohttp.org/en/stable/web_advanced.html#template-rendering))
* The CONNECT method seems supported: https://github.com/aio-libs/aiohttp/blob/master/tests/test_proxy_functional.py#L14

nfk:
Hi @dalf, I had tested aiohttp few months ago. I had abandoned the idea because I did want to move apart from upstream.

You can see my work [here](https://gitlab.e.foundation/e/cloud/my-spot/commit/d24d6e3012db55860444cb0dee035e338254ef7a) after removing [python2 support](https://gitlab.e.foundation/e/cloud/my-spot/commit/2cc4bf3eaa5cbab007b56ad4ac287d56fc57f5ae)

Whatever moving to ASGI is a really good idea. Starting with quart is probably a reasonable choice to move the project toward asyncio.

dalf:
See https://github.com/searx/searx/pull/2313

PR #1732, opened 2019-10-28
Add searchbox_text in settings.yml

michaelfm1211:
Adds a searchbox_text field in settings.yml which changes the default placeholder of "Search for ..." to whatever is in that field.

PS: Sorry for the very large diff in settings.yml, i have only added the searchbox_text field in settings.yml but for some reason Visual Studio Code deleted everything then added it back in, probably trying to auto-format the file.


return42:
> PS: Sorry for the very large diff in settings.yml, i have only added the searchbox_text field in settings.yml but for some reason Visual Studio Code deleted everything then added it back in, probably trying to auto-format the file.

Please check your patch before commit. I close this PR because we can't review the patch. Feel free to add new PR witch patches that only modify what is needed without reformatting whole files. / Thanks!

PR #1682, opened 2019-08-20
[fix] mojeek engine

Venca24:
This fixes currently not working mojeek engine while preserving pagination.

PR #1736, opened 2019-11-03
Mojeek engine hotfix

ghost:
Before the hotfix, Searx would search starting from page number 1, which Mojeek would reject with a 403 error. Changing the first page number to 0 gives the same results as doing a regular search with Mojeek: 

(403)
https://www.mojeek.com/search?q=test&s=1
(Working)
https://www.mojeek.com/search?q=test&s=0
(Same results as &s=0)
https://www.mojeek.com/search?q=test

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -744,7 +744,7 @@ engines:
     title_xpath : ./h2
     content_xpath : ./p[@class="s"]
     suggestion_xpath : /html/body//div[@class="top-info"]/p[@class="top-info spell"]/a
-    first_page_num : 1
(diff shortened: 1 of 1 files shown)


Venca24:
It's the same as #1682 by coincidence ;-)

Jyawhurh:
yes, for now it's the same, but I'm working on it and I'll try to develop it ;)



2019-11-18 16:43 GMT+01:00, Venca24 <notifications@github.com>:

asciimoo (approved):
Thank you!

PR #1744, opened 2019-11-15
[mod] speed optimization

dalf:
Speed optimizations on slow servers:
* avoid redundant call to urlparse
* get_locale(webapp.py): avoid useless call to request.accept_languages.best_match
* compile XPath only once.

About XPath: it is the optimization of the wikidata engines applied to other engines, not all.
If the modification are accepted, I will amend the commit / add a new commit to do it for all engines.


asciimoo (approved):
Great, thanks!

PR #1745, opened 2019-11-16
Fix python 3.8 compatibility 

lorddavidiii:
Archlinux has now python 3.8 and in pyhton 3.8 `escape` was removed from the `cgi` module, thus we need to import it from the `html` module.

### DIFF
--- a/searx/engines/framalibre.py
+++ b/searx/engines/framalibre.py
@@ -10,7 +10,10 @@
  @parse       url, title, content, thumbnail, img_src
 """
 
-from cgi import escape
+try:
+    from cgi import escape
+except:
+    from html import escape
 from lxml import html
 from searx.engines.xpath import extract_text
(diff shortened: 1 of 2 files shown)


unixfox:
Perfect! That's what I was looking for (I'm using Archlinux too).
@kvch @dalf Can you merge this PR?

searx/engines/framalibre.py
```diff
@@ -10,7 +10,10 @@
  @parse       url, title, content, thumbnail, img_src
 """
 
-from cgi import escape
+try:
+    from cgi import escape
+except:
```
return42, line 15: [Zen](https://www.python.org/dev/peps/pep-0020/) - lets drop python V<3.5 compatibility.

return42 (changes requested):

asciimoo (approved):
Thanks!

PR #1749, opened 2019-11-22
[fix] Force Google old UI with a new user agent

### LINKED ISSUE #1704: No results with google search
Hello,

I can't search anything if only google is selected. No matter what i search, i get the message «Sorry! we didn't find any results. Please use another query or search in more categories. »

### LINKED ISSUE #1748: Google never returns results: Sorry! we didn't find any results. Please use another query or search in more categories.
Google is never returning any results. I always get "Sorry! we didn't find any results. Please use another query or search in more categories."

To troubleshoot, I set debug : True in settings.yml

I can see the request with no errors in the searx.log file:

> DEBUG:urllib3.connectionpool:https://www.google.com:443 "GET /search?q=who+is+the+lead+developer+of+openssl&start=0&gws_rd=cr&gbv=1&lr=lang_en-GB&hl=en-GB&ei=x HTTP/1.1" 200 None
> DEBUG:searx.search:actual_timeout=2.0 (default_timeout=2.0, ?timeout_limit=None, max_request_timeout=None)

so it isn't timing out or throwing an error.

If I take the same GET and try it with curl:

`curl -A "Mozilla/5.0" "https://www.google.com/search?q=who+is+the+lead+developer+of+openssl&start=0&gws_rd=cr&gbv=1&lr=lang_en-GB&hl=en-GB&ei=x"`

I get the complete webpage result and I can the search data:

> <!doctype html><html lang="en-CA"><head><meta charset="UTF-8"><meta content="/images/branding/googleg/1x/googleg_standard_color_128dp.png" itemprop="image"><title>who is the lead developer of openssl - Google Search</title> ...

Why does searx not display the page and instead reports that it didn't find any results?

unixfox:
Google started to serve the new UI (that Searx can't parse) on Internet Explorer (all versions) so found out that by providing a very old version of Safari as a user agent Google let me load the old UI.

Why did I choose this very specific user agent? I wanted to use a quite common agent to avoid Google blocking it so picked this one which according to whatismybrowser.com is `very common`:

https://developers.whatismybrowser.com/useragents/explore/software_name/safari/1

Fixes #1748 
Fixes #1704

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -199,8 +199,9 @@ def request(query, params):
     params['headers']['Accept-Language'] = language + ',' + language + '-' + country
     params['headers']['Accept'] = 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'
 
-    # Force Internet Explorer 12 user agent to avoid loading the new UI that Searx can't parse
-    params['headers']['User-Agent'] = "Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)"
+    # Force Safari 3.1 on Mac OS X (Leopard) user agent to avoid loading the new UI that Searx can't parse
+    params['headers']['User-Agent'] = ("Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_5_4)"
+                                       "AppleWebKit/525.18 (KHTML, like Gecko) Version/3.1.2 Safari/525.20.1")
 
     params['google_hostname'] = google_hostname
 


return42:
Tested with search query "!go foofoo" before and after patched my forked branch [dm-cloud](https://github.com/return42/searx/commits/dm-cloud).

Works! / Thanks a lot!

virtadpt:
Just a thought for later: What about trying the Lynx/Links user agents later, for when this happens again?

unixfox:
@virtadpt I tried multiple user agent from this list: https://developers.whatismybrowser.com/useragents/explore/software_name/lynx/ but none of them let me load the old Google UI, so I don't think that's the good user agent for that.
Mainly because Lynx is just not that popular for Google to take into account. Whereas with Safari it's still a browser quite popular.

dimqua:
IIRC, Google Search works fine  in Lynx/Links.

unixfox:
That doesn't mean it's loading the old UI, Google has developed a special version of their new UI specially for the old browsers like IE and it's working fine even better than the old UI.

asciimoo (approved):
Great fix! Thank you!

return42:
FYI: https://github.com/asciimoo/searx/commit/2a527b818713f643b2a58fec4c61fb490e633980#commitcomment-36209713

PR #1752, opened 2019-11-24
User agents update

MarcAbonce:
I updated the user agents and I also added a shortcut in `manage.sh` to make that script more visible for everyone.


asciimoo (approved):
Great, thanks!

PR #1756, opened 2019-11-28
RFC: Makefile based boilerplate for python projects

return42:
This PR adds a boilerplate for shorter development cycles.  Its build from a subset of makefiles I used since years / [boilerplate](https://return42.github.io/boilerplate/).

I really like to have such handy make targets and the smooth emacs integration, hope other see the benefit also .. 

BTW: this series starts with a bugfix of setup.py, which is not special to this PR


asciimoo (approved):
Nice, thanks!

PR #1763, opened 2019-12-08
themes: remove CR from newlines

return42:
Adding a CR in  some files and in others not,  is  a good starting point for a
DOS+Unix mess we all have already seen in many projects.

Patch fixes all files matching (even those comming from grunt's build)::

    find ./searx -exec file {} \; | grep CR

BTW: Same with mixing TAB and SPACE indent styles in one and the same file.  So
if sources are tuched here in this patch, its also fixed.


return42:
@dalf can you please test and review this PR?

This PR removes the CR+LF mess, removes trailing white spaces, _untabify_ and removes other dust from the oscar theme.

If you think everything is fine with, please also merge this PR, since I can't merge my own PRs :)

Thanks!

return42:
Forgot to say: the comit https://github.com/asciimoo/searx/pull/1763/commits/57c6fa6e in this patch series also fix the (long broken) auto-submit when changing the language.

asciimoo (approved):
Thanks!

searx/templates/oscar/search.html
```diff
@@ -3,7 +3,7 @@
   <div class="row">
     <div class="col-xs-12 col-md-8">
       <div class="input-group search-margin">
-        <input type="search" name="q" class="form-control" id="q" placeholder="{{ _('Search for...') }}" aria-label="{{ _('Search for...') }}" autocomplete="off" value="{{ q }}">
```
dalf, line 6: In https://github.com/asciimoo/searx/blob/master/searx/static/plugins/js/vim_hotkeys.js the key is ```i``` (without control keys). Not sure if it creates confusion or not between the two shortcuts : 
* ```<alt>``` + ```<shift>``` + ```s```
* ```i```

It seems there is no de facto standard : 
* ddg uses ```/``` and ```h```
* wikipedia uses ```h```
* github uses ```/``` and ```s```

return42: > Not sure if it creates confusion or not 

both work .. why should one be confused?

> It seems there is no de facto standard ...

As far I can tell, they do not use attribute ````accesskey```` (it is without control keys)

dalf: From user point of view, I have to know that it is neither ```<alt>``` + ```<shift>``` + ```i``` nor ```s``` but the other way around.

Issue #729, opened 2016-10-12
Google Captcha


sut12:
Hey,

i have the problem that google gives me captchas. Therefor i cant see google results. My Server is not blacklistet anywhere. So im wondering what i can do to get rid of the captchas. My instance is not hosted under https. Is that maybe a reason for the google captchas?

Thanks!

dalf:
If you can, use a browser using the same IP than your server : google will throw some captchas at first, and will usually stop after few requests.
One way to do it : install Firefox on the server, then use ssh with X forwarding or VNC.

Not sure, if you need to check if your IP is blacklisted ( how to check : http://www.dnsbl.info/dnsbl-database-check.php not sure either that this is right service to  use)

sut12:
Hello,
i used elinks. It is possible to save images. So i saved the captcha and downloaded it afterwards. For now it is working but i had to enter the captcha already more than once.

I am not listed on any blacklist except on spamcannibal:
generic/anonymous/un-named IP

I need to check how i can get rid of this message. But im not sure if it has something to do with the google problem.

sut12:
It is so wired. Yesterday evening it worked after i enterd the captchas. This morning it didnt work. 

Now it works but i didnt enter any captchas since yesterday...

cy8aer:
Google is dead for me (as startpage).

sut12:
I need google. But its working sometimes and sometimes not. I even tried to analyze with google webmaster tools...

Right now its working - 2morrow maybe not.

cy8aer:
Seems that google sets GOOGLE_ABUSE_EXEMPTION cookie after captcha. And this cookie has a timeout date (sometimes). If there would be some proxying mechanism for the captcha for re-setting the cookie would be nice

cy8aer:
what about forwarding the captcha to the frontend and save the cookie in the backend then?

sut12:
I think that would be the best solution.

ghost:
I wonder how searx.me(and other searx admin) handle google captcha.
Sure they process 1K+/day request to google without getting "scroogled".

gszathmari:
I am having the same issue. It doesn't help if I solve the CAPTCHAs from a browser as the `GOOGLE_ABUSE_EXEMPTION` should be sent with my searx queries.

kvch:
@gszathmari What do you mean? What is `GOOGLE_ABUSE_EXEMPTION`?

gszathmari:
[...]

PR #1771, opened 2019-12-17
[RFC] prototyping antibot methods (#729)

return42:
This is a **R**equest **F**or **C**omments / don't merge upstream master!

Within this patch series I will prototype & discuss methods to block spam bots.

The need to block spam bots came up in issue https://github.com/asciimoo/searx/issues/729

@unixfox shares his proxy based solution: https://github.com/asciimoo/searx/issues/729#issuecomment-511997511 which inspired me to do similar builtin.

For prototyping I started my https://github.com/return42/searx/tree/antibot-proto branch I share with you in this RFC.

I test this series on my searx instance: https://darmarit.cloud/searx

With the first commit a simple blocking method, based on a Flask session cookie is implemented:

- enforce (session) cookie on client site will block spam bots without cookie storage.

Tested 24h .. seems to block most of the spam bots / no more google captcha since 24h

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -48,6 +48,7 @@ except:
 from datetime import datetime, timedelta
 from time import time
 from werkzeug.contrib.fixers import ProxyFix
+from flask import session
 from flask import (
     Flask, request, render_template, url_for, Response, make_response,
     redirect, send_from_directory
@@ -489,6 +490,27 @@ def index():
 
     Supported outputs: html, json, csv, rss.
     """
+
+    def hash_client(request):
+        _ = hashlib.sha256()
+        _.update(request.headers.get('User-Agent').encode('utf-8'))
+        _.update(request.remote_addr.encode('utf-8'))
+        return _.digest()
+
+    if not 'searx_antibot' in session:
+        # install session cookie first / redirect same URL
+        # !! requires cookies to function properly !!
+        searx_antibot = hash_client(request)
+        logger.info('searx_antibot: await session cookie: %s ', searx_antibot)
+        session['searx_antibot'] = searx_antibot
+        return redirect(request.url)
+
+    searx_antibot = session.get('searx_antibot')
+    if searx_antibot != hash_client(request):
+        logger.info('searx_antibot: got invalid digest .. redirect')
+        return redirect(request.url)
+
+    logger.info('searx_antibot: cookie OK: %s ', searx_antibot)
 
     # output_format
     output_format = request.form.get('format', 'html')


unixfox:
I certainly can confirm you that you will most likely blocked pretty soon because I have seen a large amount of bots supporting cookies and even in curl cookies jar is supported to act like a cookie storage.
That's why my antibot program doesn't filter bots just with cookies.

Moreover, asciimoo will have to change [the main page of Searx](https://asciimoo.github.io/searx/#features) to remove this feature: "Cookies are not used by default".

Also keep in mind that there are people that turn off cookies, when I deployed my solution to searx.be I received an email of a user stating that he couldn't reach my searx instance due to that and no help page was given to him to let him understand the situation (still the case for my antibot but that's because I'm still rewriting a new one from scratch).

searx/webapp.py
```diff
@@ -490,6 +491,27 @@ def index():
...
+        _.update(request.remote_addr.encode('utf-8'))
+        return _.digest()
+
+    if not 'searx_antibot' in session:
+        # install session cookie first / redirect same URL
+        # !! requires cookies to function properly !!
+        searx_antibot = hash_client(request)
+        logger.info('searx_antibot: await session cookie: %s ', searx_antibot)
+        session['searx_antibot'] = searx_antibot
+        return redirect(request.url)
```
unixfox, line 506: You need to redirect with a status code of 307 to keep the data sent by a POST method: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307
If you do a simple redirect with a status code of 302, the client/browser won't send the data again to the server and this will make the visitor kinda angry that he has to submit again his search.

unixfox (changes requested):

searx/webapp.py
return42: Correct, there is a _jitter_ at the first time. 

> ... 307 guarantees that the method and the body will not be changed when the redirected request is made.

Was what I am looking for / thanks!

searx/webapp.py
```diff
@@ -490,6 +491,27 @@ def index():
     Supported outputs: html, json, csv, rss.
     """
 
+    def hash_client(request):
+        _ = hashlib.sha256()
+        _.update(request.headers.get('User-Agent').encode('utf-8'))
```
unixfox, line 496: Using the User-Agent for the hash is a very bad idea because in the Searx community using a user agent switcher extension is quite common so this will most likely annoy these people if they get redirected too often.

return42: good point / thanks!

I really like to tag the application behind a query, not only the IP where it comes from. Did you have any idea which data might be useful to identify the application?

dalf:
Not sure, if the cookie contains the hash_client() result (signed with itsdangerous), the same cookie value can be used everywhere.
In this case, I think the secret key from settings.yml and a timestamp (rounded to hour, day, whetever) should be added.

return42:
FYI:  [Flask Sessions](https://flask.palletsprojects.com/en/1.1.x/quickstart/#sessions)

> Flask Session is implemented on top of cookies for you and signs the cookies cryptographically. What this means is that the user could look at the contents of your cookie but not modify it, unless they know the secret key used for signing.

asciimoo:
This isn't a reliable solution and it introduces cookie sessions which can harm users privacy.
I'm using [filtron](https://github.com/asciimoo/filtron) to fight against bots and it works well for me.

Bot detection is an issue only for the public instances, so I'm not sure if implementing such a feature in searx is a good idea.

unixfox:
I completely agree with you @asciimoo except for the part where you say that filtron works well because I don't think filtron is efficient enough to combat against bots due to the fact that most of them implemented a way to circumvent filtron.
As an example searx.me doesn't work with Google because it is blocked by a captcha.

virtadpt:
Where would this leave private instances?
The single reason I use Searx is because I have a fleet of bots that I use for intelligence gathering.  I have a couple of private instances because I don't want to hammer anyone else's unnecessarily.

asciimoo:
@unixfox filtron is as good as the rules you are using.

PR #1775, opened 2019-12-18
Add Ability to Change Search Box Text(placeholder)

michaelfm1211:
Note: This is a revised PR of #1732 

Adds a searchbox_text field in settings.yml(under ui) which changes the default placeholder of "Search for ..." to whatever is in that field.


return42:
> Note: This is a revised PR of #1732

Thanks for stripping the diff! .. Now I can see what you have changed. I am sorry to say, that I am not happy with this solution:

1. Whats about i18n (translation)
2. Where is the need for? .. What are the use cases?

Issue #1649, opened 2019-07-25
Results fetched from docmentation sites are shown as raw HTML


etanot:
Search results which are fetched from [git-scm](https://git-scm.com) is in HTML language.

![git2](https://user-images.githubusercontent.com/28824347/61836998-5362cf00-ae72-11e9-9a23-a83ec69fad69.png)

Whereas, on [git-scm](https://git-scm.com/docs/git-add) result is shown as follows

![git-scm](https://user-images.githubusercontent.com/28824347/61837439-5363ce80-ae74-11e9-943d-b8b24dd355a8.png)

Search any query related to git (like git add, git commit, etc), if it's fetched from git-scm, it should be in HTML language, inside the info box.

**Edit:**

Seems like, problem is not only with git-scm; but with most of the documentation site related to programming language. For instance try these searches:

* Python3 docs: [!ddd python math.sqrt](https://searx.me/?q=%21ddd%20python%20math.sqrt&categories=none&language=en-US)
* Nodejs docs: [!ddd node repl](https://searx.me/?q=%21ddd%20node%20repl&categories=none&language=en-US)
* Cppreference: [!ddd c sqrt](https://searx.me/?q=%21ddd%20c%20sqrt&categories=none&language=en-US)

dalf:
URL to reproduce the bug (doesn't happen with the simple theme):
https://searx.me/?q=!ddd+git+add&language=en-US&time_range=&category_general=on&safesearch=0&theme=oscar

Most probably, line 8 on infobox.html, missing ```| safe```:
* https://github.com/asciimoo/searx/blob/master/searx/templates/oscar/infobox.html#L8
* https://github.com/asciimoo/searx/blob/master/searx/templates/simple/infobox.html#L5

etanot:
@dalf, Can I make a pull request for this issue?

PR #1781, opened 2019-12-21
Oscar: move info box at top of the page

### LINKED ISSUE #1777: move info box at top of the page (in mobile devices)
#### Affected Theme
* Oscar

#### Description
In low width (for example mobile) devices, `Info Box` is at bottom of the page, which should be at top, with an expansion icon which expands the `info box` to its original size, if user wants to see more detailed info box.

searx me

**Fig 1:** Screenshot is taken on Android device for [radix sort](https://searx.me/?q=radix%20sort&categories=general&language=en-US) search term.

In contrast, we can look at other search engines result page, for example DDG engine:

ddg-without-expansion

ddg

**Fig 2:** Screenshot taken on same device for same search term, but on DDG engine.

#### Steps to reproduce
* Search from low width device on any Searx instance. For example, search this "[radix sort](https://searx.me/?q=radix%20sort&categories=general&language=en-US)" term.
* Scroll to the bottom of the page.

You'll see the info box is at bottom of the page.

etanot:
####  Oscar: move info box at top of the page #1777 
In low width devices like mobile, tablet etc, info box is present at
bottom of the page instead of top of the page.

This issue addresses by rearranging column grids for low width devices. See
  - https://getbootstrap.com/docs/3.3/css/#grid-column-ordering.
  - and Searx issue tracker (issue#1777), for more information.

Effect: Along with Info, Suggestion and Link boxes also move to top of
the page.

Resolves #1777

---------------------------------------------------------------------------------------------------------------

Here, how it looks like on on different devices,

* Low width devices
  * Portrait

mobile-portrait

   * Landscape

mobile-landscape

 * Desktop

desktop


unixfox:
To close an issue when a PR is merged you need to write: `fix #issuenumber`: https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

finn0:
Here, how it looks like on on different devices,

* Low width devices
  * Portrait

mobile-portrait

   * Landscape

mobile-landscape

 * Desktop

desktop

finn0:
Second,
#### Oscar: no HTML escaping prior to output #1649 
When results are fetched from any programming related documentation site
(like git-scm.com, docs.python.org etc), content in Info box is shown as
raw HTML code.

This change addresses the issue by using "safe" filter feature provided by
Django. See,
  - https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#safe
  - Searx issue tracker (issue #1649), for more information.

Resolves: #1649

----------------------------------------------------------------------------------------------------------------------
After changes, it looks like this,

no-raw-html

finn0:
Sorry, for mess (lots of spam emails). At the time of rebasing, I've rebased someone else code also, that's why I've to force push commits to this branch.

return42 (changes requested):
I reviewed this PR .. seems to work. But for me it is not clear

1. Info box on top, is it what the community wants?

> But, I don't like those boxes, Suggestion Box and Link Box at the top of the page. In my humble opinion moving them to bottom of the page would be good idea. What do you think? 

2. would this be changed in the future?

From my very personal opinion: I like info boxes on top, but Suggestion Box and Link Box should not be on top.  Patch https://github.com/finn0/searx/commit/36ca2dcc56651688cd7ca8b8d925a9370bdc1dec moves all to the top and seems not the solution.

Next point: That info boxes ATM are not at top is not a bug, so this patch is not a bug-fix, its just an "enhancement" while the patch https://github.com/asciimoo/searx/pull/1781/commits/6a5aae65308 is a bug fix.

I really want to see the fix and the "enhancement" discussed in different PRs.

Can you please separate the bug fix  https://github.com/asciimoo/searx/pull/1781/commits/6a5aae65308  into another PR / I really like to see this fix merge asap / thanks!

etanot:
> 1. Info box on top, is it what the community wants?

I asked about this to @dalf, (see https://github.com/asciimoo/searx/issues/1777#issuecomment-567527789), he said that those aren't intentional. And `info box` is for to quickly get information about search results, so it don't make any sense to have them at bottom of the page.

> 1. would this be changed in the future?

Yep, my CSS isn't good, so as soon as I find a solution, I'll makes those changes or if some else wants to take over this enhancement, your welcome :)

> Suggestion Box and Link Box should not be on top

I'm agree with you, Link box shouldn't be at the top of the page and that's why I asked for opinions on issue page about the design but I've doubts about where to put `Suggestion Box`; some search engines, show suggestions at top of the page, so having suggestion box top of the page helps users to quickly correct their mistakes, but this becomes creepy when user click on `Next Page`.

> seems not the solution.

Indeed! if you want to, you can block this merge request until we don't find a good solution.

> Can you please separate the bug fix

Okay, I'll do it now.

Edit:
And what would if `Infinite Scroll` plugin is on, user will never gonna see boxes which are present at the bottom of the page, unless they exhaustively scroll to the bottom of the page.

etanot:
@return42, done. See pull request https://github.com/asciimoo/searx/pull/1787.

return42:
> I asked about this to @dalf, (see #1777 (comment)), he said that those aren't intentional. And info box is for to quickly get information about search results, so it don't make any sense to have them at bottom of the page.

@finn0 thanks a lot for clarifying!  .. then we can switch the layout!

> I'm agree with you, Link box shouldn't be at the top of the page ...

> And what would if Infinite Scroll plugin is on, user will never gonna see boxes ...

.. these are good points to think about.

> if you want to, you can block this merge request until we don't find a good solution.
 
Close this PR or continue your work here .... its up to you / I do not have any preference.

> See pull request #1787.

again; thanks! .. i have merged PR into master

return42 (dismissed):

dalf (approved):
Works as expected:
* I've tried left to right and right to left language (English, Hebrew languages)
* phone, tablet, normal view
* infinite scroll
* the only thing is the download links box just before the results. Perhaps another PR could replace the box by a [dropdown button](https://getbootstrap.com/docs/3.3/components/#dropdowns).

etanot:
@dalf,
Thanks for your feedback. For now, I'm opening an issue to track this enhancement proposal; I'll send the PR, when I'm free (probably next month).

PR #1791, opened 2019-12-30
misc docs (WIP)

return42:


dalf:
I answer here to https://github.com/dalf/searx-stats2/issues/7

From searx-stats2 project, it would be easier if the instance list would be generated from a .yaml file. It would also remove some boilerplate from the instance list.

The file format can be:
- ```url``` : https URL (mandatory)
- ```addtional_urls```: (optional)
   - ```url```: searx instance URL (example https://search.gibberfish.org/tor/ )
   - ```relation```: comment about the link (example _Proxied through Tor_ )
- ```comment``` (optional)
- ```issuer``` (mandatory)
- ```tags```
   - Alive and running
   - Running in exclusive private walled-gardens
   - Running with an incorrect SSL certificate
   - Offline
   - see also https://github.com/dalf/searx-stats2/issues/6#issuecomment-570817601

Example:
```yaml
- url: "https://search.gibberfish.org/"
  issuer: "Let's Encrypt"
  tags: 
    - "alive"
  addtional_urls:
    - url: "http://o2jdk5mdsijm2b7l.onion/"
      relation: "Hidden Service"   
    - url: "https://search.gibberfish.org/"
      relation: "Proxied through Tor"
```

From this file, searx can generate the .rst file:
- the link to www.ssllabs.com can be generated including to the ```additional_urls``` (BTW It seems everyone agree to make a link to it).
- I think the sort order should be enforced in the .yaml file (because otherwise it introduces two sort orders: one in the .yaml file, one in the .rst file).
- searx UI can fetch and display this list (if it makes sense to filter / sort this list from the searx instance, otherwise a link as now is fine).

But:
* the .rst file is fine from searx-stats2 point of view.
* this can be the subject of another PR.

return42:
> From searx-stats2 project, it would be easier if the instance list would be generated from a .yaml file. It would also remove some boilerplate from the instance list.

This https://searx.space/instances/ is much more informative than the wiki entry or a hand made list in reST format!

KISS: the instance list is better maintained by / hosted at searx-stats2

In the searx docs we should only list meta-searx instances and a link to https://searx.space/instances/

dalf:
In this case to make it even more simple, may be:
- move https://searx.space/instances/ to https://searx.space/ (more easy to remember)
- Note: the current https://searx.space/ content must be integrated into searx-stats2 IHMO.
- perhaps rename the searx-stats2 project to searx-space to make it more straightforward.

return42:
To stop vandalism of the wiki, first I want to see this PR merged.  For this I need a review of this PR .. do you have time ;)

return42:
> In this case to make it even more simple, may be:
> move https://searx.space/instances/ to https://searx.space/

I do not have access / lets do it afterwards?

>  Note: the current https://searx.space/ content must be integrated into searx-stats2 IHMO.
>  perhaps rename the searx-stats2 project to searx-space to make it more straightforward.

I do not have any preference nor access .. if we need "other" for that, we should not wait about.

For me it is no problem to use what we have,  searx-stats2 is OK and also https://searx.space/instances/

dalf:
In docs/conf.py:
```python
html_context = {
    "project_links": [
        ProjectLink("Source", GIT_URL),
        ProjectLink("Wiki", "https://github.com/asciimoo/searx/wiki"),
        ProjectLink("Public instances", "https://github.com/asciimoo/searx/wiki/Searx-instances"), # not the right link
        ProjectLink("Twitter", "https://twitter.com/Searx_engine"),
    ]
}
html_sidebars = {
    "**": ["project.html", "relations.html", "searchbox.html"],
}
```
I'm not sure how to add the link to user/public_instances.html

----

I host searx.space (OVH server), so there is no problem to move the list to serax.space when / if it is okay to do it. Same for renaming searx-stats2 project. I prefer to move the content now rather later to avoid dead links.

If we keep searx.space/instances the "next" question is: what should display searx.space without /instances ?

return42:
> I'm not sure how to add the link to user/public_instances.html

Thanks for pointing out / done.

dalf:
when I click on _Public instances_ using ```make docs-live```, the link goes to https://asciimoo.github.io/searx/user/public_instances.html not the local version.

[EDIT] in the end, from the user point of view, the link will work as expected, it is only "disturbing" when the documentation is produce locally.

return42:
> when I click on Public instances using make docs-live, the link goes to ..

that is correct. It is only a interim solution / You are right, ATM it would be better a local ref but let things keep simple .. in long term we have to point again to a project with a full URL ( https://searx.space/instances/ )

dalf (approved):
If only I consider moving the content: go.
There are updates to do here and there, but it was already the case with existing wiki documentation ; it can be the subject of other PRs.

unixfox:
Next move is to make the wiki read only and redirect to the documentation.
Just don't delete every pages of the wiki, a simple note like "the wiki moved here" is enough. Would be great to not break the indexed wiki pages.

return42:
> Just don't delete every pages of the wiki, a simple note like "the wiki moved here" is enough.

Argh, saw your post to late .. should I rollback the wiki entry? https://github.com/asciimoo/searx/wiki/Searx-instances

unixfox:
> > Just don't delete every pages of the wiki, a simple note like "the wiki moved here" is enough.
> 
> Argh, saw your post to late .. should I rollback the wiki entry? [asciimoo/searx/wiki/Searx-instances](https://github.com/asciimoo/searx/wiki/Searx-instances)

If you are able to do that then yes and also ask @asciimoo to make the wiki read only in the settings of the repository instead of having a "DO NOT EDIT THIS wiki entry any longer!".

unixfox:
I didn't mean to actually bring back the old wiki content, just a simple note on each page that the wiki moved to a new location.
Just like for the netdata wiki: https://github.com/netdata/netdata/wiki/Configuration

About searx-stats2, probably better to wait that the project is finished than publicly display the URL. Especially when [this issue](https://github.com/dalf/searx-stats2/issues/6) is still unresolved.

return42:
@asciimoo can you please write protect the https://github.com/asciimoo/searx/wiki/Searx-instances wiki entry / thanks!

@unixfox rolled back old content .. but I think we should drop the list on a longer term, I can't see any use in a old unmaintained list.

return42:
Again I see your post after I posted the last comment .. 

I gave up .. do whatever you think with the instance list .. I am out.

unixfox:
That's fine.
Something that is missing is an "edit on GitHub" banner on top right similar to this banner in this screenshot:


Also would be great to have a better domain for the wiki, like wiki.searx.me instead of the ugly "asciimoo.github.io/searx"

return42:
I'm a little slow on the uptake  .. for what do you want a github corner? .. for the whole docs, only for https://asciimoo.github.io/searx/user/public_instances.html or  ... 

> Also would be great to have a better domain for the wiki, like wiki.searx.me instead of the ugly "asciimoo.github.io/searx"

are you talking about docs or wiki? .. Anyway please add a issue and describe what you expect.

unixfox:
> I'm a little slow on the uptake .. for what do you want a github corner? .. for the whole docs, only for [asciimoo.github.io/searx/user/public_instances.html](https://asciimoo.github.io/searx/user/public_instances.html) or ...
> 
> > Also would be great to have a better domain for the wiki, like wiki.searx.me instead of the ugly "asciimoo.github.io/searx"
> 
> are you talking about docs or wiki? .. Anyway please add a issue and describe what you expect.

For the whole documentation for both case.

asciimoo:
> @asciimoo can you please write protect the https://github.com/asciimoo/searx/wiki/Searx-instances wiki entry / thanks!

I restricted wiki editing to collaborators only. It isn't possible to protect only that page AFAIK.

>  Also would be great to have a better domain for the wiki, like wiki.searx.me instead of the ugly "asciimoo.github.io/searx"

Good idea, I'm going to do this.

return42:
> For the whole documentation for both case.

Do we really need "github" corners? .. we have "Project Links" on top of the left navigation bar and the link to the "Source" is the topmost .. compare: https://asciimoo.github.io/searx/

return42:
>> Also would be great to have a better domain for the wiki, like wiki.searx.me instead of the ugly "asciimoo.github.io/searx"

Lets name it "documentation"  or "docs" / the wiki is not the documentation.

> Good idea, I'm going to do this.

If done, please give feedback. I will fix the links and (BTW) tidy up the DOCS_URL environment.

Thanks

````
./Makefile\05:export DOCS_URL=https://asciimoo.github.io/searx
./CHANGELOG.rst\0227:New documentation page is available: https://asciimoo.github.io/searx
./docs/conf.py\09:DOCS_URL = os.environ.get("DOCS_URL", "https://asciimoo.github.io/searx/")
./docs/conf.py\097:        ProjectLink("Public instances", "https://asciimoo.github.io/searx/user/public_instances.html"),
./docs/dev/makefile.rst\049:  export DOCS_URL=https://asciimoo.github.io/searx
./docs/dev/contribution_guide.rst\0180:   doc available at --> https://asciimoo.github.io/searx
./searx/engines/searx_engine.py\05: @provide-api yes (https://asciimoo.github.io/searx/dev/search_api.html)
./searx/templates/oscar/base.html\088:                    {{ _('Powered by') }} <a href="https://asciimoo.github.io/searx/">searx</a> - {{ searx_version }} - {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
./README.rst\012:See the `documentation <https://asciimoo.github.io/searx>`__ and the `wiki <https://github.com/asciimoo/searx/wiki>`__ for more information.
./README.rst\026:For all of the details, follow this `step by step installation <https://asciimoo.github.io/searx/dev/install/installation.html>`__.
````

unixfox:
Moved to https://github.com/asciimoo/searx/issues/1785#issuecomment-571501143

PR #1818, opened 2020-01-29
Update

ghost:
new Searx instance - http://nibblehole.com


return42:
Thanks for your addition!. We are on the way to move the list of instances to https://github.com/dalf/searx-stats2 / I added your request to https://github.com/dalf/searx-stats2/issues/12

The switch to searx-stats2 is not yet fully completed, so we have not yet documented this. Sorry for the intervening chaos.

PR #1823, opened 2020-02-01
Add access tokens to engines

kvch:
From now on it is possible to configure tokens for engines. This lets instance administrators limit the access to engines. Thus, making engines private.

A new setting is added to engines, named `tokens`. It expects a list of arbitrary strings. If a user presents one of the tokens when making a search request to searx, he/she is granted access to the engine.

Example configuration with the "well-known" Frinkiac engine:

```yaml
 - name : frinkiac
   engine : frinkiac
   shortcut : frk
   tokens : 
   - token1
   - token2
```

Users can add their tokens on the preferences page. The new setting expects a list of comma separated tokens.

image

Important: At the moment, administrators are required to create tokens and distribute them to their users.

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -38,6 +38,7 @@ def check_settings_yml(file_name):
     else:
         return None
 
+
 # find location of settings.yml
 if 'SEARX_SETTINGS_PATH' in environ:
     # if possible set path to settings using the
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -54,7 +54,8 @@ engine_default_args = {'paging': False,
                        'suspend_end_time': 0,
                        'continuous_errors': 0,
                        'time_range_support': False,
-                       'offline': False}
+                       'offline': False,
+                       'tokens': []}
 
 
 def load_engine(engine_data):
@@ -160,7 +161,7 @@ def to_percentage(stats, maxvalue):
     return stats
 
 
-def get_engines_stats():
+def get_engines_stats(preferences):
     # TODO refactor
     pageloads = []
     engine_times = []
@@ -171,8 +172,12 @@ def get_engines_stats():
 
     max_pageload = max_engine_times = max_results = max_score = max_errors = max_score_per_result = 0  # noqa
     for engine in engines.values():
+        if not preferences.validate_token(engine):
+            continue
+
         if engine.stats['search_count'] == 0:
             continue
+
         results_num = \
             engine.stats['result_count'] / float(engine.stats['search_count'])
 
--- a/searx/engines/dummy-offline.py
+++ b/searx/engines/dummy-offline.py
@@ -1,0 +1,12 @@
+"""
+ Dummy Offline
+
+ @results     one result
+ @stable      yes
+"""
+
+
+def search(query, request_params):
+    return [{
+        'result': 'this is what you get',
+    }]
--- a/searx/engines/genius.py
+++ b/searx/engines/genius.py
@@ -72,6 +72,7 @@ def parse_album(hit):
             result.update({'content': 'Released: {}'.format(year)})
     return result
 
+
 parse = {'lyric': parse_lyric, 'song': parse_lyric, 'artist': parse_artist, 'album': parse_album}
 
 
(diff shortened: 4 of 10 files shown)


sevospl:
I'd love to see that being merged! It's such a great idea.

return42:
> Important: At the moment, administrators are required to create tokens and distribute them to their users.

This is a very interesting problem I have to think about.  Are there any methods or concepts solving such problems?

asciimoo:
> This is a very interesting problem I have to think about. Are there any methods or concepts solving such problems?

Do we need to solve this problem? Or, maybe a better question: is this a problem?

return42:
> Or, maybe a better question: is this a problem?

Privacy and the exchange of tokens are in conflict.

> Do we need to solve this problem?

Not necessarily, but we should consider to make suggestions if there are well known pattern do exist for this question.

BTW: I miss the admin & user documentation in this patch series

asciimoo:
> Privacy and the exchange of tokens are in conflict.

The admin can decide who has access to the private engines, why does it conflict with privacy?

return42:
> The admin can decide who has access to the private engines, why does it conflict with privacy?

Maybe I am totally  barking up the wrong tree ...

To deliver the token to the user, the admin must know the user's _address_ and store it somewhere. I now wonder if there are any procedures that allow the user to obtain a token anonymously. This would rule out any kind of tracking.

kvch:
>> This is a very interesting problem I have to think about. Are there any methods or concepts solving such problems?
>
> Do we need to solve this problem? Or, maybe a better question: is this a problem?

Ideally, it would be nice to provide advice to administrators on how to distribute tokens. But I agree with @asciimoo . I think it's similar "problem" as providing a sample NGINX configuration for the web server in our documentation. We can suggest administrators how to do it or what tools to use. However, it's up to them which software/approach to take. 

>> The admin can decide who has access to the private engines, why does it conflict with privacy?
>
>Maybe I am totally  barking up the wrong tree ...
>
>To deliver the token to the user, the admin must know the user's _address_ and store it somewhere. I now wonder if there are any procedures that allow the user to obtain a token anonymously. This would rule out any kind of tracking.  

I don't think an admin needs to know the user's address. There are ways of exchanging information without learning one's address (e.g. meeting up in person). But I am sure, that the admin does not need to store the address or anything about the user. All the admin needs to know is the token he/she generated.

The idea is that administrators not just hand out tokens to everyone anonymously. But rather, selects individuals/groups he/she wants to provide access to some engines because he/she trusts them. Or just add a private engine to his/her own instance, which should not be publicly available to anyone besides him/her. But still wants to open all other engines of the instance for everyone.

sevospl:
> > The admin can decide who has access to the private engines, why does it conflict with privacy?
> 
> Maybe I am totally barking up the wrong tree ...
> 
> To deliver the token to the user, the admin must know the user's _address_ and store it somewhere. I now wonder if there are any procedures that allow the user to obtain a token anonymously. This would rule out any kind of tracking.

I honestly don't mind it. If someone wants to set up a private Searx instance, he/she wouldn't want (like me) to share it with everyone else out there. However, I'm okay with it being open for everyone, as long as it doesn't apply for Google and torrents as well. So like DuckDuckGo or Startpage etc.

If you want to keep your privacy & be anonymous, just don't demand for a private engine. Simple as that. :) That's just my opinion though.

return42:
@kvch  and @sevospl  thanks for your excellent explanations! .. 

> BTW: I miss the admin & user documentation in this patch series

Adding your explanation in the /docs would be great / Thanks!

kvch:
> @kvch and @sevospl thanks for your excellent explanations! ..
> 
> > BTW: I miss the admin & user documentation in this patch series
> 
> Adding your explanation in the /docs would be great / Thanks!

Sure. I might add it in a follow up PR. In that case I will open an issue to make sure I won't forget it.

stef:
the admin might post the tokens to a broadcast medium that only the intended users can access, for example a company internal intranet website, where the company also hosts a public searx. i'm sure there are other examples where the admins do no not have to know per se the users.

return42:
@kvch please read ["How to contribute", chapter "Code"](https://asciimoo.github.io/searx/dev/contribution_guide.html#code) and use meaningful commits / Thanks!

kvch:
@return42 I fixed up the commits and reworded the message. I have also added a few unit tests to make sure the feature works as intended.

kvch:
I have also opened a follow-up issue for documentation.

return42 (approved):

PR #1834, opened 2020-02-07
Update settings.yml to put Wolfram|Alpha in the General category 

geckolinux:
Put Wolfram|Alpha in the general category, so it gives smart instant answers without cluttering results with scientific whitepapers.


unixfox:
Same for this PR

asciimoo:
I'm not sure if this is a good default because wolframalpha is slow and sometimes unreliable. You can get smart instant answers any time using bang syntax (e.g. with the search query `!wa !general xyz`)

PR #1835, opened 2020-02-07
[fix] Disable HTTP keep alive

dalf:
Fix HTTP 429 response from filtron
See https://github.com/searx/searx-docker/issues/24

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -31,3 +31,7 @@ touch-logrotate = /run/uwsgi-logrotate
 unique-cron = 15 0 -1 -1 -1 { touch /run/uwsgi-logrotate  }
 log-backupname = /var/log/uwsgi/uwsgi.log.1
 logto = /var/log/uwsgi/uwsgi.log
+
+# No keep alive
+# See https://github.com/searx/searx-docker/issues/24
+add-header = Connection: close


return42 (approved):
Thanks!  .. I will merge this PR, but ..

are you sure this is what caused https://github.com/searx/searx-docker/issues/24 ?

dalf:
Sorry for the lack of comment.

After seeing the filtron logs:
```
Response error: the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection HTTP/1.1 200 OK
```

* I've tried to use the last version [valyala/fasthttp](https://github.com/valyala/fasthttp/) just in case : same error.
* Reading https://uwsgi-docs.readthedocs.io/en/latest/HTTP.html#http-keep-alive I've tried ```add-header = Connection: close``` on a.searx.space as recommended by fasthttp : 
  * in rules.json, I've increased the 25 requests per 3 seconds limit  to 300 ( https://github.com/searx/searx-docker/blob/b4ce81ab604df1fe5bc6e5d8666e542335c665d0/rules.json#L9-L16 )
  * I've checked many times and the error has disappeared (until the 300 req / 3 sec)
  * filtron logs: empty after 24h (but a.searx.space is not too much used).

May be this is not the best way to solve this issue.

Perhaps fasthttp (so filtron) can make http 1.0 connections, but it wouldn't have solve the issue in the searx docker image.

return42:
@dalf thanks for explanation .. we have to wait and see ...

I was asking because when I implemented https://github.com/asciimoo/searx/pull/1803 I got also a zero Byte problem .. can't remember how I solved the problem without setting headers.

dalf:
The issue was puzzling me and I was asking myself if you had this issue with #1803

I've updated the searx/searx docker image, and  _wait and see_

PR #1845, opened 2020-02-14
Move my old instance to offline section + add my new one in running section

lonequantum:


dalf:
The instance is in the process to move here : https://github.com/dalf/searx-instances/

Can you create two issues here: https://github.com/dalf/searx-instances/issues/new/choose
* one to delete searx.zareldyn.net
* one to add searx.lnode.net

The result is displayed here: https://searx.space/

Very sorry for the lack of information.

PR #1846, opened 2020-02-14
Add Swisscows autocomplete option

ghost:
This adds an autocomplete option for the Swisscows suggest api. It pulls in a maximum of 5 items.


return42:
Should we add autocompletion from Swisscows / https://github.com/asciimoo/searx/issues/1859#issuecomment-589952307

I do not have any preference, it's just a question I'm asking around .

dalf:
The autocompletion doesn't require a token, so the restrict described in https://github.com/asciimoo/searx/issues/1859#issuecomment-589952307 doesn't apply here.

This swisscow autocompletion seems to return less results than others, but it works.

It can be merged in my opinion.

return42 (approved):
Thanks!

PR #1856, opened 2020-02-21
[enh] use httpx instead of requests

dalf:
Demo implementation of https://github.com/asciimoo/searx/issues/899#issuecomment-589593476
* All code is sync expect the httpx calls.
* Require Python 3.6 (Python 3.5 end of life: 13 Sep 2020)
* Support HTTP/2 and brotli. (require the ```httpx[http2]``` package, see https://www.python-httpx.org/http2/#enabling-http2 )
* ~~There is a bug: ... TypeError: '_asyncio.Task' object is not callable~~
* Stream is not supported ( https://github.com/asciimoo/searx/blob/master/searx/webapp.py#L822 ) (1)
* ~Multiple source IP are ignored (2). But httpcore [supports source IP](https://github.com/encode/httpcore/blob/master/CHANGELOG.md#0100-august-7th-2020). See https://github.com/encode/httpx/pull/1165~
* ~~The settings pool_connections and pool_maxsize doesn't exist in HTTPX.~~ (see https://www.python-httpx.org/advanced/#pool-limit-configuration )

Except (1) and (2), poolrequests.py API is compatible with the existing implementation, so theoretically, the choice between requests and httpx can be done at runtime.

httpx supports socks proxy using an external package: https://github.com/romis2012/httpx-socks also see https://github.com/encode/httpcore/pull/186

But frankly speaking, it would be better to first drop Python 2 support, then switch to an async architecture (*) and use httpx.

(*) modulo available packages in the different distros, but that's another topic.

Note: see https://github.com/dalf/pyhttp-benchmark/blob/master/results/output.md to have benchmarks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ stages:
 
 jobs:
   include:
-    - python: "3.5"
     - python: "3.6"
     - python: "3.7"
     - python: "3.8"
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,6 @@ pygments==2.1.3
 pyopenssl==19.1.0
 python-dateutil==2.8.0
 pyyaml==5.3.1
-requests[socks]==2.24.0
+httpx[http2]==0.16.1
+brotlipy==0.7.0
+uvloop==0.14.0
--- a/searx/engines/dictzone.py
+++ b/searx/engines/dictzone.py
@@ -60,7 +60,7 @@ def response(resp):
                 to_results.append(to_result.text_content())
 
         results.append({
-            'url': urljoin(resp.url, '?%d' % k),
+            'url': urljoin(str(resp.url), '?%d' % k),
             'title': from_result.text_content(),
             'content': '; '.join(to_results)
         })
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -208,8 +208,8 @@ def response(resp):
     results = []
 
     # detect google sorry
-    resp_url = urlparse(resp.url)
-    if resp_url.netloc == 'sorry.google.com' or resp_url.path == '/sorry/IndexRedirect':
+    resp_url = urlparse(str(resp.url))
+    if resp.url == 'sorry.google.com' or resp_url.path == '/sorry/IndexRedirect':
         raise RuntimeWarning('sorry.google.com')
 
     if resp_url.path.startswith('/sorry'):
--- a/searx/engines/spotify.py
+++ b/searx/engines/spotify.py
@@ -12,7 +12,7 @@
 
 from json import loads
 from urllib.parse import urlencode
-import requests
+import searx.poolrequests as requests
 import base64
 
 # engine dependent config
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -1,13 +1,16 @@
-import sys
 from time import time
 from itertools import cycle
-from threading import RLock, local
+import threading
+import concurrent.futures
+import asyncio
+import logging
 
-import requests
+import uvloop
+import httpcore
+import httpx
 
 from searx import settings
 from searx import logger
-
 
 logger = logger.getChild('poolrequests')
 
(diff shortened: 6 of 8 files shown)


unixfox:
How do I test it for the Google engine? It seems that it is doing by default in HTTP1 only and I don't see any settings to force the Google engine to HTTP2.

dalf:
This line
```python
client = httpx.AsyncClient(http2=True, pool_limits=user_pool_limits, verify=verify, proxies=proxies)
```

See https://www.python-httpx.org/http2/

You can remove 
```python
    for logger_name in ('httpx.client', 'httpx.config', 'hpack.hpack', 'hpack.table',
                        'httpx.dispatch.connection_pool', 'httpx.dispatch.connection',
                        'httpx.dispatch.http2', 'httpx.dispatch.http11'):
        logging.getLogger(logger_name).setLevel(logging.WARNING)
```
to have the httpx logs.

unixfox:
I don't really understand. Is it already enabled here: https://github.com/dalf/searx/blob/3923b87bc3baea74df3d0083b3d4e960878e5372/searx/poolrequests.py#L40?
So every engine should have its requests made in HTTP2 by default?

dalf:
Same behavior than a browser:
* if a website supports only HTTP/1, then httpx uses only HTTP/1.
* if a website supports HTTP/2 then httpx uses HTTP/2.

This applies to all the engines, actually to all outgoing connections. Is it a problem ?

unixfox:
How can I check that it is really doing HTTP2 requests to Google?
I tried your `httpx_experimental` branch and after a few minutes I got the reCAPTCHA. Then I checked if I were blocked in HTTP2 with CURL and I wasn't so I concluded that httpx was still doing HTTP1 requests by default.

dalf:
I see that httpx uses HTTP/2, after that it is a guess game. Perhaps curl is allowed for a time or a # of requests because it is loud and clear that it is curl. Google may understand that searx is not Firefox. What Google can use:
* the HTTP headers:
```
Host: www.google.com
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:73.0) Gecko/20100101 Firefox/73.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Accept-Language: fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3
Accept-Encoding: gzip, deflate, br
DNT: 1
Connection: keep-alive
Upgrade-Insecure-Requests: 1
Cache-Control: max-age=0
TE: Trailers
```
* the TLS version must be 1.3 (related to the OpenSSL version). 
* the client ciphers (see https://clienttest.ssllabs.com:8443/ssltest/viewMyClient.html ):
   * requests can change the cipher list: see outdated PR https://github.com/asciimoo/searx/pull/931/files#diff-e5b583989f4dda8a68e9aaad1a79705bR19
   * httpx most probably can: see https://github.com/encode/httpx/blob/82dc6f32f864d26677df8fc61249a7fc74c687e8/httpx/_config.py#L106
* the way HTTP/2 is managed.
* the fact that despite a "Firefox" connection, exception the ```/?q=...``` no other URLs are downloaded.

Can you tell if the updated PR change something ?

If it doesn't work, what can be improved:
* HTTP header are easy to do.
* the TLS version should be okay (searx-stats2 uses httpx, and use TLS 1.3 )
* Client cipher list: I need some investigation.
* HTTP/2 protocol: no go.
* download other URLs: I remember to have seen "interesting" URLs that could be fetched (not sure it is doable without JS). Most probably a dead end, but in last resort it can be tried. I include cookie storage in loop (on searx side).

unixfox:
Thank you @dalf. It seems that Google is blocking not based on the version of HTTP because now it's blocked in HTTP2 but not in HTTP1 no matter what the user agent is set to.
I think Google has different servers for HTTP1, HTTP2 and HTTP3(QUIC) so that's why Searx gets blocked in a specific HTTP version but not in another HTTP version. (All of this is kinda strange...)
About the HTTP headers I think it's something that can be tuned inside the google.py file engine. The other solutions are probably too difficult to implement into Searx and I don't think Google really check the client cipher list it's too unstable as a way to detect if it's really a real browser.
Anyway it's getting a bit off-topic. I'll investigate more and if I get something interesting I'll post it on #729.

dalf:
See:
* https://github.com/dalf/searx/tree/httpx_firefox
* https://github.com/dalf/searx/commit/b387b0655e697a438298ea7020fc9f2078781040

httpx is configured to look like Firefox.
Basically it fails at every steps.
See the comments inside the code for the details.

Nevertheless it can be improvement.

dalf:
[out of topic, but still about the HTTP client]

I've tried to use the golang HTTP client using:
* golang compile to wasm: see https://withblue.ink/2020/10/03/go-webassembly-http-requests-and-promises.html
* [wasmer-python](https://github.com/wasmerio/wasmer-python)

Show stopper: https://github.com/wasmerio/wasmer/issues/924 (also see http://macias.info/entry/202003151900_go_wasm_js.md )

PR #1861, opened 2020-02-23
fix serious bugs of the test procedure

return42:
The first commit does some pylint and code formatting, the second patch fixes the robot test ..

```
$ make test
...
File "/share/test123/searx-asciimoo/tests/robot/__init__.py", line 17, in test_404
    assert browser.is_text_present('Page not found')
```

**update**

Patch 81f106e in this series fixes a serious bug of the test procedure, we should merge this ASAP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,11 @@ $(GH_PAGES)::
 
 PHONY += test test.pylint test.pep8 test.unit test.robot
 
+test: test.pylint test.pep8 test.unit test.robot
+
 # TODO: balance linting with pylint
-test: test.pep8 test.unit test.robot
-	- make pylint
+test.pylint: pylint-exe
+	$(call cmd,pylint,searx/preferences.py)
 
 test.pep8: pyenvinstall
 	$(PY_ENV_ACT); ./manage.sh pep8_check
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ PHONY += test test.pylint test.pep8 test.unit test.robot
 test: test.pylint test.pep8 test.unit test.robot
 
 # TODO: balance linting with pylint
-test.pylint: pylint-exe
+test.pylint: pyenvinstall
 	$(call cmd,pylint,searx/preferences.py)
 	$(call cmd,pylint,searx/testing.py)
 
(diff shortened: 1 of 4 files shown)


return42:
Patch 81f106e in this series fixes a serious bug of the test procedure, we should merge this ASAP

@dalf, do you have time for a review?

dalf:
I will in about 4h.

return42:
> I will in about 4h.

don't hurry .. no one will die if you do it in the next days

dalf:
> don't hurry .. no one will die if you do it in the next days

Thanks for the reminder :-)

dalf (approved):
All good !

(Now sometimes, I understand why I had a Flask server out of nowhere)

Just some more comments about my review:
* I have check with and without SEARX_DEBUG, and saw the remaining server.
* About the preferences: I'm not how to test that, but ```make test``` does run the new code.
* ```make test.pylint``` does not return issues

return42:
> (Now sometimes, I understand why I had a Flask server out of nowhere)

:) same here  .. I wondered sometimes why my changes does not have any effect ..

> Just some more comments about my review:
>    I have check with and without SEARX_DEBUG, and saw the remaining server.

Thanks for verifying

>   About the preferences: I'm not how to test that, but make test does run the new code.

The problem here was, that `settings['ui']['default_locale']` caused an exception, when default_locale was unset in the settings.yml file .. and it was/is unset in the settings_robot.yml file. That caused an Exception every time the server starts with settings_robot.yml.

I changed `settings['ui']['default_locale']` into `settings['ui'].get('default_locale','')`. BTW I did the same for autocomplete, safe_search [and more](https://github.com/asciimoo/searx/pull/1861/commits/1cae4fa88da1cfad6f953b0ebc170946f3af6c89#diff-9299c479a8d7cc9c9fdf1d1dc15b0933)

>  make test.pylint does not return issues

Correct I changed the Makefile, the verbose output of the pylint test over all was not very helpful. Now only selected files (those which have been reworked to pass the lint test) are linted: https://github.com/asciimoo/searx/blob/6a3ef5561ba48e287f0b9c03a0b6d2f13b703077/Makefile#L78-L80

My consideration is, to do a pylint and SPDX-tag on every file I need to patch due to a bugfix. With we can successively increase the code quality and find hidden bugs (pylint is very good in finding bugs which are not immediately apparent).  The preference file was a bad example, since it has a lot of generic types. Anyway over the long run, pylint will help us a lot.

PR #1863, opened 2020-02-24
Monkey patch babel get_translations to support Occitan

### LINKED ISSUE #1143: Langue missing [Occitan]
Hello,
the Occitan is missing, I saw another issue that explains it's because it's not part of another project.
But things have changed and now the Occitan is part of it, and well documented I have to say.
Can the troubadours's language be included now?
Best regards !

MarcAbonce:
This would close #1143 if merged.

I recently noticed that a Flask-Babel just released their 1.0.0 version. Besides, another PR was merged over there that conflicts and breaks my patch over there, so I'll have to re-work my fix over there anyway. Because of those two things, I don't expect my PR to be merged any time soon.

I know this is an ugly patch, but the other alternatives would be to fork Flask-Babel (which I think would a worse idea) or to remove Occitan from the list of locales.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -45,6 +45,7 @@ try:
     from cgi import escape
 except:
     from html import escape
+from six import next
 from datetime import datetime, timedelta
 from time import time
 from werkzeug.middleware.proxy_fix import ProxyFix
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks for keeping an eye on this!

PR #1893, opened 2020-03-21
Update manage.sh

rasin-lau:
"$(command -V "$ACTION" | grep ' function$')" = ""  will be judged incorrectly by different system languages.


return42:
Hi @rasin-lau, thanks a lot for your patch.  For me it is not clear if its really needed to merge ..

>  will be judged incorrectly by different system languages.

what do you mean by "different system languages"?  What I have seen; it won't work with bash, but therefore there is the shebang `#!/bin/sh`, meaning this is not a bash script, this is a sh script.

Do you know (other) use cases where it fails?  If not, I think we should not merge this PR. BTW: in long term we will reorganize our shell scripts anyway.

return42:
One hint fore the future; In order to submit a patch, please follow the steps at https://asciimoo.github.io/searx/dev/contribution_guide.html#code and choose meaningfull commit messages. I am sorry to say: "Update manage.sh" is complete meaningless to the reader.

dalf:
@return42 
```command``` is affected by ```C_LANG```, not ```type```:

```sh
$ echo $LANG
fr_FR.UTF-8
$ command -V "test"
test est une primitive du shell
$ type -t test
builtin
$ unset LANG
$ command -V "test"
test is a shell builtin
```

But there is a problem if ```type``` is only supported by bash :
https://unix.stackexchange.com/questions/303121/what-is-the-equivalent-of-type-used-in-bash-and-sh-in-csh

return42:
> command is affected by C_LANG

See my answer here https://github.com/asciimoo/searx/pull/1479#issuecomment-569904632

     LANG=C command -V ..

..needs to be in front of.

return42:
continue in #1479 ..

PR #1900, opened 2020-03-25
export build variables / support brands

return42:
We have some variables in the build environment which are also needed in several
other process environments:

- jinja templating: see 04c68740
- sphinx-doc: see e84fc943
- setup.py: see f6b29f68
- docker: see 3dbade0a
- grunt: see 757ebb5
- URL /config see 4d648282

Additional a ISSUE_URL is implemented 33814719


return42:
@dalf: if you look at this patch series, do you have a clue how we can add the environment to the ./Docker file?  / thanks!

dalf:
Currently the docker image is built using [travis](https://github.com/asciimoo/searx/blob/609ac5795ab16ab8a80a89feb54a9e23d4dec725/.travis.yml#L47):

```
./manage.sh docker_build
```

Note about the current travis configuration: the [DOCKER_USERNAME and DOCKER_PASSWORD variables](https://github.com/asciimoo/searx/blob/609ac5795ab16ab8a80a89feb54a9e23d4dec725/.travis.yml#L46) are defined inside the [searx/searx project](https://travis-ci.org/github/searx/searx) not inside the [asciimoo/searx project](https://travis-ci.org/github/asciimoo/searx).

So I guess the way to go is:
* to move the code from [manage.sh](https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/manage.sh#L168-L234) to Makefile, so ```make docker_build``` replaces ```./manage.sh docker_build```
* in the Makefile docker_build target:
   * [to build the docker image](https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/manage.sh#L220-L228) providing these new variables.
   * Note: the docker image name is [github user]/searx. See https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/manage.sh#L215-L216
I'm not sure what to do about that: most probably replace ```git remote get-url origin``` by ```$GIT_URL```

* to add ```ARG GIT_URL = https://github.com/asciimoo/searx``` at the top the Dockerfile.

I will prepare a patch over your branch.

----

About grunt: same [the code from manage.sh](https://github.com/asciimoo/searx/blob/master/manage.sh#L117-L166) can move to Makefile, and then use [the environment variables](https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_env).

Makefile
```diff
@@ -87,6 +88,20 @@ searx.brand:
...
+PHONY += themes themes.oscar themes.simple
+themes: themes.oscar themes.simple
+
+themes.oscar:
+	$(Q)echo '[!] Grunt build : oscar theme'
+	$(Q)grunt --gruntfile  "searx/static/themes/oscar/gruntfile.js"
+
+themes.simple:
+	$(Q)echo '[!] Grunt build : simple theme'
+	$(Q)grunt --gruntfile  "searx/static/themes/simple/gruntfile.js"
```
dalf:
```sh
``` 
is missing. See https://github.com/asciimoo/searx/blob/master/manage.sh#L118-L123

return42: Yes, I dropped it since it was not needed (at least in my tests).  Do you know for what this PATH extension is needed?

manage.sh
```diff
@@ -202,13 +202,13 @@ docker_build() {
...
     # define the docker image name
-    # /!\ HACK to get the user name /!\
-    GITHUB_USER=$(git remote get-url origin | sed 's/.*github\.com\/\([^\/]*\).*/\1/')
+    GITHUB_USER=$(echo "${GIT_URL}" | sed 's/.*github\.com\/\([^\/]*\).*/\1/')
     SEARX_IMAGE_NAME="${GITHUB_USER:-searx}/searx"
 
     # build Docker image
     echo "Building image ${SEARX_IMAGE_NAME}:${SEARX_GIT_VERSION}"
     sudo docker build \
+         --build-arg GIT_URL="${GIT_URL}" \
```
dalf, line 152: Few lines below, I forgot to mention 
```
--build-arg LABEL_VCS_URL=$(git remote get-url origin)
```
-->
```
--build-arg LABEL_VCS_URL="${GIT_URL}"
```

But this change can LABEL_VCS_REF inconsistent with LABEL_VCS_URL.

return42: sorry .. I should have seen it .. i will fix it

Makefile
dalf: grunt and some other npm packages are required.

```./manage.sh npm_packages``` installs these requirements locally, also it installs required dependencies for the simple and oscar theme:
https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/manage.sh#L122-L136

Then the [npm_path_setup](https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/manage.sh#L117-L120) function setups the PATH to use the local grunt install.

Not all user will have grunt. Actually using the grunt from the Linux distribution can be a problem since a different grunt version can behave differently with the same gruntfile.js

**Comment by dalf** [comment]
The other thing that could be helpful for searx-stats2 is to expose the brand variable in the [/config URL](https://github.com/asciimoo/searx/blob/da56bda26542de5697691ff80dac6df876fb366b/searx/webapp.py#L890).

Something like 
```python
                    'version': VERSION_STRING,
                    'brand': {
                           'GIT_URL': brand.GIT_URL,
                           'DOCS_URL': brand.DOCS_URL  # bonus
                    },
```

searx/templates/oscar/base.html
```diff
@@ -85,10 +85,10 @@
             {% endblock %}
             <p class="text-muted">
                 <small>
-                    {{ _('Powered by') }} <a href="https://asciimoo.github.io/searx/">searx</a> - {{ searx_version }} - {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
-                    <a href="https://github.com/asciimoo/searx">{{ _('Source code') }}</a> |
-                    <a href="https://github.com/asciimoo/searx/issues">{{ _('Issue tracker') }}</a> |
-                    <a href="https://searx.space/">{{ _('Public instances') }}</a>
+                    {{ _('Powered by') }} <a href="{{ brand.DOCS_URL }}">searx</a> - {{ searx_version }} - {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
+                    <a href="{{ brand.GIT_URL }}">{{ _('Source code') }}</a> |
+                    <a href="{{ brand.GIT_URL }}/issues">{{ _('Issue tracker') }}</a> |
```
dalf: It supposes that the URL will always follow the same schema, even if github is not hosting the repository.

I suggest to add an ISSUE_URL variable. Perhaps far-fetched.

return42: As longer I think about I came to the conclusion that the issue tracker URL should be hard coded in the template.

return42: I mean .. if it is a brand, the issue tracker should be the same

dalf: Example:
* GIT_URL=https://framagit.org/framasoft/searx
* ISSUE_URL=https://framagit.org/framasoft/searx/-/issues 

There is an additional ```-/``` in the URL, most probably for any gitlab repository.

[EDIT] ```ISSUE_URL=${GIT_URL}/issues``` works for all the github repositories, and it can be change otherwise.

return42: OK .. done

return42:
> The other thing that could be helpful for searx-stats2 ..

Additional a ISSUE_URL is implemented 3381471

dalf:
The remaing thing is .travis.yml, and because the variables are declared in Makefile we have to use make in .travis.yml (no problem, just we have to do it right now).

* https://github.com/return42/searx/blob/issue-1898/.travis.yml#L47
  ```./manage.sh docker_build push``` should be ```make docker.push``` (or ```make docker``` with an option)

*
```
install:
  - ./manage.sh install_geckodriver ~/drivers
  - export PATH=~/drivers:$PATH
  - ./manage.sh npm_packages
  - ./manage.sh update_dev_packages
  - pip install codecov
script:
  - ./manage.sh styles
  - ./manage.sh grunt_build
  - ./manage.sh tests
after_success:
  - ./manage.sh py_test_coverage
  - codecov
```

-->

```
install:
  - ./manage.sh install_geckodriver ~/drivers  # no Makefile version ?
  - export PATH=~/drivers:$PATH
  - make node.env
  - ./manage.sh update_dev_packages	# no Makefile version ?
  - pip install codecov
script:
  - make themes	 # should include "./manage.sh styles" perhaps the usual Makefile way, one entry per .less file ?
  - make test
after_success:
  - ./manage.sh py_test_coverage
  - codecov
```

---

Another way to avoid a big bang:
* create a new file like 'brand.sh' with the variables
* include this file in Makefile
* include this file in manage.sh

MarcAbonce:
Just a little nitpicking here, but I'm not sure if the About page should point to the fork. The way that page is written implies that we're talking about searx itself, for example, by calling searx by it's own name rather than using the instance's name and also by asking people to contribute to our instance list, translations, etc. And to me it makes sense that the about page should stay as a place to point people to the upstream project, imo.

return42:
> Just a little nitpicking here

It is not nitpicking, its important point and my thoughts about are not clear at the moment.

> And to me it makes sense that the about page should stay as a place to point people to the upstream project, imo.

Not only your opinion :) .. I often have a doubt, what a "brand" is.  Therefore I am looking for the right terms and I think we should differentiate between what is "brand" and what is a "spin-off" (both are forks). Beside issue tracker, about page, there are other predicates indicating what a spin-off is, e.g the author or maintainer and properties in the builded package.

I started "brands" to support that a fork might have its own design and some special function which might bee reasonable a in special context.

But I also have the impression that the thoughtless replacement in parts has gone too far. 

I will do  a review under this aspect. If you see I am fail, please put your finger in the wound and point it out to me. Thank you for raising this point!

dalf:
> But I also have the impression that the thoughtless replacement in parts has gone too far.

👍 

Some "technical" ideas if $GIT_URL is different from https://github.com/asciimoo/searx :
* Looking at the footer of https://spot.ecloud.global/ , the searx footer can be something like:
_[$fork_name](//$GIT_URL) forked from [Searx](https://github.com/asciimoo/searx)_
* The about page:
_"About searx"_ can be _"About {{fork_name}}"_
After _"{{fork_name}} is a metasearch engine, aggregating ..."_ add _"{{fork_name}} is fork of [Searx](https://github.com/asciimoo/searx)"_
* _"See more on github"_ can be _"See the details in the [source code](//$GIT_URL)"_
* _"More about searx"_ : keep that section
* _"More about {{fork_name}}"_: new section before _"More about searx"_. Free list of links defined in the Makefile

So it makes clear that:
* the instance is a fork of searx.
* searx reference doesn't disappear.

return42:
If I do understand you right, we have a conditional text in the about page  ["is fork foo from searx|is searx].. I not really like.

ATM I implementing a patch which is similar to what you said about the "about" page .. I will push next time ..  may we discuss after by example.

return42:
@dalf & @MarcAbonce in patch 3486613 I try to find the right choices. In the commit message you can read about the thoughts behind.

What do you think .. did I hit the nail or do I now have a bloody thumb?

searx/templates/__common__/about.html
```diff
@@ -22,15 +22,14 @@ <h2>Technical details - How does it work?</h2>
 
 <h2>How can I make it my own?</h2>
 
-<p>Searx appreciates your concern regarding logs, so take the <a href="{{ brand.GIT_URL }}">code</a> and run it yourself! <br />Add your Searx to this <a href="{{ brand.SEARX_URL }}">list</a> to help other people reclaim their privacy and make the Internet freer!
+<p>Searx appreciates your concern regarding logs, so take the <a href="https://github.com/asciimoo/searx">code</a> and run it yourself! <br />Add your Searx to this <a href="{{ brand.PUBLIC_INSTANCES }}">list</a> to help other people reclaim their privacy and make the Internet freer!
```
dalf: At least a fork caches the results using redis:
https://gitlab.e.foundation/e/cloud/my-spot/-/blob/master/searx/poolrequests.py#L166

A link to the original source code can be misleading.

return42: Good point / thanks!

searx/templates/__common__/about.html
```diff
@@ -48,13 +47,13 @@ <h2 id="dev_faq">Developer FAQ</h2>
 
 <h3>New engines?</h3>
 <ul>
-    <li>Edit your <a href="https://raw.github.com/asciimoo/searx/master/searx/settings.yml">settings.yml</a></li>
```
dalf, line 51: May be wrong: 
* here it is about searx.
* bellow ```{{ brand.DOCS_URL }}/admin/installation.html#uwsgi``` is about the fork.
* I've just suggested to make a link to the fork git repository instead of the searx git repository.

I think it can be confusing.

[EDIT] I'm not sure to know what is the right solution.

[EDIT2] Another idea: two pages, one only searx, another one about the fork if it is fork.

return42: > [EDIT] I'm not sure to know what is the right solution.

me too :) .. my thought is, that settings.yml and 'custom engine' are just examples showing how it looks like.  Those should link to searx .. 

This "Installation/WSGI support" should be dropped .. this is nothing one want see on a about page.

What do you think?

return42: May we should also drop "New engines?" and "How to debug" ..

return42: > EDIT2] Another idea: two pages, one only searx, another one about the fork if it is fork.

I am not happy with .. lets drop all what is not in the scope of a common user .

An "about" page should only address higher-level issues that are important to each user. No developer aspects.

searx/templates/__common__/about.html
```diff
@@ -1,62 +1,96 @@
...
+
+  <p>
+    <a href="{{ url_for('stats') }}">Stats page</a> contains some useful data about the engines used.
+  </p>
+
+  <h2>How can I make it my own?</h2>
+
+  <p>
+    Searx appreciates your concern regarding logs, so take the
+    code from the <a href="https://github.com/asciimoo/searx">orgin searx projekct</a> and
```
dalf: typo: project

searx/templates/__common__/about.html
```diff
@@ -1,63 +1,96 @@
...
+  <h2 id='add to browser'>Can it be set as the default search engine?</h2>
+
+  <dt>Firefox</dt>
+
+  <dd>
+    <a href="#" onclick="window.external.AddSearchProvider(window.location.protocol + '//' + window.location.host + '{{ url_for('opensearch') }}');">Install</a>
+    searx as a search engine on any version of Firefox! (javascript required)
+  </dd>
+
+  <h2>How to debug engines?</h2>
```
dalf: I've never use this page as debug information.

Suggestion: _"Where to find anonymous usage statistics of this instance ?"_

dalf:
Commit 5d087ee :
* I like the "Where are the docs & code of this instance?" at the end, it makes the thing clear.
* I would put the "Why use searx?" section before the "Technical details - How does it work?" section. A matter of taste.

Otherwise, LGTM 👍

return42:
@dalf I finished now .. thanks a lot for your help! 

Don't wonder I also put 00251d7 in this series, its a bit off topic here but fits to the 'make project' aspect.

dalf:
> Don't wonder I also put 00251d7 in this series, its a bit off topic here but fits to the 'make project' aspect.

Let's focus on one topic IHMO.

There is still .travis.yml. This line does not work:
https://github.com/asciimoo/searx/blob/971752198156e6a2a4ade127032d35c54ea3563b/.travis.yml#L20

```make themes``` can replace it, but it creates a mix of manage.sh calls and make calls.

File: .gitignore [hunk 8]
```diff
@@ -15,7 +15,8 @@ setup.cfg
 */*.pyc
 *~
 
-node_modules/
+/node_modules
+/package-lock.json
```
dalf: Note: package-lock.json should be in the git repository.

See
* https://stackoverflow.com/questions/44297803/what-is-the-role-of-the-package-lock-json
* https://docs.npmjs.com/files/package-lock.json

For now, there is no package-lock.json in the respository, so package-lock.json in .gitignore makes ```git add``` easier.

return42: You mean, you want to see them as "unversioned" files .. right?

return42: If so, should I also remove them from the "clean" target https://github.com/asciimoo/searx/pull/1900/files#diff-b67911656ef5d18c4ae36cb6741b7965R112-R114

dalf: The target is to have them in the repository, but no big bang : searx exists without them since the beginning.

Yes unversioned file and ```make node.clean``` can remove them for now.

return42: OK

> The target is to have them in the repository,

I know these concepts from other package managers but I have a doubt that it make sense to fix all requirements .. in practice no one takes a regularly look at all the vulnerabilities and the software is build all time long with all the vulnerabilities .. but here is not the right place to discuss this aspect  :)

return42:
@dalf I have a question about travis and the robot test. 

In travis https://github.com/asciimoo/searx/blob/master/.travis.yml#L13 a gecko driver is installed in the `install:` section.  

In the `script:` section the `tests()` script (function) is called, see https://github.com/asciimoo/searx/blob/master/.travis.yml#L21

And if I am not wrong, in the `tests()` function another gecko driver is installed, see https://github.com/asciimoo/searx/blob/master/manage.sh#L107

What do you think ..I am wrong?

dalf:
About this one:
```
  - ./manage.sh install_geckodriver ~/drivers
  - export PATH=~/drivers:$PATH
```
It installs geckodriver in ~/drivers

You are right ```./manage.sh tests``` check geckodriver again, and install it if it is not in the PATH.
But without a path (```~/drivers```), the script installs geckodriver in the virtualenv : https://github.com/asciimoo/searx/blob/master/manage.sh#L49-L59

Since .travis.yml do not use virtualenv, if we remove ```./manage.sh install_geckodriver ~/drivers```, ```./manage.sh tests``` will fail to install geckodriver.

If we remove manage.sh and use only Makefile, this behavior can be removed.

return42:
> Since .travis.yml do not use virtualenv, ..

I know, it is not needed to build up a virtualenv in travis, since it is all done in a container.  But was it omitted only for performance reason?

My target is, that both run exactly the same steps, the developer (interactive) and the travis job.

If the the virtualenv was omitted only for (minimal) performance reason, I vote to add the same virtualenv to the travis job.

dalf:
> But was it omitted only for performance reason?

I've looked at the [.travis.yml history](https://github.com/asciimoo/searx/commits/master/.travis.yml), but I can't remember why.

I don't think there is a problem to use a virtualenv in .travis.yml

return42:
@dalf the (new) travis build seems to work, could you doublecheck?

The `make docker` target spits out some `SyntaxError` I do not know the reasons for .. do you have any idea what is going wrong?

I will continue tomorrow .. up to here, first of all thank you for your support.

dalf:
/!\ Warning: unfinished solution.

With this [.dockerignore](https://gist.github.com/dalf/b596f1413f3028fff0c3962975d96ee0) there is no more error. Explanation: 
* the python files are compiled while building the docker image
* a node modules contains some python files
* the python files inside the node module doesn't compile because... I don't know

It raises the fact that node_modules were included in the docker image which should not happen.

**[EDIT]** : same the ```local``` directory was included. Dockerfile builds searx in its own way (without virtualenv) :
* It works.
* but it should use Makefile (virtualenv and all).
* I really wish to avoid big bangs in the PR.

Makefile
```diff
@@ -71,35 +74,138 @@ $(GH_PAGES)::
...
+# ------------
+
+PHONY += styles style.legacy style.courgette style.pixart style.bootstrap
+styles: style.legacy style.courgette style.pixart style.bootstrap
+
+quiet_cmd_lessc = STYLE     $3
+      cmd_lessc = PATH="$$(npm bin):$$PATH" \
+	lessc --clean-css="--s1 --advanced --compatibility=ie9" "searx/static/$2" "searx/static/$3"
+
+style.legacy:
```
dalf: ```style.legacy``` could be renamed ```themes.legacy``` : it actually builds the files for the legacy theme, then  ```themes.legacy``` can be a dependency of ```themes```.

Same for the other styles.*

Debatable: about ```style.bootstrap```, same convention : ```theme.bootstrap``` (even it is more a toolbox for the oscar theme).

So there is no need to add the missing ```make styles``` in the help target.

return42:
> I really wish to avoid big bangs in the PR.

from my POV, no need to change .. it works.

Thanks for your suggestion (ec453e0, 8a16903 ) ...very helpful!

dalf:
The current recipe to create a new brand:
* modify Makefile variables
* call ```make searx.brand```

Not sure if ```searx.brand``` should be dependency of ```docker``` target and some other targets ? (since ```searx.brand```  is really fast, and brand.env will disapear sooner or later)

---

Scenario:
* someone forks and modifies searx
* forget to modify the variables.

Everything will work with the original brand even if the git repository is somewhere else.

A (broken) patch to ```make docker``` to avoid this:
```sh
docker_build() {
...
GIT_REMOTE_ORIGIN_URL = $(git remote get-url origin 2> /dev/null)

if [ ! ${GIT_REMOTE_ORIGIN_URL} ]; then
	echo "there is no remote origin"
	exit 1
fi

if [ "${GIT_REMOTE_ORIGIN_URL}" != "${GIT_URL}" ]; then
	# won't work if origin is git@github.com:asciimoo/searx.git
	echo "GIT_URL ( ${GIT_REMOTE_ORIGIN_URL} ) doesn't match the git remote origin ( ${GIT_URL} )"
	exit 1
fi
...
}
```

I'm not sure there is a easy fix to this, except documentation.

return42:
> Not sure if searx.brand should be dependency of docker target and some other targets ? (since searx.brand is really fast,

good point, see 1ae3978

> and brand.env will disapear sooner or later)

We need the build environment in every tool of our toolchain and bash is a part of our toolchain. E.g. the scripts in PR #1803 pick up this concept ([patch](https://github.com/asciimoo/searx/pull/1803/commits/7b4cf2eb489d8f6c95ccac5af180254b78e7c460))

> I'm not sure there is a easy fix to this, except documentation.

Are your scenarios covered with patch 1ae3978 ?

About docs .. PR #1803 touches a lot of the documentation, I want to see merged first, before documenting the buildenv in detail.

dalf:
> We need the build environment in every tool of our toolchain and bash is a part of our toolchain.

Yes my idea, may be not shared is to remove manage.sh and create different scripts called by Makefile:
```
utils/
	fabfile.py
	standalone_searx.py
	google_search.py
build.tools/
	fetch_currencies.py
	fetch_firefox_version.py
	fetch_languages.py
	makefile.include
	makefile.python
	makefile.sphinx
	update-translations.sh
	docker_build.sh  # from manage.sh
	install_geckodriver.sh # from manage.sh
```

> Are your scenarios covered with patch 1ae3978 ?

Someone can fork, modify, publish without changing GIT_URL, even if ```$(git remote get-url origin)``` gives another URL than GIT_URL. But the public git URL can be different from the private git URL (https://github.com/asciimoo/searx.git and git@github.com:asciimoo/searx.git).

---

One last thing: in .travis.yml, I'm not to sure to understand why npm and pip have been removed from the cache section.

Otherwise, LGTM.

return42:
> Yes my idea, may be not shared is to remove manage.sh and create different scripts called by Makefile:

This is also my long term aim.  PR #1803 brings all the boilerplate scripting useful to do so.  But first,lets finish this PR and merge #1803 ..

 > Someone can fork, modify, publish

I am a bit confused .. help me: do we need to handle this? .. compared to what we have, is it a regression?

> One last thing: in .travis.yml, I'm not to sure to understand why npm and pip

npm is by default in the cache (I have read somewhere in the travis docs) and the pip has changed to the pip in folder ./local (virtualenv pip is used, before OS pip was used).

dalf:
> I am a bit confused .. help me: do we need to handle this? .. compared to what we have, is it a regression?

Once this PR is merged, nothing stops me to do that :
* update my fork: https://github.com/dalf/searx
* make some modifications.
* update https://hub.docker.com/r/dalf/searx but I don't change GIT_URL in Makefile.

--> the URLs and docker image labels are wrong.

Even if I run searx directly, the URL will be wrong.

On top of that, some people will want:
* to git clone searx, make some modification without a dedicated repository.
* to download searx directly without git.

I don't have a solution to that:
* documentation can help, but except some people to skip this part.
* display a warning if GIT_URL is different from ```$(git remote get-url origin)```, but it is not reliable.

We can leave it as it is in my opinion.

return42:
> Even if I run searx directly, the URL will be wrong.

This is a problem I fight with ..  ATM the default is:

    export SEARX_URL=https://searx.me

We can change this to:

    export SEARX_URL=https://$(shell hostname)/searx

But I cant say what it does with _buildhosts_, docker or travis build.

> We can leave it as it is in my opinion.

If you think the `https://$(shell hostname)/searx` helps, I can apply the patch, otherwise I would say we leave it as is .. for the first.

dalf:
> If you think the https://$(shell hostname)/searx helps, I can apply the patch, otherwise I would say we leave it as is .. for the first.

I prefer to leave it as it is: ```hostname``` can be different from the public hostname.

[EDIT] I was more focusing on GIT_URL.

For now, where SEARX_URL is used ?

return42:
> For now, where SEARX_URL is used ?

ATM there is only one usage: `./docs/conf.py\040:extlinks['search'] = (SEARX_URL + '/%s', '#')` 

SEARX_URL is the public URL of the instance.  Scripts in #1803 will need to know the public URL and make more use of this build environment.

> I prefer to leave it as it is:

OK, if you think its OK, please give me your _reviewed_ / thanks for all the help!

dalf (approved):

dalf:
Note without solution: the travis build fails with the docker step: https://travis-ci.org/github/searx/searx/jobs/668856928

https://travis-ci.org/github/searx/searx/jobs/668856928#L495 :
```denied: requested access to the resource is denied```.

The docker image is ```asciimoo/searx:0.16.0-89-949378a0``` as this line says:
```sh
 GITHUB_USER=$(echo "${GIT_URL}" | sed 's/.*github\.com\/\([^\/]*\).*/\1/')
```

But it should be ```searx/searx:0.16.0-89-949378a0``` to keep the previous configuration.

One way to solve this issue, is to revert the change:
``` 
 GITHUB_USER=$(git remote get-url origin | sed 's/.*github\.com\/\([^\/]*\).*/\1/')
```

---

Note: https://travis-ci.org/github/searx/searx/jobs/668856928#L202
```
/home/travis/.travis/functions: line 109: local/py3/bin/activate: No such file or directory
```
the error doesn't seems to disturb the build.

searx/templates/__common__/about.html
```diff
@@ -1,63 +1,97 @@
...
+
+  <p>
+    <a href="{{ url_for('stats') }}">Stats page</a> contains some useful data about the engines used.
+  </p>
+
+  <h2>How can I make it my own?</h2>
+
+  <p>
+    Searx appreciates your concern regarding logs, so take the
+    code from the <a href="https://github.com/asciimoo/searx">orgin searx project</a> and
```
MarcAbonce, line 79: Hey, I know I'm late, but I just realized that there's a small typo here.

return42: Fixed in #1906

PR #1905, opened 2020-03-31
[fix] travis: fix docker build.

dalf:
Since #1900 docker images are published on asciimoo/searx name.
Searx docker image are published on searx/searx name: GIT_URL in Makefile need to be updated.

This PR:
* add a new Makefile target: ```docker.push```, a wrapper for ```./manage.sh docker_build push```
* call ```make -e GIT_URL=$(git remote get-url origin) docker.push``` so GIT_URL has the right value even in a fork.

You can see a successful build here:
* https://travis-ci.org/github/dalf/searx/jobs/669188184
* https://hub.docker.com/r/dalf/searx/tags

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
       install: true
       script:
         - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-        - local/py3/bin/activate; ./manage.sh docker_build push
+        - make -e GIT_URL=$(git remote get-url origin) docker.push
       after_success: true
 
 notifications:
--- a/Makefile
+++ b/Makefile
@@ -169,6 +169,9 @@ PHONY += docker
 docker: buildenv
 	$(Q)./manage.sh docker_build
 
+docker.push: buildenv
+	$(Q)./manage.sh docker_build push
+
 # gecko
 # -----
 


return42 (approved):
Thanks!  

BTW another question:  does the `local/py3/bin/activate: No such file or directory` from https://github.com/asciimoo/searx/pull/1900#issuecomment-606188508 still exists? I can't see it any more in travis log, but my travis knowledge is down by zero.

dalf:
> BTW another question: does the local/py3/bin/activate: No such file or directory

I removed it, ```./manage.sh docker_build``` doesn't require Python (with a help a [hack](https://github.com/asciimoo/searx/blob/master/manage.sh#L136-L137)).

dalf:
There is still an issue: 
* the image is pushed correctly to [hub.docker.com](https://hub.docker.com/r/searx/searx/tags?page=1&ordering=last_updated)
* but ```make buildenv``` creates two not commited files during the build process because:
  * searx/brand.py and utils/brand.env are already in the git repository  
  * and GIT_URL is different

so [```./manage.sh docker_build```](https://github.com/asciimoo/searx/blob/master/manage.sh#L123-L127) adds ```-dirty``` to the tag name. Current tag : ```0.16.0-94-8189e18d-dirty```.

[EDIT] it was the case with [my repository](https://hub.docker.com/r/dalf/searx/tags), but I haven't seen it.

return42:
I am fighting with travis in my PR https://github.com/asciimoo/searx/pull/1803 .. travis stops with error from line https://travis-ci.org/github/asciimoo/searx/jobs/669564603#L205 

But I can't see any difference to the travis build from master: https://travis-ci.org/github/asciimoo/searx/jobs/669573689#L203

Is there a chance to test travis locally or did you have any other tips for me?

dalf:
> make: ./local/py3/bin/pip: Command not found

Perhaps it should be ./local/py3/bin/pip3

So [this line](https://github.com/asciimoo/searx/blob/master/utils/makefile.python#L139) should be changed.

But doing so we drop support for Python 2.

return42:
> But doing so we drop support for Python 2.

ATM in master the PY2 build is a mix of [Python2 and local/py3](https://travis-ci.org/github/asciimoo/searx/jobs/669572042#L278)

PR  #1910 brings real python 2 support, see commit [env: PY=2](https://github.com/asciimoo/searx/pull/1910/commits/09a4062505a8d79451374139bab525457eeb28cd#diff-354f30a63fb0907d4ad57269548329e3R34) and in the travis log [line 369: /local/py2/bin/python](https://travis-ci.org/github/asciimoo/searx/jobs/669801569#L369)

dalf:
For information: the docker image made from [searx/searx](https://github.com/searx/searx) contains the https://github.com/asciimoo/searx URL since the themes are not rebuilt.

Example: https://a.searx.space/static/themes/simple/js/searx.head.min.js
```javascript
/*! simple/searx.min.js | 06-08-2019 | https://github.com/asciimoo/searx */

(function(t,e){"use strict";var a=e.currentScript||function(){...})(window,document);
//# sourceMappingURL=searx.head.min.js.map
```

So the searx instances using the docker image appear as vanilla on searx.space.

return42:
Should we add target 'project' as prerequisite to the docker targets?

````diff

index abd0eddb..76fb8147 100644
--- a/Makefile
+++ b/Makefile
@@ -166,10 +166,10 @@ themes.bootstrap:
 # ------
 
 PHONY += docker
-docker: buildenv
+docker: buildenv project
        $(Q)./manage.sh docker_build
 
-docker.push: buildenv
+docker.push: buildenv project
        $(Q)./manage.sh docker_build push

````

dalf:
> Should we add target 'project' as prerequisite to the docker targets?

What is a clean docker image: 
* [x] update brand.env / brand.py: done by ```make docker```
* [x] all files are commited otherwise the ```-dirty``` tag is added: done.
* [ ] update theme: _as far I remember_, the problem here is [uglify](https://github.com/asciimoo/searx/blob/master/searx/static/themes/oscar/gruntfile.js#L14-L23) can produces different output files for the same input: not done.

* If uglify (or any tools use to produce the static files) is not problem we can rebuild the themes before building the docker image (even if I think it can lead to issues because the output are commited files).
* If output is inconsistent, then we can't do that.

Anyway, the risk is this scenario:
1. someone forks searx
1. modify the static files, rebuild the theme without modifying the brand OR modify the static files directly.
1. make a docker image
   * if the file are not commited: the ```-dirty``` is added.
   * if the file are commited: 
     * the commit id in the version will be different from this repository (for sure, nobody will notice).
     * the modified static files will keep ```https://github.com/asciimoo/searx``` instead of the fork even if they have been modified.

So to have a safety net:
* it would be good to check the coherency between ```git remote get-url origin``` and ```$GIT_URL```.
* about the theme, it is tricky since the output of grunt is commited.

return42:
Hi @dalf ..

>> Should we add target 'project' as prerequisite to the docker targets?

sorry, my fail .. I wrote "project" but I mean target "buildenv" (like I showed in the diff) .. sorry for the confusion.
 
>  update theme: as far I remember, the problem here is uglify can produces different output files for the same input: not done.

> If output is inconsistent, then we can't do that.

With target [buildenv](https://github.com/asciimoo/searx/blob/master/Makefile#L94) the themes are not touched.

For the docker build Unfortunately, I do not have the time at the moment to familiarize myself sufficiently with the topic. I follow your recommendations. What do you think we need to do besides the "buildenv" target?

dalf:
About docker build: it is more related to what is a release, how to make a fork.

The only point to add in my opinion is the coherency check between the git origin and $GIT_URL. I will make a PR about this (or try to since there is a tricky thing about git://... and https://... URLs)

return42:
> I will make a PR about this (or try to since there is a tricky thing about git://... and https://... URLs)

Thanks! .. then we can continue the discussion under the new PR

PR #1906, opened 2020-03-31
[fix] template: fix typo in about.html

dalf:
see https://github.com/asciimoo/searx/pull/1900#discussion_r400651191


return42 (approved):
Thanks!

PR #1910, opened 2020-04-01
improve travis build

return42:
In this series I will try to improve the travis build a little bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
   firefox: "latest"
 
 install:
+  - env; which python; python --version
   - make V=1 install
   - make V=1 gecko.driver
   - make V=1 node.env
@@ -29,6 +30,8 @@ stages:
 
 jobs:
   include:
+    - python: "2.7"
+      env: PY=2
     - python: "3.5"
     - python: "3.6"
     - stage: docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ language: python
 cache:
   - directories:
     - $HOME/.cache/pip
-
+before_cache:
+    - sudo chown -R travis:travis $HOME/.cache/pip
+    - rm -f $HOME/.cache/pip/log/debug.log
 addons:
   firefox: "latest"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,12 @@ os: linux
 dist: bionic
 arch: arm64
 language: python
-cache: false
+cache:
+  - directories:
+    - $HOME/.cache/pip
+before_cache:
+    - sudo chown -R travis:travis $HOME/.cache/pip
+    - rm -f $HOME/.cache/pip/log/debug.log
 addons:
   firefox: "latest"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ stages:
 
 jobs:
   include:
-    - python: "2.7"
     - python: "3.5"
     - python: "3.6"
     - stage: docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,12 @@ language: python
 cache:
   - directories:
     - $HOME/.cache/pip
-before_cache:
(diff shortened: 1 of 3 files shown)


return42:
Just to left some notes here .. 

I don't know why, but in the travis build there is no pip command in `./local/py3/bin`,  See _commands available at./local/py3/bin_ : https://travis-ci.org/github/asciimoo/searx/jobs/669664131#L146

Before changing `os: linux` in commit 8c145ff there was a pip .. strange.

To be able to do without it, commit 9b48ae4 does `replace: $(PY_ENV_BIN)/pip --> $(PY_ENV_BIN)/python -m pip`

dalf:
For information, doesn't require answers, I'm not sure to understand:
* arch: arm64  # what does it bring ?
* before_cache: ...

return42:
> before_cache: 

See commit message https://github.com/asciimoo/searx/pull/1910/commits/9e32fb27e71ca0ae32a0eead712a0504c9d2d951 .. I had the same "Permission denied: '/home/travis/.cache/pip/...' " issue in PR #1803 and solved it by a change owner in the before_cache phase of the travis build.  I don't know why the same solution does not work in this PR .. travis caching drives me a bit crazy.

return42:
> arch: arm64 # what does it bring 

nothing, I can rewind it.

[EDIT] face palm .. I confused arm with amd .. n.c.

return42:
Even if I [disable cache](https://travis-ci.org/github/asciimoo/searx/jobs/669685945/config?utm_medium=notification&utm_source=github_status), "Permission denied: '/home/travis/.cache/pip/...' "  error pops up. WTH is it.

dalf:
About amd64, there is this error :
```/home/travis/.travis/functions: line 109: /home/travis/firefox-latest/firefox/firefox: cannot execute binary file: Exec format error```

So the tests won't be able to run successfully.

**Comment by return42** [comment]
Strange, now the github GUI hangs at "Some checks haven't completed yet" .. the build itself is done with success https://travis-ci.org/github/asciimoo/searx/pull_requests

**Comment by dalf** [comment]
I've restarted the build on travis. Now github synchronized the status.

**Comment by return42** [comment]
> I will also test the builds step by step, don't merge this until I am asking for / thanks!

@dalf what do you think .. is this PR ready to merge?

File: .travis.yml [hunk 1]
```diff
@@ -27,6 +31,7 @@ stages:
 jobs:
   include:
     - python: "2.7"
+      env: PY=2
```
dalf, line 32: ~~Why / how this variable is used ?~~
to skip the test.pylint target with the Python 2 version.

.travis.yml
```diff
@@ -5,7 +5,9 @@ language: python
 cache:
   - directories:
     - $HOME/.cache/pip
-
```
dalf, line 6: From https://github.com/falconry/falcon/blob/master/.travis.yml#L78-L79 : 
> NOTE(vytas): wheels sometimes fail to build on s390x due to cache being owned by another user.

So, perhaps these two new lines can be removed on the amd64 architecture.

dalf: It works without these two lines:
* https://travis-ci.org/github/dalf/searx/builds/670155742/config
* https://github.com/dalf/searx/blob/return42-rfc-travis/.travis.yml

.travis.yml
return42: You can run make targets with PY=2 to test in or install a python 2 environment:

      make PY=2  install test

If PY= is unset, PY=3 is the default.  To see more options run help

     make help

return42: The PY=2 condition in main Makefile was only a needed addition to skip the pylint test. Mainly the PY variable sets a py2 or py3 environment.

.travis.yml
return42: Thanks a lot for the tested solution .. I applied 41f002b

dalf (approved):
Thank you for dealing with travis !

Note: perhaps squash the commits since there are some trails / errors / fixes.

return42:
Let them in the history ..

Issue #1577, opened 2019-04-22
Add additional Wikimedia projects


nchristensen:
Wikipedia and Wikidata have already been implemented but others may be useful. I have ordered the list below roughly from more to less useful (IMO of course).

Wikimedia Commons: images, video, audio
Wiktionary: definitions
Wikibooks
Wikiquote
Wikivoyage
Wikisource
Wikiversity
Wikinews

PR #1914, opened 2020-04-05
[add] Additional Wikimedia projects

splintersuidman:
I've added some Wikimedia projects engines, using the generic Mediawiki engine. All projects mentioned in #1577 but WIkimedia Commons are added, as I think such an engine needs more features to support images and files.

The added projects:
- Wikibooks
- Wikinews
- Wikiquote
- Wikisource
- Wiktionary
- Wikiversity
- Wikivoyage

Some things that need to be done or discussed:
- [x] Are the shortcuts okay?
- [x] Is the number of results okay?
- [x] Should some or all be enabled by default?
- [x] Are the categories right?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -123,3 +123,4 @@ generally made searx better:
 - Vipul @finn0
 - @CaffeinatedTech
 - Robin Schneider @ypid
+- @splintah
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -694,7 +694,6 @@ engines:
     number_of_results : 5
     search_type : text
     disabled : True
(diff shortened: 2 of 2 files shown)


Venca24:
The shortcuts (as well as the shortcut of current Wikipedia engine) are pretty close to what [is used](https://incubator.wikimedia.org/wiki/Incubator:Wikis) also on Wikimedia Incubator. Therefore I suggest to have all of them in a form of a 2-letter code.

Namely:
* Wikiversity: wvs -> wv
* Wikivoyage: wvy -> wy

kvch (approved):

Issue #363, opened 2015-06-08
Parse !engine or :language anywhere in the search query


haasn:
I often type my search term first and then decide which engine to use, which results in something like “foo bar !wp”.

DDG would parse this identically to “!wp foo bar”, but searx does not and I find it somewhat annoying. It would be cool if you could add an option to allow parsing for engine syntax anywhere in the search term.

vindarel:
Hi devs,
I tried a bit here: https://github.com/vindarel/searx/commit/0bcecd72a6181d72e1afd2c4cb87ea6e96945f91
However it's not working yet and you may help me with a quick review ! (you can try [on my server](http://5.196.70.6:8888/)).

The thing is, I don't quite understand the type and use of `self.query_parts` (looks like we don't put all keywords in it, in the following class methods, why is it always accessed with `self.query_parts[-1]`, why the `-1` ?), and maybe I broke `self.engines` ? Since I don't see much docstring or unit test on `parse_query`, I thought I'd rather ask for a quick review.

Any thoughts then ? Thanks !

qazip:
Is there any news regarding this? I also agree with the OP that one should be able to use "!engine" anywhere in search string..

asciimoo:
added in aa3c18dda9329fff875328f6ba97483c417b149a

PR #1915, opened 2020-04-08
Allow !engine, !language, <timeout anywhere in query (#363).

splintersuidman:
This pull request adds support for specifying engines, languages and timeouts anywhere in the query, as proposed in #363.

The changes work as follows: all raw query parts (separated by whitespace in the query) are divided into _search query parts_ (the query input for engines) and _special query parts_ (engines, languages, etc.). The query `'!go search :en engine'`, for example, contains the search query parts `['search', 'engine']` and the special query parts `['!go', ':en']`.

These changes do, however, change the query behaviour considerably (the tests are updated accordingly):
- Order of query parts is not fully preserved: after the query `'!go search :en engine'` is parsed, `getFullQuery()` returns `!go :en search engine`. I think this is only noticeable when applying suggestions. When searching for `'!go serch :en engine'` (note the misspelling of search), for example, the suggestion `'search engine'` is given, with a link to the query `!go :en search engine`.
- Whitespace is not preserved; this means that all whitespace will be changed into a single space when applying a suggestion: when searching for `'serch\tengine'` and applying the suggestion `'search engine'`, the new query becomes `'search engine'`, without the many spaces. This could be the same as the old behaviour, I'm not sure. (Fixing this would not be very hard, but maybe unnecessary.)
- Incorrect special parts are not removed. When applying the above mentioned suggestion to `'<xxx serch engine'`, the new query will be `'<xxx search engine'`. (This is very easy to fix, if necessary.)

So, the question is whether these behaviour changes are acceptable (and whether allowing bangs anywhere in the search is wanted, of course).

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -40,7 +40,8 @@ class RawTextQuery(object):
         if disabled_engines:
             self.disabled_engines = disabled_engines
 
-        self.query_parts = []
+        self.search_query_parts = []
+        self.special_query_parts = []
         self.engines = []
         self.languages = []
         self.timeout_limit = None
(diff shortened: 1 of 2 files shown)


return42:
In order not to end up in a dead end, we should first consider how best to embed such tools into the search syntax. See my [comment at PR1146](https://github.com/asciimoo/searx/issues/1146#issuecomment-611003952).

dalf:
It was implemented in a similar way with these two commits
* https://github.com/searx/searx/commit/2aef38c3b9d1fe93e9d665a49b10151d63d92392
* https://github.com/searx/searx/commit/aa3c18dda9329fff875328f6ba97483c417b149a

PR #1923, opened 2020-04-15
Remove discontinued faroo engine

HLFH:
Hi,

Commit is to remove discontinued `faroo` engine.

Related to: https://github.com/asciimoo/searx/issues/1582

I confirm that `faroo` engine is gone for good.
`faroo.com` is being redirected to [`seekstorm.com`](https://seekstorm.com/), which is another service that may *come soon* or not.

Thanks,
HLFH


return42 (approved):
Thank you so much .. I am glad to see that there are still contributors who also think about cleaning up.

PR #1934, opened 2020-04-21
Update default user agents

unixfox:
This PR update the default user agents with newer versions of Firefox.

About the `.0` change, it really counts on some engines like Bing which seems to allow `Windows NT 10.0` but not `Windows NT 10` (gotta love Microsoft).

Also, this page confirms that the official user agent use `Windows NT 10.0` and not `Windows NT 10`: https://www.whatismybrowser.com/guides/the-latest-user-agent/firefox

### DIFF
--- a/searx/data/useragents.json
+++ b/searx/data/useragents.json
@@ -1,11 +1,11 @@
 {
     "versions": [
-        "74.0",
-        "73.0.1",
-        "73.0"
+        "75.0",
+        "74.0.1",
+        "74.0"
     ],
     "os": [
-        "Windows NT 10; WOW64",
+        "Windows NT 10.0; WOW64",
         "X11; Linux x86_64"
     ],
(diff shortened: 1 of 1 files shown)


return42 (approved):
> About the .0 change, it really counts on some engines like Bing which seems to allow Windows NT 10.0 but not Windows NT 10 (gotta love Microsoft)

That's also my experience. . . I couldn't believe it ... / thanks for the patch!

dalf:
This one should be updated too: https://github.com/asciimoo/searx/blob/09a4062505a8d79451374139bab525457eeb28cd/utils/fetch_firefox_version.py#L27 (bookmark for myself)

unixfox:
@dalf you are right! Would you mind do it?

return42:
> This one should be updated too ..

thanks for the hint, I send #1936

@unixfox FYI: to update user agent you can use 

    make useragents.update

PR #1938, opened 2020-04-22
 [enhancement] decoupling plugin development from the searx core

return42:
Searx discovers *external* plugins by using package metadata (python entry points). See asciimoo#1768

````
   setup(
       name = 'example',
       # ...
       entry_points = {
           'searx.plugins': [ 'foo_example_plugin = foo.example.plugin', ],
       },
   )
````

update: [Merged](https://github.com/return42/searx/projects/1) into the [searx-next branch](https://github.com/return42/searx/tree/searx-next)

### CODE REVIEW

**Comment by return42** [comment]
To have an example of an external plugin, I implemented the PR https://github.com/asciimoo/searx/pull/1878 as one (`tgwf-searx-plugins`). 

    pip install git+https://github.com/return42/tgwf-searx-plugins

File: searx/webapp.py [hunk 1]
```diff
@@ -39,8 +39,7 @@
     from pygments.formatters import HtmlFormatter
 except:
     logger.critical("cannot import dependency: pygments")
-    from sys import exit
-    exit(1)
+    sys.exit(1)
```
dalf, line 42: Does this code fix / enhance / modify the plugin system ?

We can just remove try...except, but this is another PR in my opinion.

return42: > Does this code fix / enhance / modify the plugin system ?

No, but I fixed it BTW.  The sys modul was already imported some [lines above](https://github.com/asciimoo/searx/blob/65236deb9cc255912267983546a9599ca155e1a3/searx/webapp.py#L29), no need to import the name 'exit' explicit .. or did I misunderstood your question?

dalf: Actually about this one, I don't know why there is try...except..., why not let Python crash ?

I wish there are two PR:
* one with all small fixes --> this is one can be reviewed then merged quickly.
* one about the plugins / themes / resources with a smaller diff --> no need to understand that some changes are not related.

searx/__init__.py
```diff
@@ -1,35 +1,12 @@
...
-try:
-    from yaml import safe_load
-except:
-    from sys import exit, stderr
-    stderr.write('[E] install pyyaml\n')
-    exit(2)
-
-searx_dir = abspath(dirname(__file__))
-engine_dir = dirname(realpath(__file__))
+from yaml import safe_load
```
dalf, line 9: Is it related to the plugins ?

searx/webapp.py
return42: You are right, my fail .. I have to learn not to fix things BTW.  Sorry, for the future I will improve myself. Could we leave it that way in this PR?

searx/__init__.py
return42: Same here .. I fixed it BTW .. It is incredibly difficult for me to just pass something like this. But as I said, I should not mix fixes and enhancements.

dalf:
Looking at https://github.com/asciimoo/searx/pull/1938/files#diff-0763b2abd2a8af9c055a95269c98d956R100 I understand the static files are copied ?

What about https://flask.palletsprojects.com/en/1.1.x/blueprints/#blueprint-resources ?

return42:
> I understand the static files are copied ?

TLDR; yes.  

We have different resources which are needed on the cient side.  Those resources are coming from (included in) Themes (jinja templating), from plugins and maybe later from other objects of the web application.  These resources has to be available at a static_url, for this we need to copy this resources into a _static folder_ (compare [architecture](https://asciimoo.github.io/searx/admin/architecture.html))

As long as we have only *builtin* plugins (templates, engines etc.) the handling of those static files is very simplle, we placed same all in the /static folder.   With external extensions we need to manage the static files and the static URLs.

For this, resources.py implements a [StaticFiles](https://github.com/return42/searx/blob/ep-plugins/searx/resources.py#L68) class.  ATM this class has two methods, to load (copy) resources (files) from themes and plugins.

In this patch I handling files from Themes on the filesystem just like we have done before .. new is, that these files are now copied.

I have to look where Blueprints can help by handling files (resources) from Themes (jinja templates).

But I cannot assess whether and how we have to change our templating for this. I want to change the templating as little as possible. As this may can have very far-reaching consequences.

return42:
The use of Flask-Blueprints means that we have to reorganize our entire template management.

The blueprints are registered on a route, but we want a Theme selection .. as far as I can see Theme selection is not possible with Flask-Blueprints. Therefore I first refrain from using blueprints.

But I also let myself be convinced if someone has an idea how we can use blueprints sensibly for our themes ...

dalf:
Yes good point.

I remember there is a way to combine multiple FS path (instead of blueprint), but I can't remember I found it (may be in a comment on an issue somewhere in this repository).

return42:
> I remember there is a way to combine multiple FS path 

I guess you remember about this https://buxty.com/b/2012/05/custom-template-folders-with-flask/

I found it in your https://github.com/asciimoo/searx/issues/1557#issuecomment-481934009

It is a way we have to think about .. but that also goes beyond this PR.  In this PR I don't wont to change the existing file structure nor experimenting with jija loaders.. 

In this PR I just want to put some architecture in the source code (modularize things/resources, that should not be in the webap.py) and manage the file structure as is.

return42:
Is there anything else I can/should do?

docs/dev/plugins.rst
```diff
@@ -30,25 +44,40 @@ Example plugin
...
+           'searx.plugins': [ 'foo_example_plugin = foo.example.plugin', ],
+       },
+   )
+
+.. _external plugins:
 
-Entry points (hooks) define when a plugin runs. Right now only three hooks are
-implemented. So feel free to implement a hook if it fits the behaviour of your
-plugin.
+Kown plugins from the Web
```
kvch: Kown? Is it supposed to be known?

kvch:
Could you please provide more context about why you changed the license headers of files?

searx/resources.py
```diff
@@ -0,0 +1,886 @@
...
+        os.remove(self)
+
+    def filesize(self, precision=None):
+        """Filesize in bytes or with precision"""
+        size = path.getsize(self)
+        if precision is not None:
+            size = humanize_bytes(size, precision)
+        return size
+
+    def openTextFile(self
```
kvch: Could you please use `snake_case` for function names?

kvch:
@return42 I like your PR overall. I have a few minor comments and a question about the code.
Once those are addressed and the CI turns green, I am OK with merging this.

kvch:
I've just checked out this branch to test the PR. I am having trouble with the plugins. When I open the preferences page and to the plugins tab, the list is empty. (I haven't changed anything in settings.yml or preferences.)

Am I doing something wrong?

return42:
@kvch thanks a lot for your review!

> Could you please provide more context about why you changed the license headers of files?

We are going to use  [SPDX License Identifier](https://spdx.org/using-spdx-license-identifier). It is addressed in https://github.com/asciimoo/searx/issues/1801

return42:
> and the CI turns green

On my local desktop system the `make test` runs without any issue. I guess its because my local environment is highly different. I will take a look what the problem is.

BTW: https://github.com/asciimoo/searx/pull/1803 addresses such problems, if you or anyone else have time to review #1803 .. ;)

return42:
> I've just checked out this branch to test the PR.  ... Am I doing something wrong?

If `make run` works without any issue, I guess I know what the problem is:

To get in use of the _built-in_ plugins, their entry points needs to be registered and this is only done once, when [searx's setup.py](https://asciimoo.github.io/searx/admin/installation.html#how-to-update) is called.

I totally forgot to cover the use case of a simple  update as we described [here](https://asciimoo.github.io/searx/admin/installation.html#how-to-update).

I have to think about / do you have any idea how we can handle this as best, without getting into trouble with already running instances which need to be updated?

return42:
@kvch
>  When I open the preferences page and to the plugins tab, the list is empty. (I haven't changed anything in settings.yml or preferences.)

Argh ..  now I get the same error .. I will have a look and come back.

return42:
I found the issue: in the old install instruction, we use the manage.sh script to install searx dependencies, but we never install searx itself into the virtualenv ... see my patch https://github.com/return42/searx/commit/f576cadfba11dd96788f5c70e63fba4a70899584

The patch is already in [searx-next branch](https://github.com/return42/searx/tree/searx-next), a preview of the docs can be found here:

- https://return42.github.io/searx/admin/installation-searx.html#install-searx-dependencies

_search for .._


    # jump to searx's working tree and install searx into virtualenv
    (searx)$ cd /usr/local/searx/searx-src
    (searx)$ pip install -e .

return42:
@dalf @kvch :  I merged master into #1938 and answered your questions .. do you have time to continue with the review? / thanks!

return42:
To test locally I used:

````shell
$ make install
..
$ make run
...
````

go to "preferences --> plugins" and enable  "Only show green hosted results".

**Comment by return42** [comment]
@kvch can you please continue with your review/ thanks! 

Did I missed to answer any of your questions?

File: searx/search.py [hunk 5]
```diff
@@ -530,9 +515,10 @@ def __init__(self, search_query, ordered_plugin_list, request):
         self.request = request
 
     def search(self):
-        if plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self):
-            super(SearchWithPlugins, self).search()
+        plugins = get_plugins()
 
+        plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self)
```
dalf, line 532: Why remove ```if ...: super(SearchWithPlugins, self).search()``` ?

return42: Hi @dalf, thanks for taking the time for a review.

1. The method ```isuper(SearchWithPlugins, self).search()``` should be called even if the `self.ordered_plugin_list` is empty .. otherwise the search call is not executed when all plugins are disabled. 

2. It make no sense to call the post hook if the pre hook had an empty list of plugins

I mean, if I'm not wrong, all 3 calls needs to be executed.

````python
        plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self)
        super(SearchWithPlugins, self).search()
        plugins.call(self.ordered_plugin_list, 'post_search', self.request, self)
````
It could be optimized to:

````python
        if self.ordered_plugin_list:
            plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self)

        super(SearchWithPlugins, self).search()

        if self.ordered_plugin_list:
           plugins.call(self.ordered_plugin_list, 'post_search', self.request, self)
````
But this use case is rare, since in most installations at least the _HTTPS rewrite_ plugin is enabled.

What do you think, did I overlooked something?

Comment (dalf) [reply searx/search.py:532]: I don't know the right answer right now.

But here, some notes / context:
* This comment is related to https://github.com/asciimoo/searx/pull/2027 (if the PR remains a plugin).
* Except this PR, AFIK, there is no ```pre_search``` plugin.
* This design exists since the [first commit about the plugins](https://github.com/asciimoo/searx/commit/00cc4dcbf44d9ecea89befb08cae4ee5561c4247#diff-02503eba222466e11a3f7ae36debdc5cR328) and [this was kept](https://github.com/asciimoo/searx/commit/67e11c42b973932c8f568d80a0f25bfd7fc150ab#diff-35de57dede17c7660869ae806bbc99b6R367-R368) over time

I think this ```if``` overlaps the answers (?). For example if you search for ```random string``` you get this:
image

Answer were created for privacy reason: see https://github.com/asciimoo/searx/pull/797#issuecomment-269694411

Comment (return42) [reply searx/search.py:532]: @dalf - I haven't seen #733 before and thanks for pointing to [".. to stop if there is an answer already from the answerers"](https://github.com/asciimoo/searx/pull/797#issuecomment-269694411) this was not clear to me.  If I take both into account I come to the conclusion that it was a bug from the first commit.  Maybe I'm wrong, but read by yourself ..

- side node: when I looked at the sources I was a bit confused about the names `answerers` and `answers`, to clarify:

  -  `answerers` is what we talk about, implemented in a python module with the exact same name.
  - `answers` are a part of the result container, containing the _"Answers from WEB engines"_

`answerers` is called in base class's method `Search.search()` and this method returns immediately if there are any _local_ "Answers"

````python

    # do search-request
    def search(self):
        global number_of_searches

        # start time
        start_time = time()

        # answeres ?
        answerers_results = ask(self.search_query)

        if answerers_results:
            for results in answerers_results:
                self.result_container.extend('answer', results)
            return self.result_container

````

What means, the search method of the base class needs to be called, even if there are no plugins activated. Compare old inheritance **branch master**:


````python
    def search(self):
        plugins = get_plugins()
        if plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self):
            super(SearchWithPlugins, self).search()
        plugins.call(self.ordered_plugin_list, 'post_search', self.request, self)

````

with the patched from **branch ep-plugins**..

````python
    def search(self):
        plugins = get_plugins()

        plugins.call(self.ordered_plugin_list, 'pre_search', self.request, self)
        super(SearchWithPlugins, self).search()
        plugins.call(self.ordered_plugin_list, 'post_search', self.request, self)

````

I have to make a test with the old implementation and all plugins disabled ... I guess this use case hasn't ever been tested in the past.

Comment (return42) [reply searx/search.py:532]: Pew .. hard to explain, I will try it in just a view words: 

1. in branch master `plugins.call(self.ordered_plugin_list,...)` returns True when the plugin list is empty / see: https://github.com/asciimoo/searx/blob/master/searx/plugins/__init__.py#L70

2. in branch ep-plugins  `plugins.call(self.ordered_plugin_list,...)` returns an empty list when the plugin list is empty / see: https://github.com/return42/searx/blob/ep-plugins/searx/plugins/__init__.py#L216

Conclusion:

1. the `if` condition in branch master was useless from the beginning since the `plugins.call(..)` always returns a value with a typecast into a _boolean_ `True`.
  
2. the implementation of the method `SearchWithPlugins.search(..)` in **branch ep-plugins** is _functional identical_ to the one in **branch master**.

I'm wrong / other meanings?

Comment (dalf) [reply searx/search.py:532]: https://github.com/asciimoo/searx/blob/575159b194440052d7b48aa073d7e03c80799c90/searx/plugins/__init__.py#L69-L77

Line 73, ```ret``` is assign to return value of the function (```pre_search``` in case of pre_search plugin).

---

If I try to compare.

Results:
* ```answerers ``` returns ```{url: '....', answer: '...' }```. See https://github.com/asciimoo/searx/blob/2c6531b2330a5c8ca986379dbad8812a6c6d3e49/searx/templates/oscar/results.html#L96-L106
* ```pre_search``` plugins can produce anything : infobox, result, answer, corrections, etc...

How to trigger:
* ```answerers ``` : there are [keywords](https://github.com/asciimoo/searx/blob/2c6531b2330a5c8ca986379dbad8812a6c6d3e49/searx/answerers/__init__.py#L14-L32).
* ```pre_search``` plugins: access to [SearchQuery](https://github.com/asciimoo/searx/blob/2c6531b2330a5c8ca986379dbad8812a6c6d3e49/searx/query.py#L177) object.

In way, in the current implementation (with the ```if```), ```answerers ``` are specialized ```pre_search``` plugins. The core could even create on the fly a plugin from an ```answerers```.

Comment (return42) [reply searx/search.py:532]: > Line 73, ret is assign to return value of the function (pre_search in case of pre_search plugin).

OK, I see, there is a functional change from:

````python

    def call(  # pylint: disable=missing-function-docstring, no-self-use
            self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
        ret = True
        for plugin in ordered_plugin_list:
            if hasattr(plugin, plugin_type):
                ret = getattr(plugin, plugin_type)(request, *args, **kwargs)
                if not ret:
                    break

        return ret
````

to ..

````python
    def call(self, plugin_list, name, *args, **kwargs):
        ...
        f_list = [ getattr(p, name) for p in plugin_list ]
        for func in f_list:
            ret_val.append(func(*args, **kwargs))
        return ret_val

````

But is this really a problem, do we have any drawback with?  Take into account the _execution model_; that we now have external plugins, should a plugin really cancel the execution of the other plugins .. I don't think so.

> In way, in the current implementation (with the if), answerers are specialized pre_search plugins. The core could even create on the fly a plugin from an answerers.

I didn't grasp all, I have to deal with it more .. I mean: I still do not know if I have damaged anything functionally.

Comment (dalf) [reply searx/search.py:532]: > I didn't grasp all, I have to deal with it more .. I mean: I still do not know if I have damaged anything functionally.

Sorry it was more open thought about the searx implementation.

Take make it clear, let's keep the thing as it is in your PR, and if later we need change, it can be another PR.

**Comment by kvch** [comment]
When I tested your plugin `tgwf-searx-plugins` I got the following expection:
```
ERROR:searx.webapp:search error
  File "/home/n/p/searx/searx/webapp.py", line 569, in index
    result_container = search.search()
  File "/home/n/p/searx/searx/search.py", line 522, in search
    plugins.call(self.ordered_plugin_list, 'post_search', self.request, self)
  File "/home/n/p/searx/searx/plugins/__init__.py", line 222, in call
    ret_val.append(func(*args, **kwargs))
  File "/home/n/p/searx/local/py3/lib/python3.5/site-packages/only_show_green_results.py", line 108, in post_search
    result for result in list(search.result_container._merged_results)
  File "/home/n/p/searx/local/py3/lib/python3.5/site-packages/only_show_green_results.py", line 109, in <listcomp>
    if GC.check_url(result.get('url'))
  File "/home/n/p/searx/local/py3/lib/python3.5/site-packages/only_show_green_results.py", line 67, in check_url
    ret_val = self.check_against_api(parsed_domain)
  File "/home/n/p/searx/local/py3/lib/python3.5/site-packages/only_show_green_results.py", line 98, in check_against_api
    response = requests.get(api_url).json()
  File "/home/n/p/searx/local/py3/lib/python3.5/site-packages/requests/models.py", line 898, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.5/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.5/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.5/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Otherwise your PR looks good to me.

**Review by kvch — DISMISSED** [review]

**Review by asciimoo — COMMENTED** [review]
I think this is a great proposal and it helps a lot to be able to sustainably maintain the core of searx, but after spending a couple of hours with reviewing I still can't state that this is the way to implement this feature. Perhaps I need more time to understand the whole PR and the purpose of every component/class, but currently I still have a feeling that there are much more things changing/being added than it would be necessary to achieve plugin decoupling.

@return42 could you please explain:
 - What were exactly the goals you wanted to achieve with this PR?
 - Why did you decided to implement the feature this way?
 - What is the purpose of creating an individual abstraction layer for static files, themes, templates, files, env in resources.py with lots of unused functions?

Sorry for the numerous questions, I really appreciate your work. Probably it's my fault that I don't fully understand the implementation.

File: searx/version.py [hunk 6]
```diff
@@ -1,26 +1,30 @@
...
-                                      VERSION_MINOR,
-                                      VERSION_BUILD)
+VERSION_STRING = "{0}.{1}.{2}".format(
+    VERSION_MAJOR,
+    VERSION_MINOR,
+    VERSION_BUILD
+)
+
+# built-in plugins
+BUILTIN_PLUGINS = [
```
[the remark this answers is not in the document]
return42: Me too, it is a compromise. Like the version number, we also need the list of built-in-plugins in the [setup.py](https://github.com/return42/searx/blob/ep-plugins/setup.py#L60).  But instead creating another file with package meta data -- which is needed in the setup.py -- I vote for renaming this version.py file in something like __pkginfo__.py

searx/resources.py
```diff
@@ -0,0 +1,886 @@
...
+      True, True
+      >>> for n in tmp.listdir():
+      ...     print(tmp / n)
+      ...
+      /home/user/tmp/foo
+      /home/user/tmp/bar
+
+    """
+
+    class OS:
```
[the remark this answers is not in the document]
return42: class File helps me to write much more readable code. This implementation here is the surrogate from a [fspath package I maintain](https://return42.github.io/fspath/slides/index.html#/).  The linked slide show from this project might give a impression where it can help to be more expressive and readable.

searx/resources.py
```diff
@@ -0,0 +1,886 @@
...
+        """True if path is absolute (:py:func:`os.path.isabs`)"""
+        return path.isabs(self)
+
+    @property
+    def ISLINK(self):
+        """True if path is a symbolic link (:py:func:`os.path.islink`)"""
+        return path.islink(self)
+
+    @property
+    def ISMOUNT(self):
```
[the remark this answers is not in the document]
return42: These are properties ... 

````python
if myfile.EXISTS:
     print("file size: %s" % (myfile.SIZE))
````

I decided to use uppercase since this is in tradition with [os.stat](https://docs.python.org/3/library/stat.html).

> As I see, only a few of them are actually used in the current code.

It is a lib we can make use of. It helps me (and I think others also) to handle file resources more expressive and readable (as I mentioned near to the other remarks from you).

return42:
> @return42 could you please explain:

Yes, sorry ..

> What were exactly the goals you wanted to achieve with this PR?

The motivation was to "decoupling plugin development from the searx core".  To do so first an architecture was needed to handle the different _resources_ we have.  Before I started to implement I polished source-code and documentation without any functional change (see 732452800f3).  In this commit I also added one of my long-distance targets, the pylint check (replacing pep8).

The next major patch is commit 6abebc261 which brings the "abstraction layer for the resources".  This architecture has a factory to get a *global* manager for static files see [get_static_files()](https://github.com/return42/searx/blob/ep-plugins/searx/resources.py#L51) or [get_templates()](https://github.com/return42/searx/blob/ep-plugins/searx/resources.py#L202).  Both are used in the webapp.py to build up the application.

In the next commit 169f5c3ba0 the detection of _internal-plugins_ has moved to python's entry points.  From this point on we _support_ external plugins

Afterwards there are some minor patches with bug-fixing and documentation in.  At some point the patches are coming from the review process (ATM I see only one add14e9d05).

> Why did you decided to implement the feature this way?

I tried to explain it above and in the doc-strings .. I believe that the architecture I modeled is a flexible base for managing resources, I'm open for other models.

> What is the purpose of creating an individual abstraction layer for static files, themes, templates, files, env in resources.py with lots of unused functions?

The abstraction layer is a (stable) API which can be used on both sides:

- On the application builder site (webapp.py I described above) and
- when developing external plugins.

I think that in order to decouple things better we need abstractions.

return42:
> When I tested your plugin tgwf-searx-plugins I got the following expection:

@kvch I try to reproduce your error but I can't. This might be due to the fact that my scenarios are different to yours.  I have tested on my hosted instance and I can test it in developer mode without any fail:

````bash

$ make clean install
$ pip install git+https://github.com/return42/tgwf-searx-plugins
$ make run
...
````

Can you give me some hints about your scenario?

searx/resources.py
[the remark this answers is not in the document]
asciimoo: My note isn't about the uppercasing, it's about the so many unused functions what do not add any value to the current codebase. I consider these functions as [dead code](https://en.wikipedia.org/wiki/Dead_code) or [unreachable code](https://en.wikipedia.org/wiki/Unreachable_code) which can lead to numerous issues in the future and it also makes reviewing slower and harder. Could you eliminate unused functions?

searx/resources.py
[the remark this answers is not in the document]
asciimoo: My comment was about the `OS` class which seems to be only an alias to functions can be found in the standard libs of python.

asciimoo:
> I tried to explain it above and in the doc-strings .. I believe that the architecture I modeled is a flexible base for managing resources, I'm open for other models.

I'm not sure if such architectural changes are required to be able to accomplish plugin decoupling. Maybe it is just my personal preference, but I like to keep things as simple as possible. In my experience, it reduces the maintenance costs and makes future contributions easier in general.

I'm totally open to change my viewpoint in this specific case and also in general if you could explain the benefits of these decisions, but currently I think the same functionality can be achieved with only a single `Plugin` class and/or with a few functions to handle python entry points and register the plugins. What do you think?

I'd also love to hear the opinion of others (perhaps @dalf @kvch @hlfh ? =) ).

searx/version.py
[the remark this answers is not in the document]
return42: Should I rename the file in this PR?

searx/resources.py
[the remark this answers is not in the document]
return42: >  it's about the so many unused functions

They exists for programmers convenience, to implement more readable source code. If aspects like [dead code](https://en.wikipedia.org/wiki/Dead_code) or [unreachable code](https://en.wikipedia.org/wiki/Unreachable_code) overrule _comfort_ we should remove it.   But isn't it the same with every lib we use? If _unused_ names are the only KO criteria for the merge, I can clean up the code.

dalf:
> I'd also love to hear the opinion of others 

I thought about this PR these last days, but I actually I don't have a clear answer (I was ok at first as I told return42).

Let's put some categories to the existing plugins:
* in each result, the plugin change the URL:
	* https_rewrite: ensure the usage of https
	* oa_doi_rewrite: replace reference to a DOI
	* tracker_url_remover: remove the tracker
* UI/UX:
	* infinity_scroll
	* open_results_on_new_tab (perhaps I'm wrong, but this feature is already part of the core features: https://github.com/asciimoo/searx/blob/6ab569f5588e148bf73a7d80bc2f52bc737bf5e0/searx/templates/oscar/preferences.html#L114-L121 )
	* search_on_category_select
	* vim_hotkeys
* add new results:
	* self_info: add an answer (the IP address or the user agent).

Some plugins are
* result modifiers
* UI/UX options
* content provider

All this plugins are inside the asciimoo/searx git repository.
Another name for plugins could be features.

So from here, it seems there is no need to decouple plugins/features from the core repository.

But then we can focus on the https_rewrite plugin:
* The current rules are : https://github.com/asciimoo/searx/tree/master/searx/plugins/https_rules
* It is only a subset of a very old version of the actual browser plugin.
* we could imagine that the https_rewrite plugin/feature could be a different repository. Then every day/week the rules are synchronized automatically from https://github.com/EFForg/https-everywhere (it requires a faster Python implementation for sure). In this case the update process focus only on https_rewrite nothing else.
* The searx can have a dependency to this repository (always the last version).
* If combine with https://github.com/asciimoo/searx/issues/2034 , to update searx an administrator needs:
  * to update the git repository.
  * update the dependencies (as before, but we have to make sure there is a version upgrade).
  * download the assets.

An alternative view is to focus on the build process:
* some files are built (like the .min.js, .min.css files). For now these files are commited into the repository, but there is https://github.com/asciimoo/searx/issues/2034
* some files are fetched & parsed (like https://github.com/asciimoo/searx/tree/master/searx/data )
* some files are fetched without modification (like https://github.com/asciimoo/searx/tree/master/searx/plugins/https_rules and https://github.com/asciimoo/searx/blob/f068488cabb95156807dae88cd3f73291e54b46a/searx/plugins/bangs_data/bangs.json )
* so the https_rules could be fetched at the build process (perhaps a different asset than the static files). In this view, the https_rewrite plugin can remain inside the main repository.

A third view is to look at this PR : https://github.com/asciimoo/searx/pull/1029
* The purpose was to have different maintainers inside the same repository.
* The PR was close for "lack of interrest", but I think it is more or less related to https://github.com/asciimoo/searx/issues/2025 (perhaps from a far distance, but still).

Another thing:

> I like to keep things as simple as possible. In my experience, it reduces the maintenance costs and makes future contributions easier in general.

I agree: fewer line of code, fewer maintenance.

And a last thing: the main theme is "oscar", I don't know if the "simple" theme is used. In my opinion the maintenance cost of different themes is burden. Perhaps we could put the alternative themes in a different repository, with different maintainers. Once again this is only an idea.

searx/resources.py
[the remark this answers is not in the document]
return42: > My comment was about the OS class 

Ah, sorry ,,

> which seems to be only an alias to functions can be found in the standard libs of python.

Yes, these are aliases glued at the File object, making imports from os.path obsolete / same topic: [They exists for programmers convenience](https://github.com/asciimoo/searx/pull/1938#discussion_r447614158). Here is a hypothetical  example:
````python
if myfile.DIRNAME == myfile.OS.curdir():
    # file is in the current working directory
    ....
````

If the `File.OS` namespace is a KO criteria [I can clean up the code](https://github.com/asciimoo/searx/pull/1938#discussion_r447614158). But should we really remove all _convenience_ from .. I have a strange feeling with.

return42:
> * open_results_on_new_tab (perhaps I'm wrong, but this feature is already part of the core features: https://github.com/asciimoo/searx/blob/6ab569f5588e148bf73a7d80bc2f52bc737bf5e0/searx/templates/oscar/preferences.html#L114-L121 )

just a short note: Templating existing themes from the searx core should not be the task of a _plugin interface_.  (External) plugins  can use ['js_dependencies' & 'css_dependencies'](https://github.com/return42/searx/blob/ep-plugins/searx/plugins/__init__.py#L61).

return42:
> Perhaps we could put the alternative themes in a different repository, with different maintainers. Once again this is only an idea.

This is also my intention .. this PR is focused on _"plugins"_, but I could imagine one day we do the same for _"themes"_ and  _"engines"_.

I could well imagine further separating searx more and more  into function modules that are plugged together into an application.  ATM we have webapp.py as application, but why not using searx as lib, from which everyone can plug together it's application?

asciimoo:
> in my opinion the maintenance cost of different themes is burden

I totally agree.

> Perhaps we could put the alternative themes in a different repository, with different maintainers. Once again this is only an idea.

That's a good idea, but it doesn't really require architectural changes. It can be done with two symlinks to the templates/static folder. Same true in case of engines.

> A third view is to look at this PR : #1029

I'm still all for introducing it.

return42:
> I'm totally open to change my viewpoint in this specific case and also in general if you could explain the benefits of these decisions, but currently I think the same functionality can be achieved with only a single Plugin class and/or with a few functions to handle python entry points and register the plugins. What do you think?

I guess, there is always a chance to implement it with less architectural.  This is my approach which tries to change as less as possible in existing plugins and themes.  The [resources.py](https://github.com/return42/searx/blob/ep-plugins/searx/resources.py) from  commit 6abebc261c4 emulates exactly what we need to handle (static) files in context of _plugins_ and _templating & themes_. 

In other words commit 6abebc261c4 is a [drop in replacement](https://en.wikipedia.org/wiki/Drop-in_replacement) for the way we handling (static) files. (update: there is also a [documentation](https://return42.github.io/searx/dev/sources.html))

At the same time the implementations of this PR has been lead by the thoughts I mentioned at https://github.com/asciimoo/searx/pull/1938#issuecomment-651757555:

> I could well imagine further separating searx more and more into function modules that are plugged together into an application. ATM we have webapp.py as application, but why not using searx as lib, from which everyone can plug together it's application?

searx/resources.py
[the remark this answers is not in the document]
return42: Just an idea from mine, I don't know if it fits to your request: .. I tested all functions, when I implemented them, I can also implement unit tests ..

return42:
See #2074

PR #1950, opened 2020-05-12
[fix] about page- 'How to set as the default search engine?'

### LINKED ISSUE #1616: [Feature Request] Standard right-click and add to browser
Most search engines make it possible for browsers to detect that the search box is a search box, so that when you right-click in it you can add it to your browser with all the correct info needed.
This will work for all browsers so no special links are needed, just the right info in the code.
I wanted to be more useful and provide a link, but alas I was informed by a friend how it works but not the specifics.

Mind you, I would note that if a user cannot work out how to add searx manually, it is probably not the right search engine for them ;)

### LINKED ISSUE #1849: Slight confusion when adding searx to Firefox
Having set up a local instance of searx, I was attempting to add it to Firefox as the default search engine.
Although pointing my browser at localhost:8888 worked fine, after adding it to Firefox, the browser insisted on trying to use the default port 80, so went to localhost.
After much digging about, I fell upon the solution. In settings.yml, I had defined:

`base_url : http://antwerp # Set custom base_url. Possible values: False or "https://your.custom.host/location/"` (where antwerp is my local hostname).
Under the assumption that the port defined above this line was "taken as read". However, in order for Firefox to use the correct port, the line needs to be:
`base_url : http://antwerp:8888 # Set custom base_url. Possible values: False or "https://your.custom.host/location/"`

My suggestion: either the generated xml file should be created automatically with the port appended, or the comments in the default settings.yml file should make it clear that it needs to be added, to save other users having this problem.

### LINKED ISSUE #1949: Inline JS is blocked when HTTP CSP is enabled
Adding searx to browser's search engines does not work when CSP is enabled.

https://github.com/asciimoo/searx/blob/2d4d0bdc0cdc239dee931ba98a95bd632eb645e2/searx/templates/__common__/about.html#L65

### LINKED ISSUE #2007: Set searx as the default search engine? doesn't work
Can't set searx as default search engine firefox
If in the about on install click nothing happens.

return42:
Searx supports OpenSearch, remove the inline JS from the about page.

close: https://github.com/asciimoo/searx/issues/1949
close: https://github.com/asciimoo/searx/issues/1616
close: https://github.com/asciimoo/searx/issues/1849
close: https://github.com/asciimoo/searx/issues/2007


return42:
A preview is available at: https://darmarit.org/searx/about

return42:
Sorry for bumping #1950: its a minimal change closing 4 issues .. do you have time to review / thanks!

dalf (approved):

Issue #1512, opened 2019-02-12
[Enhancement] Add rotating outgoing proxies


jfowl:
Excerpt from settings.yml:
```yaml
    proxies :
        http : http://127.0.0.1:8080
        https: http://127.0.0.1:8080
    source_ips:
        - 1.1.1.1
        - 1.1.1.2
```

Search engines, especially google, tend to captcha-block your IP after too many requests. Although this is already somehow addressed with allowing multiple source IPs, not every server does have multiple IPs available.
On the other side, there is only support for a single pair of outgoing proxies. Couldn't we apply the source_ip strategy:
```python
if settings['outgoing'].get('source_ips'):
    http_adapters = cycle(HTTPAdapterWithConnParams(pool_connections=connect, pool_maxsize=maxsize,
                                                    source_address=(source_ip, 0))
                          for source_ip in settings['outgoing']['source_ips'])
```
of cycling to outgoing proxies.
I imagine the settings.yaml format to be changed somewhat like this:
```yaml
    proxies :
        - {http : http://127.0.0.1:8080, https: http://127.0.0.1:8080}
        - {http : http://127.0.0.2:8080, https: http://127.0.0.2:8080}
```

This way, even higher load instances could circumvent some blockage and avoid regular problems like #1441.

I personally think this should not be too difficult to implement, but I suck at python, so...

SilverBut:
Maybe we can use `requests[socks]` so we can [use the socks proxy](http://docs.python-requests.org/en/master/user/advanced/#socks), and have a quick implement of this feature using `proxychains`?

But I also think this feature should be added into the roadmap. And maybe the following syntax is possible:

```yaml
proxies:
  mode: rotate
  servers:
    - http://user:pwd@proxy.example.org:8080
    - https://user:pwd@proxy.example.org:8443
    - socks5://user:pwd@proxy.example.org:1080
    - morty://morty:key@proxy.example.org:3000
```

dngray:
Out of [all the instances there](https://github.com/asciimoo/searx/wiki/Searx-instances#alive-and-running) only 11 of them are actually working with Google at the moment. Unfortunately Google is still useful.

I wonder if it would be possible to buy a bunch of IPv6 IPs to proxy to ipv6.google.com and have just AAAA record?

If google is indeed banning by IP then that will get pretty exhaustive on IPv4 addresses. We already know that most of the time if you try to use something like Tor, you will get errors.

jfowl:
@dngray 
[...]

Issue #1827, opened 2020-02-02
Using proxy only for specific engines


sevospl:
Is it possible to use a proxy only for specific engines?

unixfox:
Good idea! I'm also interested to have this feature.

return42:
I don't get that.. what is the benefit?

unixfox:
An example that I would think: to circumvent the Google block.

return42:
may I am totally wrong, do we talk about the _proxied_ link (aka morty link)

unixfox:
Ah. Maybe the author is talking about that. I was talking about when using an http proxy

sevospl:
> Ah. Maybe the author is talking about that. I was talking about when using an http proxy

HTTP proxy.

bauruine:
I did a PoC patch here https://github.com/bauruine/searx/commit/ea38ad6338e738c496b9cc77b860dc4a7644649e it's not exactly per engine but it adds another config option to limit the proxy usage for specific urls. I'm not sure if this is acceptable. I first tried to implement it like the per engine timeout but that doesn't actually work the way i thought it would. If you have multiple per engine timeouts and you are doing requests to multiple engines the biggest timeout will be used for all requests. So this doesn't work for this use case.

If someone has a better idea how to implement this please let me know.

unixfox:
@asciimoo I just tested your commit https://github.com/searx/searx/commit/4a913247b29286299e678e0756119fb62293760f and it doesn't seem to work.
I set the `proxies` parameter for the google engine like so:
```yml
  - name : google
    engine : google
    shortcut : go
    proxies: "http://127.0.0.1:8080"
```

And I'm getting this error when doing a search:
```
ERROR:searx.search:engine google : exception : 'str' object has no attribute 'get'
Traceback (most recent call last):
  File "/tmp/searx/searx/search.py", line 164, in search_one_http_request_safe
    search_results = search_one_http_request(engine, query, request_params)
  File "/tmp/searx/searx/search.py", line 101, in search_one_http_request
    response = send_http_request(engine, request_params)
  File "/tmp/searx/searx/search.py", line 85, in send_http_request
    return req(request_params['url'], **request_args)
  File "/tmp/searx/searx/poolrequests.py", line 134, in get
    return request('get', url, **kwargs)
  File "/tmp/searx/searx/poolrequests.py", line 110, in request
    response = session.request(method=method, url=url, **kwargs)
[...]

PR #1958, opened 2020-05-16
[RFC ] Add multiple outgoing proxies

### LINKED ISSUE #1899: Proxy rotation on searx
How to add more proxies to the settings.yml in searx. When to use http and when to use https?

bauruine:
This is similar to source_ips as it makes it possible to use multiple proxies to do outgoing requests.
Fixes: #1899 and #1512

I'm not sure if this is the correct way to do it comments appreciated. Also `get_proxies()` is currently copy pasted in `searx/poolrequests.py` and `searx/webapp.py` i'm not sure where to put that function. Maybe `searx/utils.py`?

### DIFF
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -87,6 +87,21 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
+def get_proxies():
+    proxy_settings = settings['outgoing'].get('proxies')
+    if proxy_settings:
+        # Backwards compatibility for single proxy in settings.yml
+        for protocol, proxy in proxy_settings.items():
+            if isinstance(proxy, str):
+                proxy_settings[protocol] = [proxy]
+        for protocol in proxy_settings:
+            proxy_settings[protocol] = cycle(proxy_settings[protocol])
+    return proxy_settings
+
+
+proxies = get_proxies()
+
+
 def request(method, url, **kwargs):
     """same as requests/requests/api.py request(...)"""
     time_before_request = time()
@@ -94,8 +109,11 @@ def request(method, url, **kwargs):
     # session start
     session = SessionSinglePool()
 
-    # proxies
-    kwargs['proxies'] = settings['outgoing'].get('proxies') or None
+    if proxies:
+        proxy = {}
+        for protocol in proxies:
+            proxy[protocol] = next(proxies[protocol])
+        kwargs['proxies'] = proxy
 
     # timeout
     if 'timeout' in kwargs:
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -87,7 +87,7 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
-def get_proxies():
+def get_proxy_cycles():
     proxy_settings = settings['outgoing'].get('proxies')
     if proxy_settings:
         # Backwards compatibility for single proxy in settings.yml
@@ -99,7 +99,15 @@ def get_proxies():
     return proxy_settings
 
 
-proxies = get_proxies()
+proxy_cycles = get_proxy_cycles()
+
+
+def get_proxies():
+    global proxy_cycles
+    proxy = {}
+    for protocol in proxy_cycles:
+        proxy[protocol] = next(proxy_cycles[protocol])
+    return proxy
 
 
 def request(method, url, **kwargs):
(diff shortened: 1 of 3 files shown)


unixfox:
Good job!
Doesn't this also close this issue: https://github.com/asciimoo/searx/issues/1512?

Also, while you are at it. Do you mind developing some code for this related issue: https://github.com/asciimoo/searx/issues/1827?

bauruine:
You are right that issue is about the same feature. I'll look into #1827 tomorrow. I don't know the code base very well so I'm not sure if it's possible to implement it without too much work.
Any other comments? Especially about where to put `get_proxies()` into.

unixfox:
@dalf

unixfox:
@dalf could you take a look at this PR? And maybe answering the author question:
> Especially about where to put `get_proxies()` into.

dalf (changes requested):
Sorry for the very long delay.

unixfox:
Hello @bauruine I would really like to move forward with this PR, could you take a look at dalf review?

searx/poolrequests.py
```diff
@@ -87,15 +87,38 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
+def get_proxy_cycles():
+    proxy_settings = settings['outgoing'].get('proxies')
+    if proxy_settings:
```
dalf, line 92: if `outgoing.proxies` is not defined in settings.yml, the result of this function is `None`, on line 108 it will trigger `'NoneType' object is not iterable`.

```python
proxy_settings = settings['outgoing'].get('proxies')
if not proxy_settings:
   return None
...
```

later in get_proxies
```python
def get_proxies():
    global proxy_cycles
    if not proxy_cycles:
        return None
    ...
```

searx/poolrequests.py
```diff
@@ -87,15 +87,38 @@ def get_time_for_thread():
...
+
+proxy_cycles = get_proxy_cycles()
+
+
+def get_proxies():
+    global proxy_cycles
+    proxy = {}
+    for protocol in proxy_cycles:
+        proxy[protocol] = next(proxy_cycles[protocol])
+    return proxy
```
dalf, line 110: lines 107 to 110 can be be sum up:
```python
return { protocol: next(proxy_cycles[protocol]) for protocol in proxy_cycles }
```

bauruine:
I just tried to rebase on master but this doesn't work with https://github.com/searx/searx/commit/4a913247b29286299e678e0756119fb62293760f and i have no idea how to fix it. The only thing i can think of right now is forgetting the cycle and just do a random.choice() on each request. But maybe that's just my limited python knowledge.

dalf:
poolrequests.py:  line 144:

```python
    if kwargs.get('proxies') is None and proxy_cycles:
		kwargs['proxies'] = get_proxies()
```

( set a proxy only if the engine has not defined a proxy and the proxy_cycles is not None = some proxies are defined in the settings )

[EDIT] I don't know if it helps ?

searx/webapp.py
```diff
@@ -48,6 +48,7 @@
 from six import next
 from datetime import datetime, timedelta
 from time import time
+from itertools import cycle
```
dalf, line 51: This line can removed: `cycle` is not required anymore.

dalf:
Some thoughts to support multiple proxies per engine.

https://github.com/searx/searx/blob/f1bcd48f02760605e39052169d95f7f2b193045b/searx/poolrequests.py#L90-L91

Change the signature to get proxy_settings from a parameter.

So:
https://github.com/searx/searx/blob/f1bcd48f02760605e39052169d95f7f2b193045b/searx/poolrequests.py#L102
becomes `get_proxy_cycles(settings['outgoing'].get('proxies'))`

In
https://github.com/searx/searx/blob/e78bfd4d8b62a521e08029bb0e708b2bcec9958d/searx/engines/__init__.py#L60

Check if there is `engine_data.proxies` attribute, if yes replace this value by `get_proxy_cycles(engine_data.proxies)`.

And last change here:
https://github.com/searx/searx/blob/e78bfd4d8b62a521e08029bb0e708b2bcec9958d/searx/search.py#L105-L106

```python
    if hasattr(engine, 'proxies'):
        request_args['proxies'] = next(engine.proxies)
```

PR #1996, opened 2020-06-12
Update google.py

K-mikaZ:
because results of `eval_xpath` can be:
```
- Cerca de 65 600 000 resultados
- About 373 000 000 results
- Результатов: примерно 22 300 000
- تقریباً ۴٬۹۲۰٬۰۰۰ نتیجه
- ...
```
`isdigit()` is more suitable for finding numbers, I think.

### Please consider only the second, I made a typo in the first.

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -226,8 +226,9 @@ def response(resp):
     if instant_answer:
         results.append({'answer': u' '.join(instant_answer)})
     try:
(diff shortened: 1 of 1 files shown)


return42:
Its already in https://github.com/asciimoo/searx/pull/1985 see https://github.com/asciimoo/searx/pull/1985/commits/0e2a8d0773a6d4508f8777f337eec086a7dfcb87

PR #2006, opened 2020-06-16
Consider HTTP request when running search categories on select is ena…

### LINKED ISSUE #1138: Preserve URL param preferences when selecting categories
At the moment, specifying the GET method pref via a param (e.g. https://searx.instance/?method=GET&q=blah) causes searx to continue to use GET when for example clicking through to page 2 (https://searx.instance/?category_general=1&q=blah&pageno=2&time_range=None&language=en-US).

However, clicking on a category, e.g. "Images", causes it to revert to POST (or at least the URL reverts to https://searx.instance/). Instead this should work the same as pagination, and result in a URL of e.g. https://searx.instance/?method=GET&categories=images&q=blah

kvch:
Previously, if GET method was selected in the query URL, and you clicked on a category, the UI sent a POST request. Now if GET is selected, the URL is preserved.

Closes #1138

### DIFF
--- a/searx/static/plugins/js/search_on_category_select.js
+++ b/searx/static/plugins/js/search_on_category_select.js
@@ -6,19 +6,37 @@ $(document).ready(function() {
             });
             $(document.getElementById($(this).attr("for"))).prop('checked', true);
             if($('#q').val()) {
+                if (getHttpRequest() == "GET") {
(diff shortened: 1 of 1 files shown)


dalf (dismissed):
Oscar theme: it works as expected
Simple theme: not affected.

asciimoo (approved):
Great, thanks!

PR #2027, opened 2020-06-26
Created new plugin type custom_results. Added new plugin bang_redirect

lukasvdberk:
**TLDR** 

- Created a new plugin type of custom_results
   - Documented the new plugin type and usage in plugins.rst
- Created a new plugin called "Bangs redirect". This will redirect the user directly to the site (with the query) instead of showing the results on the results page. Also included a extra config for this plugin (containing 7.5k sites)
- Created unit tests for my "Bang redirect" plugin. 
 
**Detailed version**
I really like Searx and have only been using it for 2 weeks. Coming from DuckDuckGo I really liked their implementation of bangs and wanted to recreate that for Searx. Their implementation is when you enter a bang it redirects directly to the site with your submitted query. Maybe I have missed something but I could not find that for Searx. I also wanted more bangs out of the box. Thats why I used by own bangs config data (located in searx > plugins > bangs_data > bangs.json). 

For all of this to work I created a new plugin type called custom_results (which is documented and anyone can use ). With this you can return a custom flask return type so you can for example redirect (like I used) or show a custom page. It runs when a user submitted a query (just like post_search) and will cancel the rest of the usual search flow when it does not return None. 

I also added unit_tests in tests > unit > test_plugins > SelfBangTest. This will test my Bang redirect plugin for all of its features. The tests ran successfully on my machine with python 3.7

I retrieved my bangs data from the following [project](https://github.com/jivesearch/jivesearch) , [the data file](https://raw.githubusercontent.com/jivesearch/jivesearch/master/bangs/bangs.toml) (I converted the TOML file to a JSON file). 

If you want to see all of this working you can checkout my instance at [https://searx.lukashisprojects.rocks/](https://searx.lukashisprojects.rocks/). Make sure to enable the plugin in the preferences section. 

**Personal note**
I am still a cs-student and this is my first opensource contribution. I really love this project and I hope I have made a useful contribution. If there is anything I can do to make this PR work I am willing to go the extra mile.


abcdan:
Interesting, really like the idea.

return42:
Thanks! .. before we add new plugins https://github.com/asciimoo/searx/pull/1938 should be merged. The `call_with_results` should  be implemented in searx. BTW I want to link https://github.com/asciimoo/searx/issues/1146

lukasvdberk:
@return42 Thanks for the quick response! I guess I will just wait before making other changes when #1938 is merged. If I can anything else to better fit this plugin mention me.

searx/plugins/bangs.py
```diff
@@ -0,0 +1,140 @@
...
+def search_bangs(raw_search_query):
+    """
+    Searches if the bang is available.
+    :param raw_search_query: The search query the user providied
+    :return: Return a list of dicts with all the bangs data (coming from bangs_data/bangs.json)
+    """
+    user_bang = get_bang_from_query(raw_search_query)
+    # Searches for the bang matching the query
+    return list(filter(
+        lambda bang: user_bang in bang["triggers"], _get_bangs_data()
```
dalf: It would be faster if ```bangs_data``` is dictionary, with the different triggers as a key. The different triggers for the same bang would reference the same structure, so the memory footprint would be the same. All this could be prepare in the ```_get_bangs_data``` function.

dalf:

I would like to keep some architecture design :
* keep the core components decouple from the web server.
* avoid parsing the query in different part of the code.

But it raises the question: should this new feature be a plugin or part of the core ?

---

Def:
* "external bang": bang as in duckduckgo
* bang: bang as in searx.

Here a try, it is only a draft / some ideas / request for comments.

The first step is to parse the "external bang" query in query.py, and store the "external bang" in a query.external_bang attribute. So this function must be changed:
https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/query.py#L51
(I know that is difficult to parse the external bangs without the "triggers", except if the syntax is "!!engine", here the question is more "which syntax to use ?", then make the code works).

Then, the redirect URL can be stored in the ResultContainer class:
https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/results.py#L125-L140
Add a new field ```redirects = None```.

The plugin can be a pre_search plugin. search.py can skip the normal search if a pre_search plugin returns False:
https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/search.py#L533-L534

What the plugin could look like (not tested & unfinished):
```python
def pre_search(request, search):
	if search.external_bang:
		search.result_container.redirect = get_redirect_from_bang(search.query.external_bang, search.query.languages, search.query.getSearchQuery())
	# return False if there is a redirect, so the normal search is skipped (see https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/search.py#L533-L534 )	
	return search.result_container.redirect is None


def get_redirect_from_external_bang(external_bang, languages, query):
	bang_info = _get_bangs_data().get(external_bang, None)
	if bang_info is not None:
		# TODO add region support.
		bang_url = bang["regions"]["default"]
		return bang_url.replace("{{{term}}}", query)
	return None


def _get_bangs_data():
	...
```

There is no need to change code here:
https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/webapp.py#L560-L567

but the result_container.redirect must be checked, so before
https://github.com/asciimoo/searx/blob/bce3830b8dd9322e4e93d0003aa86e03c03e2dca/searx/webapp.py#L575-L580

add something like:
```
if result_container.redirect:
	# see note below
	return redirect(result_container.redirect)
```

----

About redirect: searx is privacy metasearch engine. A redirect breaks this because a typo can land the user somewhere else.
But at the same time, I understand that it is convenient.
What about a new preference ```allow_redirect``` (in the preferences page):
* ```no redirect```, default value, the URL is displayed, but there is no redirect.
* ```proxify```: redirect to a proxified version of the URL (using [Morty](https://github.com/asciimoo/morty) or the internal proxy)
* ```redirect```: redirect (same behavior than ddg).

searx/plugins/bangs.py
lukasvdberk: @dalf Thanks for the feedback! I think that is something easy to change and I will try to implement it.

lukasvdberk:
> @dalf  But it raises the question: should this new feature be a plugin or part of the core ?

So far as I can see it is a highly requested feature so that it would become part of the core is not a bad idea. The idea/code you drafted with adding a redirect to the search results is a way cleaner approach then what I proposed. But maybe it is a better idea to enhance the plugin manager with maybe web (flask) specific stuff? In that case people can make easily more plugins without changing the core and the original idea of searx can be kept. And with that approach make it so people can easily add external plugins that are not part of the core (dont kow how but we can think of something). Just my input. (apologies for my bad English)

return42:
> so people can easily add external plugins that are not part of the core

Therefore I linked  https://github.com/asciimoo/searx/pull/1938

> So far as I can see it is a highly requested feature 

Yes, there are some interested in .. the subject is discussed in https://github.com/asciimoo/searx/issues/1146#issuecomment-358454371

dalf:
> But maybe it is a better idea to enhance the plugin manager with maybe web (flask) specific stuff?

I guess the wish is to be able to provide a HTTP response directly into the plugin ?
But right now the plugin doesn't manage the json / csv / rss output formats (even if a public instance disables these output formats).
See https://github.com/asciimoo/searx/pull/2027/files#diff-ce4e881f09c0d7e047393021798f4be7R63

It would make sense to display help about the "external bangs" next to the "answer" syntax: 
image

But at least one plugin already modify the searx syntax:
image

lukasvdberk:
> I guess the wish is to be able to provide a HTTP response directly into the plugin ?

Yes. Thats what I was kinda hoping to implement with the custom_results plugin (but could probably be done in a cleaner way). But maybe its better to put it on a separate plugin page or something and in separate plugin folder like web_plugins? @dalf

searx/plugins/bangs.py
lukasvdberk: @dalf I updated it to your method which I think is way better.

lukasvdberk:
> Therefore I linked #1938

 That seems like we are on the same idea. Nice! (I did read #1938  it the first time you mentioned it but forgot it to include it in my response my bad)

asciimoo:
Great feature, I really appreciate it!

>  I am still a cs-student and this is my first opensource contribution. I really love this project and I hope I have made a useful contribution. If there is anything I can do to make this PR work I am willing to go the extra mile.

Awesome, thanks a lot. I hope this is just the beginning of your contributions to the free software community (and to searx) =)

> But it raises the question: should this new feature be a plugin or part of the core ?

Good question @dalf. As I see, there isn't much benefit of implementing this feature as a plugin, it could be part of the core.

> I know that is difficult to parse the external bangs without the "triggers", except if the syntax is "!!engine",

I would definitely go with the `!!` syntax to be able to distinguish these "external bangs" from the current ones. Not only just because it's easier to implement, but there is a privacy aspect too what we have to consider.
These external bangs don't provide any privacy protection. It is more like a convenience/comfort feature. We also have to emphasize this in the documentation, but perhaps using `!!` for the external bangs can help to highlight the difference between the two types of bangs, too.

@lukasvdberk what is your opinion? Could you rewrite it as a core feature?
@dalf provided great advices about how to implement it to the core, but feel free to ask if there is anything unclear.

return42:
In an PN I discussed with @dalf my 5 cent, why I think these _external bangs_ should be a part of the searx core:

One reason therefore I see is the same base of bang names in all instances.  If there are different name-sets out there, we are going into babylonian confusion about the bang names.

> I would definitely go with the !! syntax ...

I also vote for the !! syntax.  Please take also into account, that in the searx syntax, you can use the _engine-bang_ multiple times, but you can only use one _external-bang_ (compare `!go !yh foo` vs `!!wiki foo`) / If both (external- and engine-bangs) are given in the query term, which should win?

lukasvdberk:
>  but you can only use one external-bang (compare !go !yh foo vs !!wiki foo) / If both (external- and engine-bangs) are given in the query term, which should win?

I think the !! should override the ! but I am reallly curious what you guys think.

> @lukasvdberk what is your opinion? Could you rewrite it as a core feature?

I would love to take a shot at it and will try my best (but @dalf made it really easy with his outlining of how it could work so really thanks for that :)). 

> Awesome, thanks a lot. I hope this is just the beginning of your contributions to the free software community (and to searx) =)

Hopefully so this is so much fun! 👍 (and it means a lot when people like you (@asciimoo) , @dalf and @return42 say it and give feedback!)

lukasvdberk:
I refactored the bangs plugin with the suggestions of @dalf . The bang operator now is !! instead of !
(its still a plugin but I think this is the best way to implement it for now but idea's are always welcome )

There is one issue with a rss unit test not passing but I am not sure if that is because of my refactor or that someone changed the rss function. 

This is the core of the implementation of the bangs plugin but I have some other ideas to make this better, these are the following:

1. Suggestions for bangs you could use (if the user query is github then the suggestions !gh will come up in the suggestion results box)
2. Extensive documentation of how to use the bang plugin and where you can add/remove bangs.
3. A page where you can see all the bangs available and search through them (where and how I have no idea).

I would love your guys input on these idea's and what you think of the current code of the plugin. (thanks in advance already).

abcdan:
It looks pretty good to me, I've also tried to figure out why it fails but can't find why (maybe due to my lack of unit testing skills).

asciimoo:
It looks good indeed, however it is still a plugin. If understand correctly, we agreed on it should be a core feature. @lukasvdberk do you need any help to rewrite it as a native feature instead of a plugin?

lukasvdberk:
@asciimoo Oh sorry I though you guys saw this implementation like a core implementation (that the bang redirect plugin was decoupled from the webserver). I could still rewrite it but I think a plugin has the advantage that it can be easily enabled/disabled and people know where to look. If it is a core feature should it be enabled by default and that you have no setting page to turn it off/on (just like the standard ! bang engine). Sorry for seeing it wrong I sometimes have problems with interpiting.

I have no problem rewriting it as a core feature like the ! engine but I just understood it wrong sorry.

asciimoo:
>  think a plugin has the advantage that it can be easily enabled/disabled and people know where to look

What is the advantage of the ability of completely turning off this feature? If someone doesn't want to use it they simply don't search for `!!` bangs. On/off is already controlled in the search query. In my opinion it is highly unlikely that someone wants to search for a reserved bang expression and I can't see other cases when turning off this feature would be useful.

> Sorry for seeing it wrong I sometimes have problems with interpiting.

No worries, it can happen (often). Sometimes it's hard to understand each other by just few written sentences.

lukasvdberk:
> What is the advantage of the ability of completely turning off this feature? If someone doesn't want to use it they simply don't search for !! bangs. On/off is already controlled in the search query. In my opinion it is highly unlikely that someone wants to search for a reserved bang expression and I can't see other cases when turning off this feature would be useful.

That makes a lot of sense. Im going to start rewritting it as a core feature like the ! engine. 

> No worries, it can happen (often). Sometimes it's hard to understand each other by just few written sentences.

Thanks for the understanding!

asciimoo:
Awesome, thanks! Perhaps you can check the hints given by @dalf. Those can help a lot implementing this as a core feature.

lukasvdberk:
I refactored the bangs plugin to the core with the suggestions of @dalf . It works like a charm for me. 

I also refactored my unit tests to webapp to test if it correctly redirects.The format unit tests for webapp (to format search results to json, csv, rss) are failing but I don't know why. Do you guys know if this is a issue with some other PR maybe or did I do something wrong? (the result formats do work in the browser tho)

What is still missing is documentation but I think I will start writing that when I have a clear go that this will be the way its going to work.

Looking forward to what you guys think :+1:

docs/dev/plugins.rst
```diff
@@ -9,6 +9,7 @@ Plugins
    - :ref:`plugins generic`
 
 Plugins can extend or replace functionality of various components of searx.
+To create a plugin put your python file in searx > plugins.
```
dalf: No need to modify to plugins.rst since it is now a core feature.

lukasvdberk: Oh yeah good point! Totally forgot that I documented that. Will remove

lukasvdberk: Fixed it!

searx/settings.yml
```diff
@@ -908,4 +908,4 @@ doi_resolvers :
   doai.io  : 'https://doai.io/'
   sci-hub.tw : 'https://sci-hub.tw/'
 
-default_doi_resolver : 'oadoi.org'
+default_doi_resolver : 'oadoi.org'
```
dalf: nitpicking: could you add the new line ? So settings.yml don't get involved in the PR.

searx/plugins/__init__.py
```diff
@@ -29,7 +29,8 @@
                            self_info,
                            search_on_category_select,
                            tracker_url_remover,
-                           vim_hotkeys)
+                           vim_hotkeys,
+                           )
```
dalf: nitpicking: could you revert the small modification ? Some other PR are also modifying this file.

searx/search.py
```diff
@@ -103,7 +107,8 @@ def search_one_offline_request(engine, query, request_params):
 
 def search_one_request_safe(engine_name, query, request_params, result_container, start_time, timeout_limit):
     if engines[engine_name].offline:
-        return search_one_offline_request_safe(engine_name, query, request_params, result_container, start_time, timeout_limit)  # noqa
+        return search_one_offline_request_safe(engine_name, query, request_params, result_container, start_time,
```
dalf: is it possible to remove the new line here ?

searx/plugins/__init__.py
lukasvdberk: Done!

searx/settings.yml
lukasvdberk: Done!

searx/utils.py
```diff
@@ -465,3 +464,10 @@ def get_xpath(xpath_str):
 def eval_xpath(element, xpath_str):
     xpath = get_xpath(xpath_str)
     return xpath(element)
+
+
+def get_external_bang_operator():
+    """
+    :return: Returns the external bang operator used in Searx.
+    """
+    return "!!"
```
dalf: I think this constant ```!!``` can move to query.py. Except if you see some further development, there is no need to declare a function ?

searx/plugins/__init__.py
lukasvdberk: Done!

searx/utils.py
lukasvdberk: I was not entirely sure where to put this but I currently use this in external_bangs.py and in query.py. I find hard coding strings a bad practice. But maybe external_bang.py is a better location for it?

searx/search.py
lukasvdberk: Done!

searx/query.py
```diff
@@ -23,6 +23,7 @@
 )
```
dalf: I know this file lack some comments.

A more simpler implementation (feel free to change it if you think it is better in another way, or it doesn't work).

Before this line https://github.com/asciimoo/searx/blob/99435381a84072b110c32004b2fb778af9b96f77/searx/query.py#L123
Add this
```python
            # external bang
            if len(query_part) > 2 and query_part[0:1] == '!!':
				self.external_bang = query_part[2:]
				parse_next = True
				# continue otherwise the external bang will be parsed as a searx bang.
				continue
```

Remove _isExternalBang and _parseExternalBang from RawTextQuery,

In external_bang.py, replace ```_get_bang_query(query)``` by ```query.getSearchQuery()```. The good thing about is this
```!ddg :de searx``` will search for "searx" not for ":de searx".

File: tests/unit/test_plugins.py [hunk 8]
```diff
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
-
 from searx.testing import SearxTestCase
 from searx import plugins
 from mock import Mock
 
+from searx.webapp import app
```
dalf: is this import useful ?

lukasvdberk: Not useful so removed it.

searx/query.py
lukasvdberk: > In external_bang.py, replace _get_bang_query(query) by query.getSearchQuery(). The good thing about is this
!ddg :de searx will search for "searx" not for ":de searx".

I was unable to do this because the method getSearchQuery() does not exist in either query or the search variable. The rest I did implement and is working.

searx/utils.py
lukasvdberk: After all the refactors the method was not needed anymore so it is removed.

searx/external_bang.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def _get_bang(user_bang):
+    """
+    Searches if the supplied user bang is available. Returns None if not found.
+    :param user_bang: The parsed user bang. For example yt
+    :return: Returns a dict with bangs data (check bangs_data.json for the structure)
+    """
+    try:
+        return _get_bangs_data()[user_bang]
```
dalf: can be:
```python
return _get_bangs_data().get(user_bang)
```
without try ... except

searx/external_bang.py
```diff
@@ -0,0 +1,81 @@
...
+    slitted_raw_query = raw_query.split(" ")
+
+    full_query = ""
+    # Cause the first one is the bang like &yt
+    for raw_query in slitted_raw_query[1:]:
+        full_query += raw_query + "%20"
+    return full_query
+
+
+# Dont use this variable directly but access via _get_bangs_data
```
asciimoo: Why don't use the variable directly? In the end this variable is being used anyway since `_get_bang_data()` returns this variable.

searx/external_bang.py
```diff
@@ -0,0 +1,81 @@
...
+bangs_data = None
+
+
+def _get_bangs_data():
```
asciimoo: a bit more compact and pythonish definition of this section:
```python
bangs_data = {}
with open(join(searx_dir, 'data/bangs.json')) as json_file:
   for bang in json.load(json_file)['bang']:
      for trigger in bang["triggers"]:
         bangs_data[trigger] = {x:y for x,y in bang.items() if x != "triggers"}
```
No need to create a loader function, this way `bangs_data` is populated immediately after starting the application, then we can use it freely.

searx/external_bang.py
```diff
@@ -0,0 +1,81 @@
...
+    return None
+
+
+def _get_bang(user_bang):
+    """
+    Searches if the supplied user bang is available. Returns None if not found.
+    :param user_bang: The parsed user bang. For example yt
+    :return: Returns a dict with bangs data (check bangs_data.json for the structure)
+    """
+    try:
```
asciimoo: Here you can simply write `return _get_bangs_data().get(user_bang)`. `dict().get(key)` returns `None` if the given key doesn't found.

asciimoo (changes requested):
Overall it's a good PR. I've added some minor notes, please check it.

searx/query.py
dalf: Here a patch:
```patch
diff --git a/searx/external_bang.py b/searx/external_bang.py
index 098ae53f..7e1f48a6 100644
--- a/searx/external_bang.py
+++ b/searx/external_bang.py
@@ -16,42 +16,22 @@ def get_bang_url(search):
     """
 
     if search.external_bang:
-        query = str(search.query, "utf-8")
-        bang = _get_bang(search.external_bang)
-
-        if bang:
+        external_bang = _get_bang(search.external_bang)
+        if external_bang:
+            query = str(search.query, "utf-8")
             # TODO add region support.
-            bang_url = bang["regions"]["default"]
-
-            return bang_url.replace("{{{term}}}", _get_bang_query(query))
+            bang_url = external_bang["regions"]["default"]
+            return bang_url.replace("{{{term}}}", query)
     return None
 
 
-def _get_bang(user_bang):
+def _get_bang(external_bang):
     """
     Searches if the supplied user bang is available. Returns None if not found.
     :param user_bang: The parsed user bang. For example yt
     :return: Returns a dict with bangs data (check bangs_data.json for the structure)
     """
-    try:
-        return _get_bangs_data()[user_bang]
-    except KeyError:
-        return None
-
-
-def _get_bang_query(raw_query):
-    """
-    Gets the actual user search.
-    :param raw_query: The raw user query coming from the browser
-    :return: For example with the query &yt yes yes then this function will return yes yes.
-    """
-    slitted_raw_query = raw_query.split(" ")
-
-    full_query = ""
-    # Cause the first one is the bang like &yt
-    for raw_query in slitted_raw_query[1:]:
-        full_query += raw_query + "%20"
-    return full_query
+    return _get_bangs_data().get(external_bang)
 
 
 # Dont use this variable directly but access via _get_bangs_data
```

searx/external_bang.py
lukasvdberk: true that.

searx/external_bang.py
lukasvdberk: way simpler!

searx/external_bang.py
lukasvdberk: I prefer getters and setter but with your example below it is alternative way (your parser is way nicer/cleaner tho!)

searx/external_bang.py
lukasvdberk: Fixed it!

searx/external_bang.py
lukasvdberk: Fixed it!

searx/external_bang.py
lukasvdberk: Fixed it!

searx/external_bang.py
lukasvdberk: Fixed it!

searx/query.py
lukasvdberk: I refactored its cleaner now. I thought the %20 was necessary but it was not I guess. Thanks for the patch! (and rest of the feedback)

lukasvdberk:
@dalf @asciimoo 

The feedback you gave was really valuable and I have implemented it all (or commented).
Is it a good idea that you resolve the conversations (if it is off course all correct)? (or comment back). 

Really thanks :+1:

searx/external_bang.py
lukasvdberk: >          bangs_data[trigger] = {x:y for x,y in bang.items() if x != "triggers"}
Wow I did not know you could do that in 1 line btw. Learned something new thanks!

asciimoo:
Tests are failing because the `result_container` in the webapp unit tests is mocked: https://github.com/asciimoo/searx/blob/master/tests/unit/test_webapp.py#L50 . You have to add the new `redirect_url=None` attribute to the mock object which will solve the currently failing tests, but the new test you added will fail because `redirect_url` is `None` in the mock object.
Probably https://github.com/asciimoo/searx/blob/master/tests/unit/test_search.py would be the proper place to test this new feature.
Could you fix the webapp tests by extending the `result_container` mock object and could you move the new test to `test_search.py`?
Lastly, a commit [fixup](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) would be appreciated as a final touch to keep the git log clean and I think after that it's ready to merge.

lukasvdberk:
> Tests are failing because the result_container in the webapp unit tests is mocked: https://github.com/asciimoo/searx/blob/master/tests/unit/test_webapp.py#L50 .

Oh got it! Added the mock all tests are passing now. Thanks for the tip!

>  Probably https://github.com/asciimoo/searx/blob/master/tests/unit/test_search.py would be the proper place to test this new feature.

I agree and I refactored it.

>  Lastly, a commit fixup would be appreciated as a final touch to keep the git log clean and I think after that it's ready to merge.

I did not know this existed but I tried to implement it(don't know If I did it correctly). The you link sended for fixup was helpful for the explanation but I did not got it working. I followed this [video](https://youtu.be/2E23I9PzplM) maybe is good to add to reference in future PR's maybe? @asciimoo

abcdan:
Looks pretty good to me!

searx/external_bang.py
```diff
@@ -0,0 +1,43 @@
...
+# NOTE only use the get_bang_url
+
+bangs_data = {}
+with open(join(searx_dir, 'data/bangs.json')) as json_file:
+    for bang in json.load(json_file)['bang']:
+        for trigger in bang["triggers"]:
+            bangs_data[trigger] = {x: y for x, y in bang.items() if x != "triggers"}
+
+
+def get_bang_url(search):
```
dalf: A tiny thing: search should be renamed search_query to avoid confusion about the type of the object.

lukasvdberk: Oh good point will add that!

lukasvdberk: Fixed it!

asciimoo (approved):
Looks good to me, thanks!

lukasvdberk:
> Looks good to me, thanks!

Im really grateful to have worked with you guys @dalf and @asciimoo ! You provided good feedback and gave really good insights of how the code should be structured to made this PR work (learned some stuff to apply to my own projects). 

One last thing. Should I add documentation and if so what should I include?
I am thinking of the usage of the external bang where to find the external bangs and maybe how to add your own bang.

return42:
I can write the documentation .. I want to auto generate the list of bangs in the docs, this might be to complicate if this is the first time you are using the sphinx framework (reST).

lukasvdberk:
@return42 If you want to do that would be awesome, since my English and writting skills are not so great..... But if you have any questions I am always available to help!

return42:
> @return42 If you want to do that would be awesome, since my English and writting skills are not so great.

@lukasvdberk believe me, it is better than mine ;)

To document the bangs I send PR #2046.  You might have been noticed that I also opened issue #2045 in which I call for a significant reduction of bangs (redirect URLs) .. your thoughts and remarks are welcome / thanks!

Issue #2057, opened 2020-07-10
http 429 Internal server error


joshu9h:
Just a fresh install on Ubuntu server 20-04
Gives after the installation guide a http 429 on http://ipadrres/searx
Searx in debug gave the following error in the searx.log

Fri Jul 10 16:07:15 2020 - *** no app loaded. going in full dynamic mode ***
Traceback (most recent call last):
  File "/usr/local/searx/searx-src/searx/webapp.py", line 74, in <module>
    from searx.search import SearchWithPlugins, get_search_query_from_webapp
  File "/usr/local/searx/searx-src/searx/search.py", line 32, in <module>
    from searx.external_bang import get_bang_url
  File "/usr/local/searx/searx-src/searx/external_bang.py", line 13, in <module>
    for bang in json.load(json_file)['bang']:
  File "/usr/lib/python3.8/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib/python3.8/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 31341: ordinal not in range(128)
Fri Jul 10 16:07:15 2020 - unable to load app 0 (mountpoint='') (callable not found or import error)
Fri Jul 10 16:07:15 2020 - *** no app loaded. going in full dynamic mode ***

joshu9h:
If i go back in time with git checkout c21220c6714475cdbf247eba4046ba735d9d402a 2July it works fine
When i do git checkout 4829a76aae8fa6af8d3f814ff780dc6e792b84e5 Created new plugin type custom_results. Added new plugin bang_redirect (  i got the internal server error looks something goes wrong with the bangs

dalf:
I don't know how you installed your searx instance, but it is probably related to https://github.com/asciimoo/searx/pull/2043
```
LANG=C.UTF-8
LANGUAGE=C.UTF-8
LC_ALL=C.UTF-8
```

joshu9h:
I have changed line12 external_bang.py

searx/searx/external_bang.py

Line 12 in 4829a76
 with open(join(searx_dir, 'data/bangs.json')) as json_file: 
with open(join(searx_dir, 'data/bangs.json'), encoding='utf-8') as json_file:

now it works fine thanks Dalf

dalf:
I close the issue since the problem is solved.

PR #2043, opened 2020-07-04
 [fix] external_bang - UnicodeDecodeError: 'ascii' codec can't decode

return42:
Python's default encoding depends on the platform, set (python) default encoding
UTF-8 in uwsgi ini files:

    LANG=C.UTF-8
    LANGUAGE=C.UTF-8
    LC_ALL=C.UTF-8

Error pattern:

    Traceback (most recent call last):
      File "/usr/local/searx/searx-src/searx/webapp.py", line 74, in <module>
        from searx.search import SearchWithPlugins, get_search_query_from_webapp
      File "/usr/local/searx/searx-src/searx/search.py", line 32, in <module>
        from searx.external_bang import get_bang_url
      File "/usr/local/searx/searx-src/searx/external_bang.py", line 13, in <module>
        for bang in json.load(json_file)['bang']:
      File "/usr/lib/python3.8/json/__init__.py", line 293, in load
        return loads(fp.read(),
      File "/usr/lib/python3.8/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
      UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 31341: ordinal not in range(128)

close: #2041

Signed-off-by: Markus Heiser <markus.heiser@darmarit.de>

Issue #2041, opened 2020-07-04
Running "python searx/webapp.py" works, running via uwsgi doesn't


tjharman:
Hi,

I'm new to this, so I'm probably doing something really dumb. Apologies in advance if I'm time wasting.

I've installed Searx in a Ubuntu 20.4 Container (running under Proxmox)

All's gone well, except if I try to access my site, I get an Internal Server Error.

Looking into the /var/log/uwsgi/app/searx.log file I see

```
Sat Jul  4 17:19:29 2020 - setgid() to 998
Sat Jul  4 17:19:29 2020 - setuid() to 998
Sat Jul  4 17:19:29 2020 - Python version: 3.8.2 (default, Apr 27 2020, 15:53:34)  [GCC 9.3.0]
Sat Jul  4 17:19:29 2020 - PEP 405 virtualenv detected: /usr/local/searx/searx-pyenv
Sat Jul  4 17:19:29 2020 - Set PythonHome to /usr/local/searx/searx-pyenv
Sat Jul  4 17:19:29 2020 - Python main interpreter initialized at 0x55595f83fc40
Sat Jul  4 17:19:29 2020 - python threads support enabled
Sat Jul  4 17:19:29 2020 - your server socket listen backlog is limited to 100 connections
Sat Jul  4 17:19:29 2020 - your mercy for graceful operations on workers is 60 seconds
Sat Jul  4 17:19:29 2020 - mapped 218760 bytes (213 KB) for 2 cores
Sat Jul  4 17:19:29 2020 - *** Operational MODE: preforking ***
Sat Jul  4 17:19:29 2020 - added /usr/local/searx/searx-src/ to pythonpath.
Sat Jul  4 17:19:29 2020 - spawned uWSGI master process (pid: 24971)
Sat Jul  4 17:19:29 2020 - spawned uWSGI worker 1 (pid: 24977, cores: 1)
Sat Jul  4 17:19:29 2020 - spawned uWSGI worker 2 (pid: 24978, cores: 1)
Sat Jul  4 17:19:29 2020 - spawned uWSGI http 1 (pid: 24979)
Traceback (most recent call last):
  File "/usr/local/searx/searx-src/searx/webapp.py", line 74, in <module>
    from searx.search import SearchWithPlugins, get_search_query_from_webapp
  File "/usr/local/searx/searx-src/searx/search.py", line 32, in <module>
    from searx.external_bang import get_bang_url
[...]

PR #2059, opened 2020-07-11
[mod] external_bang.py: always read bangs.json with a UTF-8 file.

dalf:
## What does this PR do?

Make sure that ```searx/data/bangs.json``` is always read as an UTF-8 file.

The PR #2043 makes sure the locale is set to UTF-8 using the [installation scripts](https://asciimoo.github.io/searx/admin/installation.html#installation-scripts).

Unfortunately searx is not always install this way. 
We can either document the UTF-8 requirement: https://github.com/asciimoo/searx/pull/2043#issuecomment-653869595
However, the external files are read as UTF-8 files except bangs.json.

So this PR may not solve the locale setting requirement, it makes sure searx starts whatever the locale is (see the related issues).

1. Set the locale something different than UTF-8 and C.
2. starts searx
3. check that searx doesn't crash.
4. check an external bang is working as expected.

## Author's checklist

N/A

## Related issues

Related to the PR #2043, and issues #2041, #2057

### DIFF
--- a/searx/external_bang.py
+++ b/searx/external_bang.py
@@ -9,7 +9,7 @@ from searx import searx_dir
 # NOTE only use the get_bang_url
 
 bangs_data = {}
-with open(join(searx_dir, 'data/bangs.json')) as json_file:
+with open(join(searx_dir, 'data/bangs.json'), encoding='utf-8') as json_file:
     for bang in json.load(json_file)['bang']:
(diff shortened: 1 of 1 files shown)


idanoo (approved):
Confirms this resolves the issue for me.

return42 (approved):
Thanks!

PR #1878, opened 2020-03-02
Add plugin to allow filtering of domains by whether they run on green infrastructure

mrchrisadams:
This PR adds checking against every search engine against a local snapshot of the Green Web Foundation database we make available as the url2green dataset at the link below:

https://www.thegreenwebfoundation.org/green-web-datasets/

This allow us to return only results that are running on internet infrastructure known to be running on green/renewable power.

It also provides a fallback to the API when we can't find a local database.

By default this is set not to default to looking up urls, as a privacy safeguard.

PR #2074, opened 2020-07-14
Add external plugin support

asciimoo:
This is a second proposal to accomplish plugin decoupling. I think #1938 is highly complicated and does much more than this feature requires, so here is an alternative implementation for the same feature. Please review it and let me know your opinion.

This solution supports the use of any kind of standard python modules which implements the required attributes of a plugin, so new plugins can be installed by standard python tools (pip/setup.py).

Downsides:
 - Localization of plugins name/description isn't possible
 - Plugins have to be updated manually

## What does this PR do?

Implements external plugin extensibility.

Makes us able to decouple plugins from searx.

Can be tested with https://github.com/asciimoo/searx_external_plugin_example .

## Related issues

#1938 #1716 #1878


searx/plugins/__init__.py
```diff
@@ -77,6 +86,60 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+        pkg.__base_path = dirname(abspath(pkg.__file__))
+
+        fix_package_resources(pkg, name)
+
+        plugins.append(pkg)
+        logger.debug('plugin "{0}" loaded'.format(name))
+    return plugins
+
+
+def check_resource(base_path, resource_path, name, dir_prefix):
```
dalf: I guess a symbolic link doesn't work ?
The copy can be a problem: if the install folder is read-only, the copy won't work. More over, if there are different versions, the old files won't deleted (if the names change).

I don't know if [blueprint](https://flask.palletsprojects.com/en/1.1.x/blueprints/#static-files) can be a solution ? According to https://stackoverflow.com/questions/9513072/more-than-one-static-path-in-local-flask-instance :

```python
blueprint = Blueprint('site', __name__, static_url_path='/static/site', static_folder='path/to/files')
app.register_blueprint(blueprint)
```

```jinja
{{ url_for('site.static', filename='filename') }}
```

It could even be a way to include templates (I think about answer like "color picker" that display an actual color picker).

return42:
Before I start my review just some _bikeshedding_

- should we add [python entry points](https://github.com/asciimoo/searx/issues/1768)?

long-distance goals:

- should we add [SPDX](https://github.com/asciimoo/searx/issues/1801)?
- does it make sense to add pylint checks gradually?

I quote a post I made somewhere ..

> My experience is that changes such as code quality (pep8 vs pylint) and the Py3 changeover take years. This can only be achieved (especially in open source projects) by continually changing the code that is being patched (bugfixed) anyway in line with the long-distance goals. This ensures that the code has been tested sufficiently even after the changeover. You do this until the "old" has grown out of the sources and in 5 or 10 years you will reache your long distance target.

searx/plugins/__init__.py
return42: About blueprint I want to refer my comment: [as far as I can see Theme selection is not possible with Flask-Blueprints. Therefore I first refrain from using blueprints.](https://github.com/asciimoo/searx/pull/1938#issuecomment-618424236)

asciimoo: > I guess a symbolic link doesn't work ?
> The copy can be a problem

What would be the advantage of using symlinks?

> the old files won't deleted (if the names change).

@dalf, you are right, it would be better to create a folder for each plugin and check files in the reverse direction too.

> I don't know if blueprint can be a solution ?

Blueprints won't work if you serve your static files with your production webserver - which is recommended in bigger instances.

asciimoo:
> should we add python entry points?

What would be the advantage compared to this solution?

> should we add SPDX?

Why?

> does it make sense to add pylint checks gradually?

I wouldn't do that. These are external components maintained by the community, I would let them write their code as they want.

return42 (dismissed):
LGTM / successfully tested with ..

````bash
$ make install
(py3) $ pip install git+https://github.com/asciimoo/searx_external_plugin_example.git
````
register py modules as searx-plugin

````diff
modified   searx/settings.yml
@@ -60,10 +60,9 @@ outgoing: # communication with search engines
 # External plugin configuration
 # See http://asciimoo.github.io/searx/dev/plugins.html for more details
 #
-# plugins:
-#   - plugin1
-#   - plugin2
-#   - ...
+plugins:
+  - plugin1
+  - only_show_green_results
 
 engines:
   - name: apk mirror
````

File: searx/plugins/__init__.py (hunk 1 above)
Comment (dalf) [reply searx/plugins/__init__.py]: > Blueprints won't work if you serve your static files with your production webserver - which is recommended in bigger instances.

> What would be the advantage of using symlinks?

If searx maintains a static folder made of symbolic links:
* Flask / starlette whatever can serve the files
* nginx / caddy can serve the files too.

In this folder, there is:
* for each plugin, one symbolic link to the static folder.
* (same for the themes ?)
* a symbolic link to the core components.

It is a "temporary" folder outside the searx code base: the git repository / distrib package stays clean.

(Just an idea)

It can be related to https://github.com/asciimoo/searx/issues/2034 : the build process takes all the files, rename with a hash in the name, so the HTTP server can be configure to have a very long expiration date. It works well only if the plugin are included in the process.

**Comment by return42** [comment]
> I wouldn't do that. These are external components maintained by the community, I would let them write their code as they want.

OK, then lets drop pep8 tests (quality gate)! .. The pep8 we use has discontinued [pycodestyle (formerly called pep8)](https://pypi.org/project/pycodestyle/2.2.0/) and is more a hostage than a help.

I really vote for pylint test (quality gate) / pylint finds most of your daily typos and fails.

File: searx/plugins/__init__.py (hunk 1 above)
Comment (return42) [reply searx/plugins/__init__.py]: > It is a "temporary" folder outside the searx code base: the git repository / distrib package stays clean.
> (Just an idea)

I like this idea ..

Comment (asciimoo) [reply searx/plugins/__init__.py]: @dalf this sounds good, one disadvantage of symlinks is that you have to configure your webserver to follow them. It isn't a big deal. In case of nginx it is just a `disable_symlinks off;` line in the config, but we have to emphasize it in the documentation at least, however it still can be confusing to somebody who installed searx before this change and tries to add an external plugin.
Also I'm thinking about security impacts of using symlinks. So far I've found these issues:
 - Directory listing with symlinks can be dangerous
 - https://serverfault.com/questions/244592/followsymlinks-on-apache-why-is-it-a-security-risk/244612
 - https://blog.hboeck.de/archives/873-The-tricky-security-issue-with-FollowSymLinks-and-Apache.html
 - https://magento.stackexchange.com/questions/35243/are-symlinks-on-templates-a-security-problem-and-if-yes-why

The last one mentions an interesting issue. It isn't symlink specific, but if a malicious external plugin says that its resource is `/etc/passwd` then we make it accessible from the web. I should check this scenario too.

Copying files feels a bit more straight forward solution, however symlinks could have advantages over file duplication as @dalf mentioned. It's hard to decide.. =) @dalf / devs what is your opinion?

> It is a "temporary" folder outside the searx code base: the git repository / distrib package stays clean.

I'm not sure if I'm right, but as far as I know the symlink also "contaminates" the distro packages.

Comment (return42) [reply searx/plugins/__init__.py]: >  but if a malicious external plugin says that its resource is /etc/passwd then we make it accessible from the web. I should check this scenario too.

A malicious external plugin can do everything the searx process can do / there are no special _access controls_ for plugins.

> Also I'm thinking about security impacts of using symlinks .. Copying files feels a bit more straight forward solution, however symlinks could have advantages over file duplication  ...

The more I think about it, I prefer a solution with (only one) `static/` folder where all needed files are _synced_ in.  ATM this folder is our [searx/static](https://github.com/asciimoo/searx/tree/master/searx/static).

This has some drawbacks, e.g. resources from external plugins are copied into the [searx/static](https://github.com/asciimoo/searx/tree/master/searx/static) folder of the repository.

OT: #1938 iterates over [static files and sync them](https://github.com/return42/searx/blob/ep-plugins/searx/resources.py#L111) / triggered by [webapp.py](https://github.com/return42/searx/blob/ep-plugins/searx/webapp.py#L117) calling [get_plugins()](https://github.com/return42/searx/blob/ep-plugins/searx/plugins/__init__.py#L78)

> It can be related to #2034 : the

I vote for a _temporary folder_ which can be loaded (copy files in) by the searx / on the other side a _static-URL_ is needed for this folder under which a HTTP server can export the static files.  The _temporary folder_ is loaded by webapp.py (_writeable_) and _readable_ by the HTTP server process.. The  _temporary folder_ on the file system should be outside of the searx repository.  For developer purpose (and may be the default) we can create the _temporary folder_ in the the git working tree and add them to the `.gitignore`.

Comment (asciimoo) [reply searx/plugins/__init__.py]: > A malicious external plugin can do everything the searx process can do / there are no special access controls for plugins.

Thats true, you are right.

**Comment by asciimoo** [comment]
Resource handling is updated, it uses dedicated directories for every external plugin to be able to remove unused resource files.

File: searx/plugins/__init__.py [hunk 2]
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
         return ret
 
 
+def load_external_plugins(plugin_names):
+    plugins = []
+    for name in plugin_names:
+        logger.debug('loading plugin: {0}'.format(name))
+        try:
+            pkg = import_module(name)
+        except Exception as e:
```
return42, line 96: What kind of exception do you are expecting here?  I mean, if there is a syntax error (or any other error) in the imported module .. should we really catch it?  I guess you had some experience where exceptions has been thrown you wanted to catch .. what kind of exception did you have seen?

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+    return plugins
+
+
+def check_resource(base_path, resource_path, name, target_dir, plugin_dir):
+        dep_path = join(base_path, resource_path)
+        file_name = basename(dep_path)
+        resource_path = join(target_dir, file_name)
+        if not exists(resource_path) or sha_sum(dep_path) != sha_sum(resource_path):
+            try:
+                copyfile(dep_path, resource_path)
```
return42: I guess its better to explicit  catch [OSError](https://docs.python.org/3/library/exceptions.html#OSError) / compare [Exception hierarchy](https://docs.python.org/3/library/exceptions.html#exception-hierarchy).

Its the same with the other OS related try/except blocks.

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+            except:
+                logger.critical('failed to copy plugin resource {0} for plugin {1}'.format(file_name, name))
+                exit(3)
+
+        # returning with the web path of the resource
+        return join('plugins', plugin_dir, file_name)
+
+
+def fix_package_resources(pkg, name):
```
return42: `prepare_pkg_resources` or similar might be a better function name?

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+        pkg.__base_path = dirname(abspath(pkg.__file__))
+
+        fix_package_resources(pkg, name)
+
+        plugins.append(pkg)
+        logger.debug('plugin "{0}" loaded'.format(name))
+    return plugins
+
+
+def check_resource(base_path, resource_path, name, target_dir, plugin_dir):
```
return42: For me this function is more an _sync_ or a _push_ than a _check_

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+        dep_path = join(base_path, resource_path)
+        file_name = basename(dep_path)
+        resource_path = join(target_dir, file_name)
+        if not exists(resource_path) or sha_sum(dep_path) != sha_sum(resource_path):
```
return42: Sorry for  question, but what is the intention of the sha_sum .. performance? I mean, reading both files completely into the memory to see that the sha is different and than copy the file from the resource to the destination (which includes another read) is this really faster?  .. wouldn't it be better to just compare the timestamps on the _right_ and _left_ hand?

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+
+        fix_package_resources(pkg, name)
+
+        plugins.append(pkg)
+        logger.debug('plugin "{0}" loaded'.format(name))
+    return plugins
+
+
+def check_resource(base_path, resource_path, name, target_dir, plugin_dir):
+        dep_path = join(base_path, resource_path)
```
return42: indentation is 8 instead 4 space

return42:
I made a lot remarks but didn't find the time to implement an alternative .. may be tomorrow I have more time /sorry.

searx/plugins/__init__.py
```diff
@@ -77,6 +87,81 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+        # returning with the web path of the resource
+        return join('plugins', plugin_dir, file_name)
+
+
+def fix_package_resources(pkg, name):
+    plugin_dir = 'plugin_' + name
+    target_dir = join(static_path, 'plugins', plugin_dir)
+    if not isdir(target_dir):
+        try:
+            mkdir(target_dir)
```
return42: I guess `os.makedirs(target_dir, exist_ok=True)` is more robust.

searx/plugins/__init__.py
```diff
@@ -86,3 +171,6 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
 plugins.register(search_on_category_select)
 plugins.register(tracker_url_remover)
 plugins.register(vim_hotkeys)
+# load external plugins
+if 'plugins' in settings:
+    plugins.register(*settings['plugins'], external=True)
```
return42, line 179: Be aware: this creates folders and copies files when importing the `plugins` package (`__init__`).

searx/plugins/__init__.py
dalf: I would copy atime_ns and mtime_ns, so the weak ETags (generated by the HTTP server) do not change.

searx/plugins/__init__.py
```diff
@@ -54,7 +62,9 @@ def __iter__(self):
         for plugin in self.plugins:
             yield plugin
 
-    def register(self, *plugins):
+    def register(self, *plugins, external=False):
+        if external:
+            plugins = load_external_plugins(plugins)
```
return42, line 67: check of the mandatory names (`required_attrs`) is missed for external plugins

searx/__init__.py
```diff
@@ -30,6 +30,7 @@
 
 searx_dir = abspath(dirname(__file__))
 engine_dir = dirname(realpath(__file__))
+static_path = abspath(join(dirname(__file__), 'static'))
```
return42, line 33: Missed support of  [custom static path](https://github.com/asciimoo/searx/blob/master/searx/settings.yml#L21)

return42:
> @dalf, you are right, it would be better to create a folder for each plugin and check files in the reverse direction too.

@asciimoo I picked up the concept in  #2096

searx/plugins/__init__.py
asciimoo: > What kind of exception do you are expecting here? 

Import error

searx/plugins/__init__.py
asciimoo: This happens only once at startup, don't think that the performance difference really matters, but checking timestamps is another good way to achieve similar functionality

searx/plugins/__init__.py
asciimoo: Why? It happens later in this function.

asciimoo:
@return42 thanks for reviewing. It's good that conceptually there is nothing wrong with this PR from your side. I'm fixing the minor issues you found.

searx/plugins/__init__.py
asciimoo: > I guess its better to explicit catch OSError / compare Exception hierarchy.

Why?

return42: Explicit catch what you can handle / never catch exceptions you do not handle.

kvch (dismissed):

return42:
Is it only me? .. I enabled (and installed) the `plugin1` ...

````
$ make run
DEBUG:searx.plugins:loading plugin: plugin1
  File "./searx/webapp.py", line 74, in <module>
    from searx.search import SearchWithPlugins, get_search_query_from_webapp
  File "./searx/searx/search.py", line 37, in <module>
    from searx.plugins import plugins
  File "./searx/searx/plugins/__init__.py", line 175, in <module>
    plugins.register(*settings['plugins'], external=True)
  File "./searx/searx/plugins/__init__.py", line 67, in register
    plugins = load_external_plugins(plugins)
  File "./searx/searx/plugins/__init__.py", line 102, in load_external_plugins
    prepare_package_resources(pkg, name)
  File "./searx/searx/plugins/__init__.py", line 148, in prepare_package_resources
    for f in listdir(target_dir):
NameError: name 'listdir' is not defined
````

Code analyzer pylint (if you do not want to install the `plugin1`) spits out :

     ./searx/searx/plugins/__init__.py:148: [E0602(undefined-variable), prepare_package_resources] Undefined variable 'listdir'

FWIW it also spits out the following warnings (just to demonstrate where liniting can help) ..

````
./searx/searx/plugins/__init__.py:23: [W0622(redefined-builtin), ] Redefining built-in 'exit'
./searx/searx/plugins/__init__.py:65: [W0621(redefined-outer-name), PluginStore.register] Redefining name 'plugins' from outer scope (line 164)
./searx/searx/plugins/__init__.py:91: [W0621(redefined-outer-name), load_external_plugins] Redefining name 'plugins' from outer scope (line 164)
./searx/searx/plugins/__init__.py:96: [W0703(broad-except), load_external_plugins] Catching too general exception Exception
./searx/searx/plugins/__init__.py:100: [W0212(protected-access), load_external_plugins] Access to a protected member __base_path of a client class
./searx/searx/plugins/__init__.py:116: [W0702(bare-except), sync_resource] No exception type(s) specified
./searx/searx/plugins/__init__.py:129: [W0702(bare-except), prepare_package_resources] No exception type(s) specified
./searx/searx/plugins/__init__.py:138: [W0212(protected-access), prepare_package_resources] Access to a protected member __base_path of a client class
./searx/searx/plugins/__init__.py:144: [W0212(protected-access), prepare_package_resources] Access to a protected member __base_path of a client class
./searx/searx/plugins/__init__.py:153: [W0702(bare-except), prepare_package_resources] No exception type(s) specified
./searx/searx/plugins/__init__.py:160: [W0622(redefined-builtin), sha_sum] Redefining built-in 'bytes'
````
update: I added the import for `listdir` / seems to work now

File: searx/plugins/__init__.py [hunk 12]
```diff
@@ -77,6 +87,80 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+
+
+def sync_resource(base_path, resource_path, name, target_dir, plugin_dir):
+    dep_path = join(base_path, resource_path)
+    file_name = basename(dep_path)
+    resource_path = join(target_dir, file_name)
+    if not exists(resource_path) or sha_sum(dep_path) != sha_sum(resource_path):
+        try:
+            copyfile(dep_path, resource_path)
+        except:
```
return42, line 120: If we do not catch exceptions, they go upwards and the effect is the same, the application exits .. by example, if you drop the try/catch block you get:

````
FileNotFoundError: [Errno 2] No such file or directory: '/800GBPCIex4/share/searx/local/py3/lib/python3.8/site-packages/plugin1/resources/bg.js'
````

File: searx/plugins/__init__.py (hunk 2 above)
Comment (return42) [reply searx/plugins/__init__.py:96]: I vote for catching explicit exception we can handle or not to catch at all / by example, when we catch:

     CRITICAL:searx.plugins:failed to load plugin module plugin1: name 'y' is not defined

and if we do not catch:

````
  File "./searx/webapp.py", line 74, in <module>
    from searx.search import SearchWithPlugins, get_search_query_from_webapp
  File "./searx/searx/search.py", line 37, in <module>
    from searx.plugins import plugins
  File "./searx/searx/plugins/__init__.py", line 166, in <module>
    plugins.register(*settings['plugins'], external=True)
  File "./searx/searx/plugins/__init__.py", line 67, in register
    plugins = load_external_plugins(plugins)
  File "./searx/searx/plugins/__init__.py", line 94, in load_external_plugins
    pkg = import_module(name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "./searx/local/py3/lib/python3.8/site-packages/plugin1/__init__.py", line 9, in <module>
    x=y
NameError: name 'y' is not defined
````

There could be many reasons why a import fails, bring it to the top so developers have a meaningful error message.

Here is another example which might be worse to catch ModuleNotFoundError ...

````
  File "./searx/webapp.py", line 74, in <module>
    from searx.search import SearchWithPlugins, get_search_query_from_webapp
...
    pkg = import_module(name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named 'plugin2'
````

File: searx/plugins/__init__.py [hunk 13]
```diff
@@ -77,6 +87,80 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+        ])
+
+    for f in listdir(target_dir):
+        if basename(f) not in resources:
+            resource_path = join(target_dir, basename(f))
+            try:
+                remove(resource_path)
+            except:
+                logger.critical('failed to remove unused resource file {0} for plugin {1}'.format(resource_path, name))
+                exit(3)
```
return42, line 159: FYI: [sys.exit()](https://docs.python.org/3/library/sys.html#sys.exit) exits by raising a [SystemExit](https://docs.python.org/3/library/exceptions.html#SystemExit) .. which can be catched by the outer block :) .. sometimes it make sense to raise such exception (calling sys.exit()). Itis like the `die` command which might know some of us from perl.  In general this exception remains for the application code on the top, libraries (modules) should pass exceptions upwards.

asciimoo:
> Is it only me? .. I enabled (and installed) the plugin1 ...

Thanks for noticing, I've messed up the imports last night. Now it is fixed.

return42:
> Now it is fixed.

Commit d3f5fbf looks good to me..   The _unresolved_  remarks from mine are more _conceptual_ (non KO criteria).

dalf:
I don't know if this is the right place to talk about this: what about the update procedure ?

For example: if an external plugin includes a list of domain name to block, the list will most probably change over time at a different speed than searx core components.

Is ```pip install -U some_external_plugin``` enough ?

asciimoo:
> Is pip install -U some_external_plugin enough ?

Yup, that's all what is needed. I added this topic to the "downsides" section of the PR description, admins have to update external plugins independently from searx.

asciimoo:
I also added 164e472 to keep git clean. Is there anything else required to be able to merge this PR?

searx/plugins/__init__.py
```diff
@@ -77,6 +87,80 @@ def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
...
+    return plugins
+
+
+def sync_resource(base_path, resource_path, name, target_dir, plugin_dir):
+    dep_path = join(base_path, resource_path)
+    file_name = basename(dep_path)
+    resource_path = join(target_dir, file_name)
+    if not exists(resource_path) or sha_sum(dep_path) != sha_sum(resource_path):
+        try:
+            copyfile(dep_path, resource_path)
```
dalf, line 115: I would add:
```python
# copy atime_ns and mtime_ns, so the weak ETags (generated by
# the HTTP server) do not change.
dep_stat = os.stat(dep_path)
os.utime(resource_path, ns=(dep_stat.st_atime_ns, dep_stat.st_mtime_ns))
```

otherwise LGTM.

asciimoo: Great, thanks, mtime modification added.

dalf: It doesn't work as expected:
* mtime, atime are copied from the site-package directory.
* but pip  and/or setuptools don't copy these atributes.

dalf: pip:
* tar: OK(os.utime): https://github.com/pypa/pip/blob/8bf5731b84907bd1b41846334e98530bbe0cceaf/src/pip/_internal/utils/unpacking.py#L238
* zip: KO: https://github.com/pypa/pip/blob/8bf5731b84907bd1b41846334e98530bbe0cceaf/src/pip/_internal/utils/unpacking.py#L154
* https://github.com/python/cpython/blob/3.8/Lib/tarfile.py#L2281

setuptools:
* OK (shutil.copystat): https://github.com/pypa/setuptools/blob/990237c64404876791dc5985e7cfd260852dac0e/setuptools/archive_util.py#L88

So according to what I read:
* KO: pip install https://github.com/asciimoo/searx_external_plugin_example/archive/master.zip
* OK: pip install https://github.com/asciimoo/searx_external_plugin_example/archive/master.tar.gz <-- I don't know why it is KO.
* KO: in searx_external_plugin_example git clone (mtime)

The only consequence I see : it shows publicly when the plugin was installed (the reverse proxy will pick mtime).

asciimoo: > The only consequence I see : it shows publicly when the plugin was installed (the reverse proxy will pick mtime).

Why is this a problem?

dalf: It shows the installation time of the plugin, I don't think it is a problem.

dalf (approved):

PR #2079, opened 2020-07-16
Allow localization of "Self Informations"

KucharczykL:
## What does this PR do?

Allows the string "Self Informations" in the plugin of the same name to be localized by wrapping it with gettext().

Makes the translation consistent.

Pull the PR.


asciimoo (approved):
Thanks!

PR #2096, opened 2020-07-24
Add external plugin support

return42:
This is a third proposal to accomplish plugin decoupling.

This solution supports the use of any kind of standard python modules which implements the required attributes of a plugin, so new plugins can be installed by standard python tools (pip/setup.py). Hint: does not make use of [python entry points](https://github.com/asciimoo/searx/issues/1768)

**_Note: its a proposal .. the unit test not work right now._**

## What does this PR do?

Implements external plugin extensibility.

Makes us able to decouple plugins from searx.

````bash
$ cp -r searx/static /tmp/searx-static
$ make install
(py3) $ pip install git+https://github.com/asciimoo/searx_external_plugin_example.git
````
register py modules as searx-plugin

````diff
modified   searx/settings.yml
@@ -18,7 +18,7 @@ server:
     http_protocol_version : "1.0"  # 1.0 and 1.1 are supported
 
 ui:
-    static_path : "" # Custom static path - leave it blank if you didn't change
+    static_path : "/tmp/searx-static" # Custom static path - leave it blank if you didn't change
     templates_path : "" # Custom templates path - leave it blank if you didn't change
     default_theme : oscar # ui theme
     default_locale : "" # Default interface locale - leave blank to detect from browser information or use codes from the 'locales' config section
@@ -60,10 +60,9 @@ outgoing: # communication with search engines
 # External plugin configuration
 # See http://asciimoo.github.io/searx/dev/plugins.html for more details
 #
-# plugins:
-#   - plugin1
-#   - plugin2
-#   - ...
+plugins:
+  - plugin1
+  - only_show_green_results
````


asciimoo:
What is the purpose of this PR? What are the advantages of this solution? Also, it contains changes that doesn't related to this feature.

return42:
> What is the purpose of this PR?

It is an alternative implementation of your and @dalf concept (https://github.com/asciimoo/searx/pull/2074#issuecomment-663642988).

I made a lot of comments to your PR, so I thought I have to show how I would implement the concept of a folder where the plugins are synced in .. I started first on top of your PR but after a while I realized that nothing of your PR would reused in my implementation .. so I started on top of master.

>  What are the advantages of this solution?

- It harmonize the structure of internal and external plugins, to have one implementation for both ([provide_static_resources](https://github.com/return42/searx/blob/ext-plugins/searx/plugins/__init__.py#L153)).  

- It removes (plugin) code from the searx/static/plugins/  ... since I think we should distinguish between _build_ products and code (https://github.com/asciimoo/searx/issues/2034) / its one step in that direction.

- This implementation also does not catch exceptions without appropriate handling

- It does not start to modify the filesystem when a python module is imported (which is a horrible concept I see here the first time .. or did you ever seen file modification when you import modules from other projects or packages?

- It does not import the settings file / the setup of an application should not imported scattered around the modules / the application should read the setup once and pass it through it components. The application should use _factories_ (a simple get function is enough) to setup instances of the components according  to the applications config.

Things that ~~make my stomach pain~~ can be improved:

- all modules are backed together and can't used stand alone
- nearly all application setup is spread around the project and is done while importing modules
- no _factories_  / less separation of tasks
- heavy usage of globals / cross-relationships that are not obvious and are not documented [(1)](https://github.com/asciimoo/searx/pull/2038#discussion_r448267642) [(2)](https://github.com/asciimoo/searx/pull/2038#discussion_r448379890)
- mixing of [code and binaries](https://github.com/asciimoo/searx/tree/master/searx/static/themes/oscar)  / exporting files like package.json, .gitignore (and all the rest)  over HTTP is questionable / I have never seen before, did you?
- the separation of URLs paths and paths on the file system is not clear
- horrible code checker / if you have ever used pylint, you know what I mean .. give it a try and you will see how your performance and code quality increase!
- line size of more than 120 chars which is a horror when u are using terminals / who wants to read a news paper with only one column, scrolling the eyes from the left to the rigtht?  Python syntax was developed with the motivation t have _readable_ code.

>  Also, it contains changes that doesn't related to this feature.

What parts do you think of when you say that? Is it about the SPDX tag is it about the Makfeile, is it about code checking ..  how is the quality of a implementation measured? .. by minimize the count of diff lines or by minimize my stomach pain .. I usually choose to limit my stomach pain .. if it does not meets your expectations I need more words.

kvch:
@return42 I find @asciimoo's current PR simpler than this proposal. I think we should go with the least complicated solution and then later on adjust the functionality as we gather more feedback from users.

Also, as I mentioned before, we should separate refactoring from feature development. One can hide many things behind a refactoring, we might miss some issues with the original feature if the two are mixed together.

Now that is out of the way, I find completely unacceptable the way you talk about this project. You are a maintainer of searx, you should be aware that this project is the result of a 100+ people's work with different skill levels. You should treat every one of those people who gave their time and effort to searx with respect and empathy. Yes, this project has issues. Yes, some of the code could use some refactoring. But preparing a list titled "things that make my stomach pain" and collecting past discontents in an unrelated PR is not helping anyone.

return42:
@kvch I apologize if I have hurt anyone, it was not my intention. But I want to make it clear where I think things can be improved. 

Things continue to evolve, procedures are improved and I also have "stomach pain" when I look at my _implementations from yesterday_ ... What was perfect in the past may be solved differently today ..

The phrase "things that make my stomach pain" is harmless in my mother tongue, you can use it to start a discussion and draw attention to things ("headache" is much more harder, but this might different in other languages). This is how I start a discussion in my job .. and yes, customers pay me _to put the finger in the wound_ (which is another phrase in my mother tongue).

I would be happy if we could discuss the points of criticism and not so often bother about the tone. Look at the topics and overlook it if the tone does not seem appropriate to you .. for which I apologize.

----

> I think we should go with the least complicated solution and then later on adjust the functionality as we gather more feedback from users.

Even if we go with an other solution, why not make suggestions and discuss?

What exactly do you think is complicated in this implementation?


> Also, as I mentioned before, we should separate refactoring from feature development. 

To make things less complicated some files has been renamed (moved to a different folder) / separation of code and HTTP static.

kvch:
> @kvch I apologize if I have hurt anyone, it was not my intention. But I want to make it clear where I think things can be improved.

Apologies accepted. But I would be super happy if I would not have to call out it again in the future.

> Things continue to evolve, procedures are improved and I also have "stomach pain" when I look at my _implementations from yesterday_ ... What was perfect in the past may be solved differently today ..

I see. Also, have similar feelings about my work from time to time. However, I do not find it acceptable to talk about other people's work this way.
 
> The phrase "things that make my stomach pain" is harmless in my mother tongue, you can use it to start a discussion and draw attention to things ("headache" is much more harder, but this might different in other languages). This is how I start a discussion in my job .. and yes, customers pay me _to put the finger in the wound_ (which is another phrase in my mother tongue).

At my job, we are expected to talk to both users and colleagues with respect and empathy. ;) After all we are not only solving problems, but people's problems. These people have different backgrounds, feelings and opinions. Furthermore, we are working in international settings, it is best to avoid using language specific idiosyncrasies. (Or at least, if you are using something like that add more context to your sentences.) Your words might sound acceptable to you, but not everything translates well to other languages.

> I would be happy if we could discuss the points of criticism and not so often bother about the tone. Look at the topics and overlook it if the tone does not seem appropriate to you .. for which I apologize.

I think we should bother with the tone. We are adults who should be able to have a polite conversation with each other. 

> > I think we should go with the least complicated solution and then later on adjust the functionality as we gather more feedback from users.
> 
> Even if we go with an other solution, why not make suggestions and discuss?

There is no problem with providing different solutions. I simply expressed that this new implementation is not superior to the previously provided.

> What exactly do you think is complicated in this implementation?

This implementation seems simple as well, but the other one is __more__ simple. Also, that PR does not contain refactorings which is a big pro for that solution.

return42:
... thanks for accepting

> This implementation seems simple as well, but the other one is more simple.

Any remarks or ideas about the details? .. would be happy if we have a discussion about.

PR #2097, opened 2020-07-25
Filtron sample: Add missing comma

gardouille:
## What does this PR do?

Add missing commas in the filtron's sample configuration file.

Based on this sample configuration, Filtron can't start :
```
Cannot parse rules: invalid character '"' after object key:value pair
```

```
make docs-clean docs
```

### DIFF
--- a/docs/admin/filtron.rst
+++ b/docs/admin/filtron.rst
@@ -64,7 +64,7 @@ of:
                 "Param:q",
                 "Path=^(/|/search)$"
             ],
-            "interval": "<time-interval-in-sec (int)>"
+            "interval": "<time-interval-in-sec (int)>",
             "limit": "<max-request-number-in-interval (int)>",
             "subrules": [
                 {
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks!

gardouille:
Also, it's not so obvious (at least for me…), that `"<time-interval-in-sec (int)>"` has to be fully replaced (including the double quotes).

```
        "interval": <time-interval-in-sec (int)>,
        "limit": <max-request-number-in-interval (int)>,
```
Might be enought.

Any opinion @return42 ?

return42:
> Also, it's not so obvious (at least for me…), that "<time-interval-in-sec (int)>" has to be fully replaced (including the double quotes).

Yes, this is a compromise, if we do like you suggested, the syntax highlighter is no longer to highlight (this invalid) syntax .. you can test it in your local clone .. may we should add a [admonition](https://asciimoo.github.io/searx/dev/reST.html#admonitions) about the replacement?

OT tip:  try `make docs-live` to see your changes rendered [live](https://asciimoo.github.io/searx/dev/contribution_guide.html#id8)

return42:
FYI: I update gh-pages https://asciimoo.github.io/searx/admin/filtron.html#id3 / if you think, I should add a admonition about the replacement, leave a note / thanks.

PR #2102, opened 2020-07-27
Fix Wikipedia's paragraph extraction

### LINKED ISSUE #2090: Wikipedia crash on specific search query
Instance: searx.be
Browser: Chrome Version 84.0.4147.89

Hello,

Not sure whether this is a bug, problem with the instance itself , or my browser, but when I search for certain query e.g. `port dover`, I get an error message:
```
Engines cannot retrieve results:
wikipedia (unexpected crash 'NoneType' object has no attribute 'replace')
```

Image 6

This error doesn't happen if I change the language to Deutsch, Espanol, or other language. Seems like it only happens with English language.  

Image 2

Image 3

This error doesn't happen with other search query.

Image 4

I also test this on other searx instances and the results are mixed. 

No error:
https://searx.ninja
https://search.snopyta.org/
Both are still on 0.16.0

Same error:
https://searx.fmac.xyz
https://searx.xyz/

MarcAbonce:
## What does this PR do?

Fix Wikipedia paragraph extraction bug from #2090.

1. Search for `!wp Port Dover` or `!wp WSGI` and check if it works.
2. Search for a Wikipedia article with a disambiguation or some other paragraph at the top and verify that the infobox only displays the actual first paragraph.

## Author's checklist

I removed the `extract_first_paragraph` function because it's not needed anymore. The issue back then was that the API would return the disambiguation paragraph or any other notice as the first paragraph, but apparently this has already been fixed upstream in the Mediawiki API.

## Related issues

Closes #2090.

### DIFF
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -49,29 +49,6 @@ def request(query, params):
     return params
 
 
-# get first meaningful paragraph
-# this should filter out disambiguation pages and notes above first paragraph
-# "magic numbers" were obtained by fine tuning
-def extract_first_paragraph(content, title, image):
-    first_paragraph = None
-
-    failed_attempts = 0
-    for paragraph in content.split('\n'):
-
-        starts_with_title = paragraph.lower().find(title.lower(), 0, len(title) + 35)
-        length = len(paragraph)
-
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Great, thanks!

PR #2108, opened 2020-07-30
Change isAlive() to is_alive() as isAlive() is unsupported in newer P…

ashtasht:
…ython versions

## What does this PR do?

As you can see [here](https://docs.python.org/3.10/library/threading.html#threading.Thread.is_alive), `Thread.isAlive() `is no longer supported and `Thread.is_alive()` should be used.


kvch (approved):
Thank you!

PR #2113, opened 2020-07-31
make logicodev follow OS/browser dark/light mode

mikeri:
## What does this PR do?
Makes the default theme follow dark/light color scheme as set by OS/browser. Implemented as hree variants of the logico style; auto, dark and light. The default (auto) theme falls back to light mode if the setting is not supported in the OS and/or browser, that way this commit should not be disruptive.

It is not, it is merely a suggestion. But it's nice for users that has their OS of browser set to dark mode to have searx respect their preference by default.

Change the color scheme of your OS or browser and refresh searx.


julianfairfax:
Can we get an update on this? This is the one of the features I'm missing in searx that I'd like added, and since the work has already been done almost a year ago, it shouldn't be too difficult to do so.

julianfairfax (approved):
Not sure why the checks haven't been completed, but this is the one of the features I'm missing in searx that I'd like added, and the work has already been done. This shouldn't be too difficult to implement and the changes proposed seem reasonable.

PR #2115, opened 2020-08-02
Fix google images 'get image' button bug from issue #2103

### LINKED ISSUE #2103: Get image button from the Google Images engine doesn't give the raw URL to the image
When searching for images using the "Google Images" engine, the "Get image" button doesn't give the raw URL (that usually ends with `.jpg` or `.png`) to the image of every result.
In fact the button gives the same exact URL as the button "View source".

On other engines (like Bing images) it gives the raw URL to the image.

Here is a screenshot explaining the issue:
getimagegoogleimages

Here is a URL for testing the bug: https://darmarit.org/searx/?q=%21goi+ok&time_range=&language=en-US&category_images=on

@return42 Any thoughts about that?

qbitroot:
## What does this PR do?

Fix Google-images engine bug where clicking on 'Get image' button don't give original image URL (issue #2103).

This change is important because it allows Searx users to search full-resolution photos via Google-images. This is very convenient.

To test only Google images:
- Go to preferences of your Searx instance.
- Click "Engines" tab and choose "images" section.
- Disable all but google images.
1. Search something like 'wallpaper' and go to images section.
2. Click any image (check if image retrieved with google-images).
3. Press 'Get image' button.
4. See that you have full image URL now instead of page URL.

## Author's checklist

I created `scrap_img_by_id` function with arguments `script` and `data_id` where script is tag selected in line 151. That tag contains all full image URLs. `data_id` is id of image obtained with line 204. It is possible to get image URL with `data_id`, my function do it. Please see [screenshots from my comment](https://github.com/asciimoo/searx/issues/2103#issuecomment-667547267) at issue #2103 for more info.

## Related issues

Closes #2103


unixfox:
@return42 could you take a look at this PR please :)?

return42:
> @return42 could you take a look at this PR please :)?

I hope I have time next week / sorry for the delay.

searx/engines/google_images.py
```diff
@@ -77,6 +77,22 @@ def scrap_out_thumbs(dom):
...
+    """Get full image URL by data-id in parent element
+    """
+    img_url = ''
+    _script = script.split('\n')
+    for i, line in enumerate(_script):
+        if 'gstatic.com/images' in line and data_id in line:
+            try:
+                url_line = _script[i + 1]
+                img_url = url_line.split('"')[1]
+                img_url = img_url.replace(r'\u00', '%')
```
return42: @gangural thanks for your PR / static code analysis is not easy, I find this solution very good.  There is only one point I would improve: This replacement does not worked in some examples I found.

I implemented a patch on top of your branch: https://github.com/return42/searx/commit/63bb1d65616e2ddfb4f45b9c0c0bd395370d8e4f

You find a example in the commit message. If you think my patch fits better, pull it from my branch: https://github.com/return42/searx/tree/google-img-fix

Side note: in my patch I removed the try/except block since I do not had any exception  .. Did you had any?

return42:
> To test only Google images:

FYI: to select the engine you can use `!google-images` ([see](https://asciimoo.github.io/searx/user/search_syntax.html)) / e.g. in my tests I used:

    !google-images dali

searx/engines/google_images.py
qbitroot: I had exceptions while writing code. Now they are probably don't appear. About url-encoding, I thought that all web servers will interpret `%3d` as `=` but I was wrong.

qbitroot:
> to select the engine you can use `!google-images`

Oh, right, I had to use this.

unixfox:
I just tested locally your PR and it works great. Thank you for that!

return42 (approved):
Thanks!

return42:
@asciimoo for me this patch looks good .. if you have time please merge.

kvch (approved):

kvch:
@gangural Thank you for the fix! @return42 Thanks for the review! @unixfox Thanks for the testing!

kennedn:
Hi,

It looks like this merge has recently introduced an "Out of Index" bug when google image results are being parsed.

I am unsure if this is due to something changing on googles end. It appears that the passed 'script' variable in scrap_img_by_id can sometimes contain only 1 item in its list and the line specifically at https://github.com/searx/searx/blob/a8988885a5151a7fb152bff107dd43a26fa6b42e/searx/engines/google_images.py#L92 expects at least 2 items to exist so it can reference i + 1

This causes the following error in the logs:

```
  File "/usr/local/searx/searx-src/searx/search/processors/online.py", line 139, in search
    search_results = self._search_basic(query, params)
  File "/usr/local/searx/searx-src/searx/search/processors/online.py", line 123, in _search_basic
    return self.engine.response(response)  
  File "/usr/local/searx/searx-src/searx/engines/google_images.py", line 194, in response
    src_url = scrap_img_by_id(img_src_script, img_src_id)
  File "/usr/local/searx/searx-src/searx/engines/google_images.py", line 92, in scrap_img_by_id
    url_line = _script[i + 1]
IndexError: list index out of range
```

I have remediated this on my own instance by adding an additional check to https://github.com/searx/searx/blob/a8988885a5151a7fb152bff107dd43a26fa6b42e/searx/engines/google_images.py#L91

Which now reads
```python
if 'gstatic.com/images' in line and data_id in line and i + 1 < len(_script):
```

I no longer get an unexpected crash when performing an image search after amending this line.

PR #2116, opened 2020-08-02
Include author and video length in Invidious results

mikeri:
## What does this PR do?
It include author and video length in Invidious results

Users cannot see author or video length from invidious results without these changes

Perform an invidious search


asciimoo (approved):
Great, thanks!

PR #2125, opened 2020-08-07
Split autocompleter into (browser)suggestions and API

### LINKED ISSUE #2105: autocomplete in the browser URL doesn't work
Scenario:
* I pick one instance
* in the searx preferences, I select one autocomplete provider, then save
* in Firefox, I add the new search engine
* in the Firefox preferences, I select this new search engine as the default one
* in the Search bar, I type a query
* there is no autocomplete

I've replaced
https://github.com/asciimoo/searx/blob/fc5d1c69cc01ec1c44e5d2d9644269b6b737fa5a/searx/templates/__common__/opensearch.xml#L16
by
```html
<Url rel="suggestions" type="application/x-suggestions+json" template="{{ host }}autocompleter?q={searchTerms}"/>
```

but I'm not sure it works on http://localhost:8888

Nebukadneza:
## What does this PR do?
Since current browers do not seem to either POST requests at all, or
supplying a format-parameter as the current assumption is, the easiest
reasonably correct way seems to be to split the autocompleter endpoint
into:
- /autocompleter as used currently, continuing to work for the API-usage
  on the web-frondend
- /autocompleter_suggestions which is included in opensearch.xml defined
  for browser-suggestios-requests

While it is sad no get rid of POST for browser-autocompleter, it seems
neither chromium nor firefox currently support it — and opensearch
cannot reasonably specify it, as far as I understand. Thus removing it
seems fair for the time being.

Fixes: #2105

For users, suggestions can be vitally important. It is a integrated part of todays searching workflow.

- Start up searx using this branch
- Add it as search-engine via your browsers opensearch UI
- Enable it as a suggestion engine
- See it work or fail

## Author's checklist
- [ ] Do I understand opensearch.xml correct that POST suggestions do not work?
- [ ] Does this break any browsers I can’t try at the moment?
- [ ] Is having GET instead of POST a good idea?

## Related issues
Closes #2105

### DIFF
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -13,6 +13,6 @@
     </Url>
   {% endif %}
   {% if autocomplete %}
-    <Url rel="suggestions" type="application/json" template="{{ host }}autocompleter"/>
+    <Url rel="suggestions" type="application/json" template="{{ host }}autocompleter_suggestions?q={searchTerms}" method="get" />
   {% endif %}
 </OpenSearchDescription>
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -743,8 +743,7 @@ def about():
     )
 
 
-@app.route('/autocompleter', methods=['GET', 'POST'])
-def autocompleter():
+def _autocompleter(is_suggestion_query):
     """Return autocompleter results"""
 
     # set blocked engines
@@ -790,12 +789,24 @@ def autocompleter():
         results.append(raw_text_query.getFullQuery())
 
     # return autocompleter results
-    if request.form.get('format') == 'x-suggestions':
+    if is_suggestion_query:
         return Response(json.dumps([raw_text_query.query, results]),
                         mimetype='application/json')
 
     return Response(json.dumps(results),
                     mimetype='application/json')
+
+
+# TODO: It seems current browers only support suggestions via GET.
+# Re-evaluate opensearch-spec and browser behavior and add POST if possible!
+@app.route('/autocompleter_suggestions', methods=['GET'])
+def autocompleter_suggestions():
+    return _autocompleter(is_suggestion_query=True)
+
+
+@app.route('/autocompleter', methods=['GET', 'POST'])
+def autocompleter():
+    return _autocompleter(is_suggestion_query=False)
 
 
 @app.route('/preferences', methods=['GET', 'POST'])


qbitroot:
Oh, travis checks failed because of PEP8 (problems with code formatting and style, a common problem). A log was given:
```
TEST      pep8
. ./local/py3/bin/activate; pep8 --exclude='searx/static, searx/engines/gigablast.py' --max-line-length=120 --ignore "E402,W503" searx tests
searx/webapp.py:800:67: W291 trailing whitespace
searx/webapp.py:804:4: E111 indentation is not a multiple of four
searx/webapp.py:806:1: E302 expected 2 blank lines, found 1
searx/webapp.py:808:4: E111 indentation is not a multiple of four
searx/webapp.py:810:1: E302 expected 2 blank lines, found 1
Makefile:246: recipe for target 'test.pep8' failed
make: *** [test.pep8] Error 1
```

Nebukadneza:
Thanks for the ping, and sorry for not running pep8 locally first. Fixed …

Build look green now ^_^

dalf:
Thank you for your PR.

If only [this line](https://github.com/asciimoo/searx/pull/2125/files#diff-1a3127efb52515b63627f49aad3596c1R16) is changed does it solve the issue :
```
<Url rel="suggestions" type="application/json" template="{{ host }}autocompleter?format=x-suggestions&amp;q={searchTerms}" method="get" />
```
?

Nebukadneza:
Unfortunately not, as far as my testing went. I couldn’t figure out why, but when i add the format query-parameter, Firefox rejects the `opensearch.xml`, only stating it’s `invalid` with no further details (in the browser-console). :(

dalf:
> Unfortunately not, as far as my testing went. 

:(

Out of topic: how do you make sure that opensearch.xml is reloaded ? If I remove the engine from FF, add it again it seems the method settings stick to the previous value.

Nebukadneza:
For me, explicitly navigating to the `opensearch.xml` and force-reloading (ctrl+shift+r) did the trick …. It’s not a great workflow, though :|

msafadieh:
I was debugging this issue on my personal instance before seeing this PR. I ended up avoiding splitting routes by checking for `X-Requested-With: XMLHttpRequest`. Changing `opensearch.xml` to explicitly include the query param is still necessary though (currently using an FF extension to do that).

```
    # return autocompleter results
    if request.headers.get('X-Requested-With') == 'XMLHttpRequest':
        return Response(json.dumps(results),
                        mimetype='application/json')

    return Response(json.dumps([raw_text_query.query, results]),
                    mimetype='application/x-suggestions+json')

```
This was tested on FF79. Not sure if all browsers include that header when making a request.

dalf:
jquery adds the ```X-Requested-With: XMLHttpRequest``` header: https://api.jquery.com/jquery.ajax/ (see "headers" parameter), so it will work whatever the browser.

Nebukadneza:
Uuhh! Using `X-Requested-With` is a great idea, I didn’t think of that. I think your PR is superior, so I’m closing this one …

PR #2127, opened 2020-08-09
Use query params for browser autocomplete

### LINKED ISSUE #2105: autocomplete in the browser URL doesn't work

msafadieh:
## What does this PR do?

Sending query params over GET seems to be the only way to be able to
enable autocomplete in the browser. This commit adds the necessary URL
formatting to opensearch.xml. In order to identify queries coming from
the URL bar (rather than an AJAX request), which requires a different
JSON format and MIME type, the request headers are checked for
"X-Requested-With: XMLHttpRequest" which is added by jQuery request.

Fixes autocomplete

Remove old Searx, make sure autocomplete is enabled, make sure to clear any cache (or force-load opensearch.xml), add engine, test suggestions.

## Related issues

Fixes #2105
Closes #2125

### DIFF
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -13,6 +13,6 @@
     </Url>
   {% endif %}
   {% if autocomplete %}
-    <Url rel="suggestions" type="application/json" template="{{ host }}autocompleter"/>
+    <Url rel="suggestions" type="application/x-suggestions+json" template="{{ host }}autocompleter?q={searchTerms}"/>
   {% endif %}
 </OpenSearchDescription>
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -790,12 +790,12 @@ def autocompleter():
         results.append(raw_text_query.getFullQuery())
 
     # return autocompleter results
-    if request.form.get('format') == 'x-suggestions':
-        return Response(json.dumps([raw_text_query.query, results]),
+    if request.headers.get('X-Requested-With') == 'XMLHttpRequest':
+        return Response(json.dumps(results),
                         mimetype='application/json')
 
-    return Response(json.dumps(results),
-                    mimetype='application/json')
+    return Response(json.dumps([raw_text_query.query, results]),
+                    mimetype='application/x-suggestions+json')
 
 
 @app.route('/preferences', methods=['GET', 'POST'])


dalf:
@kvch , @asciimoo can you review this PR ? 
(no matter what I do, the autocomplete doesn't work on my setup).

msafadieh:
> @kvch , @asciimoo can you review this PR ?
> (no matter what I do, the autocomplete doesn't work on my setup).

I have autocomplete enabled on the server-side and not on the client-side, if that's any useful.

dalf:
> I have autocomplete enabled on the server-side and not on the client-side

What do you mean by "server-side" ?

My use case:
* I create a Firefox profile ( ```rm -Rf p1; mkdir p1; firefox --profile p1``` )
* go to http://localhost:8888 (<-- http, localhost is it a problem ?)
* set autocomplete to whatever that is not "-"
* save
* add the searx as a search engine
* set the default search engine to searx in the FF preferences.
* try to the auto-completion.
* the auto-completion doesn't work and there is no request according to the searx log

But I suspect there is something wrong with my use case (Mozilla Firefox 79.0, Ubuntu 18.04).

unixfox (approved):
Works for me on:
- Firefox nightly 81.0a1 (2020-08-03) (64-bit)
- Firefox stable 79.0 (64-bit)

I tested with a new Firefox profile on both browser.

dalf:
@unixfox just to be sure, you have try on http://localhost:8888/ ?

msafadieh:
> What do you mean by "server-side" ?

In the config file set by `SEARX_SETTINGS_PATH`, I set the following:

```
search:
    safe_search : 0
    autocomplete : "qwant"
```

unixfox:
@dalf 

Yes I did but I had to set the autocomplete directly into the `settings.yml` for it to work because if I reproduce the way you explained in https://github.com/asciimoo/searx/pull/2127#issuecomment-671075811 then it doesn't work.
~~I think it's expected because Firefox probably cache the opensearch.xml or something is wrong with the autocomplete setting in the preferences.~~
**EDIT**: Maybe the autocomplete setting in the preferences is just for the autocomplete on the home page of Searx and not for the autocomplete provided by the browser in the URL bar.

msafadieh:
Firefox's request to `/opensearch.xml` does not include the browser cookies. This is how I debugged it:

```
@app.route('/opensearch.xml', methods=['GET'])
def opensearch():
    method = 'post'

    print(list(request.cookies.keys()))
```

dalf:
It is the same setting:
* def render(...) :
   https://github.com/asciimoo/searx/blob/3c45fb7a99c09166c36001d442312363b436b6ac/searx/webapp.py#L387-L388
* ```def __init__(self, themes, categories, engines, plugins):```
   https://github.com/asciimoo/searx/blob/3c45fb7a99c09166c36001d442312363b436b6ac/searx/preferences.py#L336-L339

If it is set in the settings.yml, I confirm that is work as expected.

@msafadieh thank you for the PR and the debug about the cookies (and @Nebukadneza for the initial PR).

dalf (approved):

Issue #1801, opened 2020-01-09
SPDX License Identifier


return42:
We should stop C&P the license text into every file.  We have one https://github.com/asciimoo/searx/blob/master/LICENSE in the root folder and the source files should be tagged with a [SPDX License Identifier](https://spdx.org/using-spdx-license-identifier):

     SPDX-License-Identifier: AGPL-3.0-or-later

SPDX reduces redundant work by providing a common format for companies and communities to share important data about software licenses, copyrights, and security references, thereby streamlining and improving compliance.

gkaklas:
I recently found out that FSFE has developed a set of recommendations and a tool to help with licensing a project, [reuse](https://reuse.software/). Among other things, it seems like the tool [has a command](https://github.com/fsfe/reuse-tool/tree/0001e831665856d548a9fabdbd1ad37d38cb6df3#usage) that adds these headers in source files; just wanted to leave this here in case it seems useful!

PR #2128, opened 2020-08-09
Add command line engines: git grep, find, etc.

kvch:
## What does this PR do?

This PR adds new command line engines to searx. From now on users can execute searches using the command line tools of the host machine. The returned values are displayed using the `key-value` template of searx.

### New engines

1. **git grep**: This engine lets you grep in the searx repo. Example query: `!gg search`
2. **locate**: If `locate` is installed and initialized, you can search on the FS. Example query: `!loc search.py`
3. **find**: You can find files with a specific name from where you started searx. Example query: `!fnd search.py`
4. **pattern search in files**: This engine utilizes the command `fgrep`. Example query: `!fgr common`
5. **regex search in files**: This engine runs `grep` to find a file based on its contents. Example query: `!gr def search`

All of the engines above are commented out to make sure no one enables anything accidentally.

The commands or its **dependencies** included in these engines **must be installed by the instance admin**. Searx does not install any command on the host machine for you.

### The command engine

A new "base" engine called `command` is introduced. It is the foundation for all command line engines for now.

You can use this engine to create your own command line engine. The following options can be configured:

* `command`: A comma separated list of the elements of the command. A special token `{{QUERY}}` tells searx where to put the search terms of the user. Example: `['ls', '-l', '-h', '{{QUERY}}']`
* `delimiter`: A dict containing a delimiter `char` and the "titles" of each element in `keys`.
* `parse_regex`: A dict containing the regular expressions for each result key.
* `query_type`: The expected type of user search terms. Possible values: `path` and `enum`. `path` checks if the uesr provided path is inside the working directory. If not the query is not executed. `enum` is a list of allowed search terms. If the user submits something which is not included in the list, the query returns an error.
* `query_enum`: A list containing allowed search terms if `query_type` is set to `enum`.
* `working_dir`: The directory where the command has to be executed. Default: `.`
* `result_separator`: The character that separates results. Default: `\n`

For more examples you can consult the test of the engine.

This PR adds more possible use cases to searx. Private instances might become more powerful as admins/users can execute command line searches from the UI of searx.

Comment out the engines in `settings.yml`. Run various searches. For example try to get out of the configured working dir when the engine expects path search terms.


virtadpt:
A couple of suggestions:
- It would be handy if the Git repo in question could be specified by the user.  Use case: It would be possible to search a repository of, say, Markdown files constituting a personal wiki because PHP's search function leaves a bit to be desired.
- Very useful (!), but it seems that arbitrary users being able to search the `locate` database on a server (especially a personal one) would constitute a potential privacy leak.  Is it turned off by default?  Is there a warning of this in the documentation?
- Same with find(1).
- grep(1) and fgrep(1) - same.  But fantastic feature, love the idea.

searx/templates/oscar/result_templates/key-value.html
```diff
@@ -6,7 +6,7 @@
         {% continue %}
     {% endif %}
     <tr>
-        <td><b>{{ key|upper }}</b>: {{ value }}</td>
+        <td><b>{{ key|upper }}</b>: {{ value|truncate }}</td>
```
dalf, line 9: There is a similar file the simple theme: https://github.com/asciimoo/searx/blob/ccacd2f2c4cca2f813e4784a8eb601a796519f02/searx/templates/simple/result_templates/key-value.html#L7

searx/utils.py
```diff
@@ -465,3 +465,7 @@ def get_xpath(xpath_str):
 def eval_xpath(element, xpath_str):
     xpath = get_xpath(xpath_str)
     return xpath(element)
+
+
+def str_split(string, delimiter, count):
+    return string.split(delimiter, maxsplit=count)
```
dalf: Why not use string.split directly ?

kvch: It is a leftover. I've prepared the PR a few weeks ago and I forgot that we have dropped Python 2. :D I will fix it. Thanks for spotting the problem.

searx/engines/command.py
```diff
@@ -0,0 +1,182 @@
...
+    cmd = _get_command_to_run(query)
+    if not cmd:
+        return []
+
+    results = []
+    process = Popen(cmd, stdout=PIPE, stderr=PIPE)
+    reader_thread = Thread(target=_get_results_from_process, args=(results, process, params['pageno']))
+    reader_thread.start()
+
+    return_code = process.wait(timeout=timeout)
```
dalf: If after ```process.wait(....)``` the process is still alive, searx can start an huge number of processes.

I would change to something like:
```python
process = Popen(...)
try:
        reader_thread = Thread(...)
        try:
                reader_thread.start()
                return_code = process.wait(timeout=timeout)
                if return_code != 0:
                        raise ...
        finally:
                reader_thread.join(timeout=timeout)
                # more tricky: there is no clean way to kill a thread
finally:
        if process.is_alive():
            process.kill()
```

searx/engines/command.py
```diff
@@ -0,0 +1,182 @@
...
+    if delimiter:
+        elements = str_split(raw_result, delimiter['chars'], len(delimiter['keys']) - 1)
+        if len(elements) != len(delimiter['keys']):
+            return {}
+        for i in range(len(elements)):
+            result[delimiter['keys'][i]] = elements[i]
+
+    if parse_regex:
+        for result_key, regex in parse_regex.items():
+            found = re_search(regex, raw_result, flags=MULTILINE)
```
dalf: If the regex is compiled in the function init(...):
* it gives some speed improvement (not sure here it is relevant)
* if the regex has a syntax error, the developer doesn't have to make a search to see the error.

searx/templates/oscar/result_templates/key-value.html
kvch: Should we add `truncate` to that template as well? Or why do you mention it?

kvch:
@virtadpt Thank you for the feedback! The engines I have prepared a rather examples for how one can use the new feature. Ofc, these can be used as they are but these are way too general. 

The first use case you mention can be solved by the current implementation:

```yaml
  - name: personal wiki
    engine: command
    command: ['git', 'grep', '{{QUERY}}']
    shortcut: pw
    disabled: True
    working_dir: /path/to/your/wiki
    delimiter:
        chars: ':'
        keys: ['filepath', 'line']
```

Furthermore, you can make it private by configuring a token under `tokens`. WDYT?

As I mentioned all engines are commented out, so it cannot be enabled accidentally. However, I will add a warning to each engine in settings.yml. Also, users can restrict access to these engines by configuring the `tokens` attribute of the engine.

virtadpt:
@kvch That looks great!  It would totally work.  Only thing is (because I didn't know about `git grep` and tried it`) is that the command runs the output through the system's default pager.  Looking through the manpage for git-grep(1) the `core.pager` config option needs to be changed.  There's probably a way to do it with a command line switch.  Or maybe it's just my setup, I'm not sure..

kvch:
@virtadpt Alternatively you could set the `GIT_PAGER` environment variable. ATM it is not possible to pass env variables, but it seems like a reasonable addition. Tomorrow I will find the time to add it to the PR. Thank you again! :)

searx/templates/oscar/result_templates/key-value.html
dalf: > Should we add truncate to that template as well ?

Yes I guess the same constraint will apply to the content length.

searx/engines/command.py
```diff
@@ -0,0 +1,189 @@
...
+        try:
+            reader_thread.start()
+            return_code = process.wait(timeout=timeout)
+            if return_code != 0:
+                raise RuntimeError('non-zero return code when running command', cmd, return_code, process.stderr.read())
+        finally:
+            reader_thread.join(timeout=timeout)
+    finally:
+        if process.is_alive():
+            process.kill()
```
dalf: I'm not sure if that should be addressed:
* SIGKILL may leave some forked processes : https://stackoverflow.com/questions/21869242/killing-parent-process-along-with-child-process-using-sigkill (if ```cmd``` fork).
* https://stackoverflow.com/questions/32222681/how-to-kill-a-process-group-using-python-subprocess 
   * SIGHUP then SIGKILL might be better ?
   * or process group ?

virtadpt:
@kvch Even better!

dalf (approved):

searx/engines/command.py
```diff
@@ -0,0 +1,184 @@
...
+
+searx is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with searx. If not, see < http://www.gnu.org/licenses/ >.
+'''
+
```
return42, line 15: Just some minor nit-picking, the licences text can be replaced by (https://github.com/searx/searx/issues/1801):

    # SPDX-License-Identifier: AGPL-3.0-or-later

dalf: Yes ! there are a lot of other files to modified too, perhaps it is better to address this issue in another PR.

return42: This file is new created and I thought it is more straight forward to tag new files from the beginning.

searx/engines/command.py
```diff
@@ -0,0 +1,184 @@
...
+            _compiled_parse_regex[result_key] = re.compile(regex, flags=MULTILINE)
+    if 'delimiter' in engine_settings:
+        delimiter = engine_settings['delimiter']
+
+    if 'environment_variables' in engine_settings:
+        environment_variables = engine_settings['environment_variables']
+
+
+def search(query, params):
+    cmd = _get_command_to_run(query)
```
bAndie91, line 71: it takes the whole user input query if i'm not mistaken. so when i query eg. `!mylocalcmd xyz` (mylocalcmd being a custom command engine) then the given command gets the whole "`!mylocalcmd xyz`" as `{{QUERY}}` parameter, eventhough it should interested only in the user part of the query, ie. "xyz".
i found that it extracts the user part into the `user_query_parts` variable in `query.py` (current master commit 2fc3b17), but i can not see how to pass it to the engine.

PR #2130, opened 2020-08-09
Make default query submission method configurable

### LINKED ISSUE #2129: Allow server admins to choose default search method
Currently, the default search method used by Searx is `POST` which breaks compatibility with Firefox containers. Since FF's query to `/opensearch.xml` does not include the cookies, the user's preferred method is not reflected and they're forced create a custom search engine with the correct URL formatting and method.

A good solution is to make the method parameter configurable in `settings.yml`. This was already mentioned in #703 and looks fairly easy to implement. Let me know if you want me to open a PR.

msafadieh:
Very small PR :-)

## What does this PR do?

This commit allows admins to modify the default method on their side so they can set it to GET if needed.

Sending queries through POST, while better for privacy, breaks functionality with certain extensions (e.g. Firefox containers). Since Firefox does not send cookies when requesting `/opensearch.xml`, users cannot easily switch to GET on the client side unless they make a custom search engine.

Change `method` in your `settings.yml` file to `GET`.

## Related issues

Closes #2129

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -348,7 +348,7 @@ class Preferences(object):
                 }
             ),
             'method': EnumStringSetting(
-                'POST',
+                settings['server'].get('method', 'POST'),
                 choices=('GET', 'POST')
             ),
             'safesearch': MapSetting(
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Great, thanks!

PR #2131, opened 2020-08-10
Make infobox shorter by default and add an expand toggle

MarcAbonce:
## What does this PR do?

1. Make infoboxes shorter by default.
2. Hide the main image by default as well and set a maximum height even when expanded.
3. Add a toggle at the bottom of the infobox to expand it or to shrink it again.
    - This toggle works even without JS enabled.
    - Unfortunately, the only part I couldn't make without JS is hiding the toggle in the cases where all the content already fits into the shorter default infobox. This means that users with no JS may occasionally see an expand toggle even though there's nothing else to expand. But this is just an aesthetic problem, functionally it still works the same.
4. Some changes in the infobox's headers.
    - List the engines that the infobox extracted information from. This works the same way as with results and it's done for the same reason. Sometimes there's too many buttons at the end of the infobox and it's not clear which one is the source/engine.
    - Show full URL in the header and trim if too long. Only showing the domain led me to some weird results, for example a movie's official website can be the subpath of the studio's domain.
5. Add missing CSS class in Pointhi style.

Sometimes an infobox can get too large. This is not necessarily a problem for desktop users, but this could be specially annoying for mobile users who have to scroll a lot just to get to the first result.

* Try querying any engine that returns an infobox, such as Wikipedia, Wikidata, Duckduckgo Definitions or WolframAlpha.
    - For really long infoboxes, try popular movies with both Wikipedia and Wikidata.
    - For long portraits, try famous people with Wikidata.
    - For short infoboxes, try small towns with Wikipedia only or currency conversions in WolframAlpha
* Verify it looks and works fine in desktop and mobile.
* Verify it works the same way with either Pointhi or Logicodev styles.

## Author's checklist

* I'm pretty sure my toggle is currently not ARIA compliant, but I don't know what exactly am I supposed to add to make it accessible.
* I think we should extract some of the shared Less code for Oscar, because there's probably a lot more duplicated Less code between Logicodev and Pointhi. But I didn't think that's on the scope of this PR.


asciimoo (approved):
Great, thanks!

Issue #2069, opened 2020-07-13
opensearch.xml: add documentation about the GET/POST methods.


dalf:
https://github.com/asciimoo/searx/blob/6e37255e59693d7a3768491b2b8d06441051131f/searx/templates/__common__/opensearch.xml#L3 
If would be helpful to change this line to ```<ShortName>{{ instance_name }} ( {{opensearch_method}} )</ShortName>```.

It seems a lot of user get confused about the link between the "GET" / "POST" setting and the opensearch.xml behavior.

asciimoo:
I agree, it can be confusing, however I'm not sure that adding the method name here solves the problem. Perhaps better documentation of GET/POST methods would help. Also, there we could emphasize that this whole GET/POST switching nonsense is required because chrome cannot handle POST searches properly.

dalf:
> Perhaps better documentation of GET/POST methods would help.

When an user adds the search engine to the browser, it is not from the preferences page.

The users will either read the documentation because they went the preferences page, or either because the opensearch doesn't work as expected.

So I guess the idea is at least to add an entry about opensearch.xml into the documentation.

dalf:
After a second thought, I think it is more than documentation.

When the browser (at least Firefox) reads opensearch.xml, the cookies are not sent (see https://github.com/asciimoo/searx/pull/2127#issuecomment-671077893 ).
Consequence: the autocomplete setting doesn't work in opensearch.xml (except if it is hard coded in settings.yml).

dalf:
Related to the initial issue: I think an additional change would be helpful, since the cookies are not sent: the /opensearch.xml URL can change according to the settings.
* For example: ```/opensearch.xml?method=GET&autocomplete=duckduckgo```.
* I'm not sure if the query is sent, so an alternative approach can be ```/opensearch.get.duckduckgo.xml```.

The additional benefit: there is no need to erase / force reload to update the opensearch.

asciimoo:
@dalf you are right, probably your suggested method is the good way to handle the issue, especially because the missing cookie thing.

dalf:
With the [multi account container](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/) Firefox extension, the following use case doesn't work:
1. add a searx instance, use the POST method, make searx the default search engine.
1. in a page inside a container, select some text
1. right click "Search <selected text>"

The selected text won't be sent, the index page of searx is shown.

Related issue: https://github.com/mozilla/multi-account-containers/issues/1736

[...]

PR #2132, opened 2020-08-10
[fix] opensearch on chromium for android

### LINKED ISSUE #1666: Chromium-based browsers (Android) don't detect searx

unixfox:
## What does this PR do?

This PR set all the searches made by a Chrome/Chromium browser for Android to the "GET" method.

In order to have Searx added a search engine on Chrome/Chromium for Android, the browser wants you to visit the correct specific URL `/?q=yourkeyword`.

1. Run Searx locally
2. Install a chrome/chromium based browser on Android ([Bromite](https://www.bromite.org/) is a good choice).
3. Visit your Searx instance then make a search on it.
4. You should now see Searx in the "Recently visited" section of "Search engine" from the settings.
Screenshot_20200810-174631_Chromium_1_1

## Author's checklist

- I don't know how to allow the ability for the user to switch back to POST requests in the preferences. Maybe a developer that is used to the Searx code base could help me about that.
- I tested this PR on most of the popular browser and it doesn't break their opensearch discovery nor their search ability.
	- Microsoft Edge 15: [proof](https://i.imgur.com/vhCAkPF.png)
	- Microsoft Edge 84: [proof](https://i.imgur.com/EQE5C5c.png)
	- Google Chrome 84: [proof](https://i.imgur.com/4Mp81tZ.png)
	- Safari 13.1: [proof](https://i.imgur.com/pBVu1t9.png)
	- Firefox 79: [proof](https://i.imgur.com/xdS9O60.png)

## Related issues

Closes #1666

--------------------------

This PR is still a draft because I want to see if the maintainers and the contributors agree with my solution or if there is a better one.

### DIFF
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -6,7 +6,7 @@
   <Image>{{ urljoin(host, url_for('static', filename='img/favicon.png')) }}</Image>
   <LongName>searx metasearch</LongName>
   {% if opensearch_method == 'get' %}
-    <Url rel="results" type="text/html" method="get" template="{{ host }}search?q={searchTerms}"/>
+    <Url rel="results" type="text/html" method="get" template="{{ host }}?q={searchTerms}"/>
   {% else %}
     <Url rel="results" type="text/html" method="post" template="{{ host }}">
         <Param name="q" value="{searchTerms}" />
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -463,6 +463,9 @@ def pre_request():
     request.errors = []
 
     preferences = Preferences(themes, list(categories.keys()), engines, plugins)
+    user_agent = request.headers.get('User-Agent', '').lower()
+    if 'webkit' in user_agent and 'android' in user_agent:
+        preferences.key_value_settings['method'].value = 'GET'
     request.preferences = preferences
     try:
         preferences.parse_dict(request.cookies)


dalf:
I've tried:
* Chrome Desktop
* Chrome Android
* Firefox 79 desktop
* Firefox Android (build 2015757107)

All seems to work. 

---

I would move the check just after this line https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/webapp.py#L468
```python
    user_agent = request.headers.get('User-Agent', '').lower()
    if 'webkit' in user_agent and 'android' in user_agent:
        preferences.key_value_settings ['method'].value = 'GET'
```
(even if it seems to me to be a hack, the code should be in preferences.py)

---

> I don't know how to allow the ability for the user to switch back to POST requests in the preferences. Maybe a developer that is used to the Searx code base could help me about that

I don't think there is a reliable way to update opensearch.xml for now:
* either settings.yml has to be modified.
* [automatic update](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#Supporting_automatic_updates_for_OpenSearch_plugins) won't work since [cookies are not sent](https://github.com/asciimoo/searx/pull/2127#issuecomment-671077893)
* https://github.com/asciimoo/searx/issues/2069#issuecomment-671099077 can be a solution (not tested).

---

I understand the POST is not available on Chrom* browser. So actually the settings should be disabled in the preferences.

https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/templates/oscar/preferences.html#L77-L84

https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/templates/simple/preferences.html#L200-L209

I don't know how to implement that right now.

unixfox:
> > I don't know how to allow the ability for the user to switch back to POST requests in the preferences. Maybe a developer that is used to the Searx code base could help me about that
> 
> I don't think there is a reliable way to update opensearch.xml for now:
> 
>     * either settings.yml has to be modified.
> 
>     * [automatic update](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#Supporting_automatic_updates_for_OpenSearch_plugins) won't work since [cookies are not sent](https://github.com/asciimoo/searx/pull/2127#issuecomment-671077893)
> 
>     * [#2069 (comment)](https://github.com/asciimoo/searx/issues/2069#issuecomment-671099077) can be a solution (not tested).
> 
> 
> I understand the POST is not available on Chrom* browser. So actually the settings should be disabled in the preferences.
> 
> https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/templates/oscar/preferences.html#L77-L84
> 
> https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/templates/simple/preferences.html#L200-L209
> 
> I don't know how to implement that right now.

~~I wasn't talking about the method for the opensearch but more about the method used in the search bar on Searx home page.~~
~~How do I make it so the user can still switch back to POST requests in this search bar by changing the method in the preferences?~~
~~image~~

~~The summary is that I would like to force the GET method only if the user didn't change the method in the preferences.~~

**EDIT**: Moving the code to https://github.com/asciimoo/searx/blob/1eb00e0e717ff449ead9d49262a03cfbc4f7a53e/searx/webapp.py#L468 solved my question.

Also, I don't think we should disable the method preference because what if like I said earlier the user wants to do manual searches (from the Searx home page) in POST?

asciimoo:
> I understand the POST is not available on Chrom* browser. So actually the settings should be disabled in the preferences.

A minor correction: POST is available on chrome if the search query is submitted from the search form and if the url isn't the same as the url in the opensearch.xml

dalf:
LGTM.

---

If we want to be clean about the code:

in webapp.py:
```python
user_agent = request.headers.get('User-Agent', '').lower()
preferences = Preferences(themes, list(categories.keys()), engines, plugins, user_agent)
```

in preferences.py, before https://github.com/asciimoo/searx/blob/093dd42bb05feafb2cc77304d3b1a9a1a3016949/searx/preferences.py#L324

```python
default_method = 'POST'
if 'webkit' in user_agent and 'android' in user_agent:
    default_method = 'GET'
```
then replace this line https://github.com/asciimoo/searx/blob/093dd42bb05feafb2cc77304d3b1a9a1a3016949/searx/preferences.py#L351 by
```python
settings['server'].get('method', default_method),
```

Then all the calls to the Preferences constructor must be updated :
* tests/unit/test_search.py
   * with a FF user agent
   * add a test with a Chrome Android user agent, check the method is GET instead of POST.
* utils/standalone_searx.py
   * add a fake user agent.

dalf (approved):

dalf:
Thank you

unixfox:
@dalf Didn't you want to have a cleaner implementation based on what you suggested here: https://github.com/asciimoo/searx/pull/2132#issuecomment-672780292?

dalf:
@unixfox it would be better / cleaner, but the current implementation is simple enough, so I have merged the PR.

[EDIT] My wish is to merge some PR before a freeze period because of the "Drop Python 2" PR.

PR #2135, opened 2020-08-10
[add] yggtorrent engine

unixfox:
## What does this PR do?

This PR adds the famous French torrent tracker [Yggtorrent](https://www2.yggtorrent.si) on Searx.

1. Run Searx locally
2. Type `!ygg test` in the search bar.

### DIFF
--- a/searx/engines/yggtorrent.py
+++ b/searx/engines/yggtorrent.py
@@ -1,0 +1,124 @@
+#  Yggtorrent (Videos, Music, Files)
+#
+# @website     https://www2.yggtorrent.si
+# @provide-api no (nothing found)
+#
+# @using-api   no
+# @results     HTML (using search portal)
+# @stable      no (HTML can change)
+# @parse       url, title, seed, leech, publishedDate, filesize
+
+from lxml import html
+from operator import itemgetter
+from datetime import datetime
+from searx.engines.xpath import extract_text
+from searx.url_utils import quote
+from searx.utils import get_torrent_size
+from searx.poolrequests import get as http_get
+
+# engine dependent config
+categories = ['videos', 'music', 'files']
+paging = True
+
+# search-url
+url = 'https://www2.yggtorrent.si/'
+search_url = url + 'engine/search?name={search_term}&do=search&page={pageno}&category={search_type}'
+
+# yggtorrent specific type-definitions
+search_types = {'files': 'all',
+                'music': '2139',
+                'videos': '2145'}
+
+cookies = dict()
+
+
+def init(engine_settings=None):
+    global cookies
+    # initial cookies
+    resp = http_get(url, allow_redirects=False)
+    if resp.ok:
+        for r in resp.history:
+            cookies.update(r.cookies)
+        cookies.update(resp.cookies)
+
+
+# do search-request
+def request(query, params):
+    search_type = search_types.get(params['category'], 'all')
+    pageno = (params['pageno'] - 1) * 50
+
+    params['url'] = search_url.format(search_term=quote(query),
+                                      search_type=search_type,
+                                      pageno=pageno)
+
+    params['cookies'] = cookies
+
+    return params
+
+
+# get response from search-request
+def response(resp):
+    results = []
+    dom = html.fromstring(resp.text)
+
+    search_res = dom.xpath('//section[@id="#torrents"]/div/table/tbody/tr')
+
+    # return empty array if nothing is found
+    if not search_res:
+        return []
+
+    # parse results
+    for result in search_res:
+        link = result.xpath('.//a[@id="torrent_name"]')[0]
+        href = link.attrib.get('href')
+        title = extract_text(link)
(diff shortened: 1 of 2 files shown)


dalf:
It works but there is a two HTTP redirects each time:
1. ```https://www2.yggtorrent.si/engine/search?name=<query>&do=search&page=0&category=all``` sets two cookies: ```__cfduid``` and ```sec```
2. ```http://www2.yggtorrent.si/engine/search?name=<query>&do=search&page=0&category=all&attempt=1``` (**HTTP** request)
3. ```https://www2.yggtorrent.si/engine/search?name=<query>&do=search&page=0&category=all&attempt=1``` sets one cookie: ```ygg_```

With this patch there is only a redirect on the first request:
```diff
diff --git a/searx/engines/yggtorrent.py b/searx/engines/yggtorrent.py
index 6430d056..0cc69c47 100644
--- a/searx/engines/yggtorrent.py
+++ b/searx/engines/yggtorrent.py
@@ -14,6 +14,7 @@ from datetime import datetime
 from searx.engines.xpath import extract_text
 from searx.url_utils import quote
 from searx.utils import get_torrent_size
+from searx.poolrequests import get
 
 # engine dependent config
 categories = ['videos', 'music', 'files']
@@ -29,20 +30,31 @@ search_types = {'files': 'all',
                 'videos': '2145'}
 
 
+cookies = dict()
+
+
 # do search-request
 def request(query, params):
+    global cookies
     search_type = search_types.get(params['category'], 'all')
     pageno = (params['pageno'] - 1) * 50
 
     params['url'] = search_url.format(search_term=quote(query),
                                       search_type=search_type,
                                       pageno=pageno)
+    params['cookies'] = cookies
 
     return params
 
 
 # get response from search-request
 def response(resp):
+    # update cookies
+    for r in resp.history:
+        cookies.update(r.cookies)
+    cookies.update(resp.cookies)
+
+    # parse results
     results = []
     dom = html.fromstring(resp.text)
```

As long as https://github.com/asciimoo/searx/issues/1813 is not fixed, it won't work with uwsgi.

But the engine is disabled.

@asciimoo what do you think ?

unixfox:
Thanks for the solution about fixing the redirection.

I haven't seen any issue with these redirections though. The engine loads under 1 second which is still pretty good.

dalf:
> The engine loads under 1 second which is still pretty good.

Yes, it is fast, the redirect is not issue by itself. 

The two points that worries me:
* the HTTP request instead of HTTPS.
* the redirect may be a bot protection (not triggered with very few requests).

unixfox:
~~Is it possible to do two requests at once in a Searx engine?~~
~~One request for fetching the cookies and then another one for fetching the HTML page with the retrieved cookies from the first request?~~

unixfox:
So I used an init function just for fetching the required cookies and it works!
I got inspired by the init function in the soundcloud engine: https://github.com/asciimoo/searx/blob/master/searx/engines/soundcloud.py#L71.
This should also work on uwsgi even if all the processes won't share the same cookies but the point is that the unsecure **HTTP** redirection is gone.

searx/engines/yggtorrent.py
```diff
@@ -0,0 +1,125 @@
...
+    global cookies
+    # initial cookies
+    resp = http_get(url, allow_redirects=False)
+    if resp.ok:
+        for r in resp.history:
+            cookies.update(r.cookies)
+        cookies.update(resp.cookies)
+
+# do search-request
+
```
dalf: nitpicking: can you remove the two blank lines ?

unixfox: Unfortunately no because without that pep8 complains about not having enough blank lines:
```
/usr/lib/python3.8/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
/usr/lib/python3.8/site-packages/pep8.py:2123: UserWarning: 

pep8 has been renamed to pycodestyle (GitHub issue #466)
Use of the pep8 tool will be removed in a future release.
Please install and use `pycodestyle` instead.

$ pip install pycodestyle
$ pycodestyle ...

  warnings.warn(
searx/engines/yggtorrent.py:44:1: E302 expected 2 blank lines, found 0
```

dalf: It is okay with this:
```python
        cookies.update(resp.cookies)


# do search-request
def request(query, params):
```

This warning:
```
/home/alexandre/code/searx/local/py3/lib/python3.7/site-packages/pep8.py:109: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```
remains, but not related: https://github.com/PyCQA/pycodestyle/issues/728

unixfox:
It is good for you @dalf?

dalf:
Yes

dalf (approved):

dalf:
Thank you!

PR #2137, opened 2020-08-11
Drop Python 2

### LINKED ISSUE #1471: Drop Python 2 support ?
Whats the plan on dropping Python 2 support for searx, PY2 will go EOL in just a year, so maybe its time to consider dropping its support ?

It seems this project have only a handful of places that would need to be changed, so I can do that if required.

dalf:
Disclaimer: this is a work in progress.

## What does this PR do?

Remove Python 2.

Python 2 support makes the code more complex especially about the text encoding, url parsing.

It is difficult to make a list, basically check each modified files.

## Related issues

Close #1471


searx/webapp.py
```diff
@@ -424,7 +408,7 @@ def render(template_name, override_theme=None, **kwargs):
 
     kwargs['results_on_new_tab'] = request.preferences.get_value('results_on_new_tab')
 
-    kwargs['unicode'] = unicode
```
dalf, line 427: The unicode function can be removed.
The function is used only the 404.html pages.

dalf:
There 4 commits, but this PR is huge beast, honestly I don't know how it can be reviewed. 

If one reviewer thinks it should be split into different PR, please tell me.

kvch (changes requested):
In `tests/unit/test_preferences.py` the class `PluginStub` is still "inherited" from `object`.

Otherwise great PR!

utils/fetch_languages.py
```diff
@@ -174,14 +174,17 @@ def write_languages_file(languages):
                    + '# this file is generated automatically by utils/update_search_languages.py\n'\
                    + '\nlanguage_codes = ('
     for code in sorted(languages):
-        file_content += '\n    (u"' + code + '"'\
-                        + ', u"' + languages[code]['name'].split(' (')[0] + '"'\
-                        + ', u"' + languages[code].get('country', '') + '"'\
-                        + ', u"' + languages[code].get('english_name', '').split(' (')[0] + '"),'
+        if 'name' in languages[code]:
```
dalf, line 177: @MarcAbonce : There is an additional check here, otherwise the script crash on this entry:
```python
{'counter': ['qwant', 'qwant images', 'qwant news', 'qwant social'], 'country': ''}
```

kvch (approved):
Thank you for tackling this. :bow:

PR #2149, opened 2020-08-17
Add "Advanced settings" toggle in mobile results page 

### LINKED ISSUE #2086: On mobile, the Links box obscures results
The Links are in the way of the real results, when they should be at the bottom of the page. On mobile, this obscures a lot of the real results.

Screenshot_20200720-134215

MarcAbonce:
## What does this PR do?

The advanced search options, suggestions, corrections and links in the results page are hidden by default when using a small screen size. There's an "Advanced settings" toggle, just like the one in index, to show all the stuff that's hidden by default. In a bigger screen size everything is the same.

The results page currently wastes too much vertical space in mobile for features that are less commonly used and users have to scroll down a lot just to get to the first search result.

* Make any search with both a desktop and mobile browsers.
    - In desktop everything should look and work as normal.
    - In mobile the results should be immediately visible on screen and the Advanced toggle should be visible. Clicking that option should display the advanced search options and a couple of toggles for the download links and suggestions/corrections if available.
* Try an engine that returns corrections and search with a typo.

## Author's checklist

I decided to put the suggestions/corrections and links toggles under the advanced toggle because otherwise you still end up with a long list of toggles in the results page. I'm still not convinced that adding a toggle under another toggle is the best solution, but I couldn't come up with any better and this works.

## Related issues
Closes  #2086


MarcAbonce:
This code is kind of shuffling too many things around and I'm still not convinced this is the best UX, so I'm going to close it for now and instead I'll just open a couple of smaller PRs with the stuff that actually matters, starting with #2458.

PR #2154, opened 2020-08-21
[fix] requirements-dev.txt: set fixed versions for all the dependencies

dalf:
## What does this PR do?

Set the versions in requirements-dev.txt

An alternative change is to keep only the major version (```mock==2.*``` instead of ```mock==2.0.0``` for example)

Note this change:
```
twine==3.2.0; python_version >= "3.6"
twine==1.15.0; python_version < "3.6"
```

See https://twine.readthedocs.io/en/latest/changelog.html for reference

Without a fix version (or at least the major version), the different tools can evolve and change their behaviors.
For example, pylint version 2.6.0 add a new check:
```
R1725: Consider using Python 3 style super() without arguments (super-with-arguments)
```

Obviously, the code must be updated to remove the old Python 2 style, but it shows also we don't have reproducible build: out of nowhere travis can fail because of the dependencies has a new version.

With python 3.5 and python 3.6, check:
* ```make test```
* ```make docs```

## Related issues
N/A


kvch (approved):

kvch:
@dalf Do you mind rebasing and merging the PR? :)

PR #2159, opened 2020-08-27
Do not start a thread for an unacceptable request

### LINKED ISSUE #903: An useless thread is created on each request in the general category

kvch:
## What does this PR do?

This PR is the updated version of #992 created by @dalf . From now on engines can implement the function `is_accepted` to determine if the request submitted by the user can be executed. If not, the engine is never called.

We can get rid of unnecessary threads which we know in advance won't retrieve any results.

## Related issues

Closes #992 
Closes #903

### DIFF
--- a/searx/engines/translated.py
+++ b/searx/engines/translated.py
@@ -48,9 +48,8 @@ def request(query, params):
 
     params['url'] = url.format(from_lang=params['from_lang'][2],
                                to_lang=params['to_lang'][2],
-                               query=params['query'].decode('utf-8'))
+                               query=params['query'].decode('utf-8'),
                                key=key_form)
-
     return params
 
 
@@ -64,7 +63,7 @@ def response(resp):
         'title': '[{0}-{1}] {2}'.format(
             resp.search_params['from_lang'][1],
             resp.search_params['to_lang'][1],
-            resp.search_params['query']),
+            resp.search_params['query'])
         'content': resp.json()['responseData']['translatedText']
     })
     return results
--- a/searx/engines/translated.py
+++ b/searx/engines/translated.py
@@ -20,30 +20,36 @@ parser_re = re.compile(b'.*?([a-z]+)-([a-z]+) (.{2,})$', re.I)
 api_key = ''
 
 
-def request(query, params):
+def is_accepted(query, params):
     m = parser_re.match(query)
     if not m:
-        return params
+        return False
 
+    params['parsed_regex'] = m
     from_lang, to_lang, query = m.groups()
 
     from_lang = is_valid_lang(from_lang)
     to_lang = is_valid_lang(to_lang)
 
     if not from_lang or not to_lang:
-        return params
+        return False
 
-    if api_key:
-        key_form = '&key=' + api_key
-    else:
-        key_form = ''
-    params['url'] = url.format(from_lang=from_lang[1],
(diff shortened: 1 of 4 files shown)


searx/engines/translated.py
```diff
@@ -58,7 +63,7 @@ def response(resp):
         'title': '[{0}-{1}] {2}'.format(
             resp.search_params['from_lang'][1],
             resp.search_params['to_lang'][1],
-            resp.search_params['query']),
+            resp.search_params['query'])
```
dalf, line 66: Without the coma, there is a syntax error:
https://travis-ci.org/github/asciimoo/searx/jobs/721829981#L938

asciimoo:
Wouldn't it be easier to just leave the url blank in the `request` function and check it from the caller?

Minor note: it would be good to add documentation about this feature.

dalf:
It is too late in the ```request``` function: the thread has already been started.

The callstack:
* Search.search <-- this PR checks if the engine can accept the request in this method.
* search_multiple_requests <--- here one thread is started per engine request
* search_one_request_safe
* search_one_http_request_safe
* [search_one_http_request](https://github.com/asciimoo/searx/blob/6f9aa0e2581815189db06d3e7815f1e11b9a4f15/searx/search.py#L84-L101)
* request <-- if ```request_params['url'] is None```, then there is no request, but a thread has been started for nothing.

asciimoo:
> It is too late in the request function: the thread has already been started.

True. Then my question is: is it worth to introduce a new function just to be able to avoid starting a thread? Starting a few threads isn't that resource consuming tasks.

dalf:
Does it worth a new function ? It requires some benchmark (on high end hardware it will be nothing, on a small SBC, perhaps it is interesting).

Some brainstorming most probably outside the scope of this PR, but still related: we can imagine that these syntaxes are generic to all currency and translation engines:
    * ```10 eur in usb```
    * ```en-it house```

In this case:
* ```is_accepted``` would be generic (one per syntax)
* the ```currency_convert``` engine declares itself as a currency convert engine
* the ```dictzone``` and ```translated``` engines declare themself as translation engines.
* the ```request``` method expect some well define parameters (currency converter --> from / to / amount)
* we can display this information to the user.

[EDIT] in the new architecture, the ```request``` and ```response``` functions will be async, and no more thread.
[EDIT2] possible implementation : https://github.com/dalf/searx/tree/processors/searx/search/processors

kvch:
Closing this PR after an offline conversation with @asciimoo. We agreed that starting a new thread is not a big issue. The search service never gets the invalid request, so we are not putting unnecessary load on anyone. Besides @dalf is refactoring the request and respons mechanism, so threads will be gone.

PR #2160, opened 2020-08-28
[enh] test: load each engine to check for syntax errors

dalf:
## What does this PR do?

In the tests (```make test```), all the engines are loaded to check for syntax errors.

In the https://github.com/asciimoo/searx/pull/2159 , Travis, ```make test``` are ok despite a syntax error: https://travis-ci.org/github/asciimoo/searx/jobs/721829981#L938

Note:
* Before this PR, the engines are loaded as soon as searx.webapp is imported.
* After this PR, the engines are loaded inside searx.webapp.run()
* Before and after this PR, searx starts even if some engines can't start.

Call 
* ```make test```
* ```make run```

## Related issues

Possible enhancement, later:
* call ```is_accepted(...)```
* call ```request(...)```
* it is not possible to call ```response(....)``` without the previous version of the tests.

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -72,11 +72,8 @@ def load_engine(engine_data):
 
     engine_module = engine_data['engine']
 
-    try:
-        engine = load_module(engine_module + '.py', engine_dir)
-    except:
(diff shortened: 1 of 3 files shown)


asciimoo (approved):
Good change, thanks!

PR #2167, opened 2020-09-01
Add key to settings.yml for opening result links in a new tab

### LINKED ISSUE #444: Results links open in the same tab
How to set searx to open links in a new tab?

### LINKED ISSUE #1552: Add option to set "Results on new tabs" option in settings.yml
I run a personal instance of SearX on my home server in a docker container. Because of the way this container is set up, when I restart it I lose any "preferences" I've set from within the searx <DOMAIN>/preferences web interface. The only preferences that remain across restarts are those in my settings.yml. That's generally fine, but I would like the option of being able to set the option "Results on new tabs" to "on" in the settings.yml so that this is the default setting. Is there a way to make this possible?

See also issue #444

scottwallacesh:
## What does this PR do?

Adds a configuration key in `settings.yml` for opening result links in new tab by default.

Gives admins the ability to set opening result links in a new tab by default.  Users can still override this functionality in the preferences cookie(s).

Add `results_on_new_tab: True` to the `ui` section in `settings.yml`, restart Searx, perform a search, click a result link and it should open in a new tab.  Comment out the key or set the value to `False`, restart Searx, perform a search, click a result link and it should open in the same tab.

## Author's checklist

Does not alter functionality by default.  It only activates when the correct key and value is added to `settings.yml`.

## Related issues

Closes #1552 
Closes #444

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -364,7 +364,7 @@ class Preferences(object):
                 choices=themes
             ),
             'results_on_new_tab': MapSetting(
-                False,
+                settings['ui'].get('results_on_new_tab', False),
                 map={
                     '0': False,
                     '1': True,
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Great, thanks!

Issue #1802, opened 2020-01-10
Engine code: describe which XPath can fail, which must not.


dalf:
In the result parsing, engines parse the HTML using lxml (most of the time). If an XPath request doesn't return at least one result, it may be fine or trigger an error later. In the later case, it is difficult to know exactly what is going on without looking at the downloaded HTML.

This issue suggests:
* to add new exception classes.
* to add two optional parameters to [eval_xpath](https://github.com/asciimoo/searx/blob/85b37233458c21b775bf98568c0a5c9260aa14fe/searx/utils.py#L465) function to check the result count.

It may help to know when an engine starts to be broken if the engine codes says which XPath request should not fail (?).

I'm not sure if it is useful and /or a privacy problem if searx makes statistics about broken XPath ?

### Class hierarchy

```
SearxException
	SearxParameterException
	SearxEngineException
		SearxEngineCaptchaException (instead of RuntimeWarning in google.py)
		SearxEngineXPathException
```

### eval_xpath
```python	
def eval_xpath(element, xpath_str, eq=None, gte=None):
    xpath = get_xpath(xpath_str)
    result = xpath(element)
    # new code: check result count now
    if eq is not None and len(result) != eq:
	raise SearxEngineXPathException(xpath, eq=eq)
    if gte is not None and len(result) < gte:
	raise SearxEngineXPathException(xpath, gte=gte)
    return result
```

#### extract_url
https://github.com/asciimoo/searx/blob/master/searx/engines/xpath.py#L53
```python
def extract_url(xpath_results, search_url):
    if xpath_results == []:
        raise Exception('Empty url resultset')	
```
--> Make the check before calling extract_url

#### bing engine
```python
	...
    for result in eval_xpath(dom, '//div[@class="sa_cc"]'):
        link = eval_xpath(result, './/h3/a', eq=1)[0]
	...
    for result in eval_xpath(dom, '//li[@class="b_algo"]'):
        link = eval_xpath(result, './/h2/a', eq=1)[0]
	...
```

#### google engine
```python
	title = extract_text(eval_xpath(result, title_xpath, eq=1)[0])
	url = parse_url(extract_url(eval_xpath(result, url_xpath, eq=1), google_url), google_hostname)
```
The huge try/catch to ignore all the parsing errors would be able to display the XPath in the logs.

Another way without try/catch and without modification to the eval_xpath function:
```python
	title_xpr = eval_xpath(result, title_xpath)
	url_xpr = eval_xpath(result, url_xpath)
[...]

Issue #585, opened 2016-06-11
crashing when sites block


ibrokemypie:
numerous sites appear to be blocking searx.space now, including soundcloud.com. after updating to the latest commit and restarting, i found that searx simply would not run. debug mode shows `INFO:searx:Initialisation done
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): soundcloud.com
Traceback (most recent call last):
  File "searx/webapp.py", line 55, in <module>
    from searx.engines import (
  File "/root/searx/searx/engines/__init__.py", line 208, in <module>
    engine = load_engine(engine_data)
  File "/root/searx/searx/engines/__init__.py", line 60, in load_engine
    engine = load_module(engine_name + '.py')
  File "/root/searx/searx/engines/__init__.py", line 53, in load_module
    module = load_source(modname, filepath)
  File "/root/searx/searx/engines/soundcloud.py", line 61, in <module>
    guest_client_id = get_client_id()
  File "/root/searx/searx/engines/soundcloud.py", line 41, in get_client_id
    response = http_get("https://soundcloud.com")
  File "/root/searx/searx/poolrequests.py", line 86, in get
    return request('get', url, **kwargs)
  File "/root/searx/searx/poolrequests.py", line 79, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests-2.9.1-py2.7.egg/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests-2.9.1-py2.7.egg/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests-2.9.1-py2.7.egg/requests/adapters.py", line 437, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='soundcloud.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x30c5fd0>: Failed to establish a new connection: [Errno 110] Connection timed out',))` which has lead me to believe that due to the block by soundcloud, searx is crashing.

dalf:
The problem occurs at this line : 
https://github.com/asciimoo/searx/blob/master/searx/engines/soundcloud.py#L61

The call is from this line : 
[...]

PR #2177, opened 2020-09-07
[enh] stop searx when an engine raise an SyntaxError exception

dalf:
and some other exceptions:
* KeyboardInterrupt
* SystemExit
* RuntimeError
* SystemError
* ImportError: an engine with an unmet dependency will stop everything.

## What does this PR do?

Update version of https://github.com/searx/searx/pull/2160

This PR make stop searx loading when Python raises a SyntaxError exception

In the https://github.com/asciimoo/searx/pull/2159 , Travis, ```make test``` are ok despite a syntax error: https://travis-ci.org/github/asciimoo/searx/jobs/721829981#L938

1. check searx starts as before
1. add an syntax error in an engine
1. check searx stops

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -74,6 +74,9 @@ def load_engine(engine_data):
 
     try:
         engine = load_module(engine_module + '.py', engine_dir)
+    except (SyntaxError, KeyboardInterrupt, SystemExit, SystemError, ImportError, RuntimeError) as e:
+        logger.exception('Fatal exception in engine "{}"'.format(engine_module))
+        sys.exit(1)
     except:
         logger.exception('Cannot load engine "{}"'.format(engine_module))
         return None


return42:
Isn't it better to remove the entire try/except? Catching base class Exception, see [They all do the same massive disservice: silently and invisibly hiding error conditions that can otherwise be quickly detected and dispatched.](https://realpython.com/the-most-diabolical-python-antipattern/)

dalf:
@return42 short answer: yes but not now.

long answer:
* the try...except was introduced by this commit: https://github.com/searx/searx/commit/54d987636e4b03d19a99ad9d143bf63b119af208
* the commit fixed this issue : https://github.com/searx/searx/issues/585 (which remind me I had suggested this fix...)

Sum up: currently, there is no / few exception handling in the engine code. An unexpected result from a HTTP result may raise an exception. If searx stops everything, it will be a mess to manage (the service will may restart automatically or not, perhaps it may raise the same exception or not).

My wish is to implement an extended version of https://github.com/searx/searx/issues/1802 where things fail safely, and where an exception is really an unexpected behavior ; but it can't be implemented in few days.

So this PR is temporary fix to get one thing: at least check the syntax error / import errors to avoid "the tests are okay, we can merge... and actually we shouldn't have because there is a syntax error".

Other things may fails silently, for now, as it used to be (no unexpected behavior when searx is upgraded).

return42:
>  yes but not now.

Ah! OK .. thanks for clarifying!

return42 (approved):

return42 (approved):
LGTM

Issue #2034, opened 2020-06-29
lets establish a build process for all the build products


return42:
We should no longer commit build products to the repository. ATM we are committing build products like minified CSS and JS from themes, which is the cause of ongoing merge-conflicts. 

Beside themes there are other topics we can generalize in a build phase. I think about the logo, there should be one logo image (e.g. a SVG used as the single source of definition) and derived logos in the themes.

More suggestions are welcome, add your toughs  ...

asciimoo:
I agree with the proposal but it raises new questions:
The most important one is: do we want to build the artifacts or let the administrators do it? Both have pros and cons.

If the build is done by the admins, it greatly increases the number of dependencies required to host searx (nodejs, npm, etc..), probably there are many administrators who would dislike installing nodejs to their servers just to be able to host searx.
And if we want to produce the artifacts, there are multiple questions we have to address:
 1. Where should we build resources?
 2. How can we distribute the artifacts with a privacy respecting manner?
 3. How should we handle the versioning of the products?

dalf:
In https://github.com/sethmlarson/hstspreload , there is a [cron](https://github.com/sethmlarson/hstspreload/blob/master/.travis.yml#L26-L27) which build a pypi package automatically: https://pypi.org/project/hstspreload/#history 

Can we do something similar ?

One idea, but frankly speaking I don't like too much because I'm afraid it will be burden rather than a help:
* a side project called ```searx-bin``` can contains the built files, so people can just ```pip install searx-bin``` (trigger: a commit in ```searx``` repository).
* ```searx-bin``` could contains a hash of the source file, so searx checks if the ```searx-bin``` are out of sync.

asciimoo:
> In https://github.com/sethmlarson/hstspreload , there is a cron which build a pypi package automatically: https://pypi.org/project/hstspreload/#history

Its not entirely clear to me how could we use this solution in our case. Would it mean that we have to build pypi packages for every PR to be able to test it?

> a side project called searx-bin can contains the built files, so people can just pip install searx-bin (trigger: a commit in searx repository).

Yeah, I was thinking about a similar solution. Perhaps instead of a repo, we can use a build server (e.g. searx.me) to create and serve the artifacts - not sure if it would be better.

return42:
> Yeah, I was thinking about a similar solution. Perhaps instead of a repo, we can use a build server (e.g. searx.me) to create and serve the artifacts - not sure if it would be better.

[...]

PR #2178, opened 2020-09-08
Use Wikipedia's REST v1 API 

MarcAbonce:
## What does this PR do?

After the changes in #2102 I realised that sometimes the Spanish Wikipedia still returns disambiguation paragraphs in the infobox. The new REST API doesn't have that problem.

This also fixes a little issue with the infobox toggle button when sometimes the button is hidden even though there is more content to be shown.

This is the API that Wikipedia currently uses to retrieve an article's preview when you hover your mouse over a link in Wikipedia. Therefore, using this API is way simpler to use and should also be more accurate than the older, general purpose API that's not really meant for article previews.

* Try to query an article with a disambiguation paragraph, such as `:es !wp naruto`. No disambiguation (*desambiguación*) paragraph should be shown.
* Same tests as #2102.
* The toggle issue could be harder to reproduce because the error happens when the image is loaded a fraction of a second after the JS code was run. But try `:en !wp Port Dover` with master and see if the toggle button is hidden even though there's a hidden image. That bug should never happen in this branch.

## Author's checklist

The official documentation (https://en.wikipedia.org/api/rest_v1/) mentions two rules:
> 
> * Limit your clients to no more than 200 requests/s to this API. Each API endpoint's documentation may detail more specific usage limits.
> * Set a unique User-Agent or Api-User-Agent header that allows us to contact you quickly. Email addresses or URLs of contact pages work well.

I set the Wikipedia engine to send searx's user agent to comply with that.


searx/engines/wikipedia.py
```diff
@@ -41,51 +31,37 @@ def url_lang(lang):
...
-
-    image = page.get('thumbnail')
-    if image:
-        image = image.get('source')
+    results = []
+    api_result = loads(resp.text)
 
-    summary = page.get('extract', '').split('\n')[0].replace('()', '')
+    # skip disambiguation pages
+    if api_result['type'] != 'standard':
```
dalf, line 53: The type ```disambiguation``` is ignored. I don't know if this is intent?

Example with ```!wp :en search```

```curl -L -vvv -X GET "https://en.wikipedia.org/api/rest_v1/page/summary/search" -H  "accept: application/json; charset=utf-8; profile="https://www.mediawiki.org/wiki/Specs/Summary/1.4.2""```

Comment (MarcAbonce) [reply searx/engines/wikipedia.py:53]: Yeah. Otherwise, with your example you would end up with an infobox that just says: 
> Searching or search may refer to:  

...and that's it, but that would be rather useless in my opinion. Although maybe I should still add the URL to the normal results in that case...

File: searx/static/themes/oscar/js/searx.min.js [hunk 2]
```diff
@@ -1,2 +1,2 @@
-/*! oscar/searx.min.js | 10-08-2020 | https://github.com/asciimoo/searx  */
```
SuperSandro2000, line 1: Minified files do not belong into git and easily create merge conflicts.

return42: See https://github.com/searx/searx/issues/2034

dalf: @SuperSandro2000 minified files exist in the git repository before this PR. I don't say we shouldn't change this (see @return42 issue), but not in this PR.

dalf (approved):

PR #2182, opened 2020-09-09
[fix] docker and k8s: run searx

### LINKED ISSUE #2181: The docker image no longer runs in k8s
Due to a change in the docker entrypoint script in [this PR](https://github.com/searx/searx/pull/2099), the docker image no longer runs in k8s (the help function gets called and then the script exits).  I'm using microk8s on ubuntu, if that helps.

I'm not sure how important kubernetes support is, so I'm opening an issue versus a PR.

dalf:
## What does this PR do?

docs/admin/installation-docker.rst calls dockerfiles/docker-entrypoint.sh to include the help message inside the documentation.

dockerfiles/docker-entrypoint.sh detects if docker is running using ```grep docker /proc/1/cgroup -qa```.

This method doesn't work with k8s, see #2181

This PR fixes this issue:
* add a new parameter ```docs``` to dockerfiles/docker-entrypoint.sh 
* docs/admin/installation-docker.rst calls ```dockerfiles/docker-entrypoint.sh  docs```.

Without this PR it is possible to run the searx docker with k8s.

* ```make docker```
* add a new docker image to microk8s
* check if searx has started

## Related issues

Close #2181

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -19,20 +19,24 @@ help() {
 	exit 0
 }
 
-if ! grep docker /proc/1/cgroup -qa; then
-    help
-fi
-
 export DEFAULT_BIND_ADDRESS="0.0.0.0:8080"
 if [ -z "${BIND_ADDRESS}" ]; then
     export BIND_ADDRESS="${DEFAULT_BIND_ADDRESS}"
 fi
 
-export SEARX_VERSION=$(su searx -c 'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' 2>/dev/null)
-printf 'searx version %s\n\n' "${SEARX_VERSION}"
-
 export UWSGI_SETTINGS_PATH=/etc/searx/uwsgi.ini
 export SEARX_SETTINGS_PATH=/etc/searx/settings.yml
+
+# Parse special command line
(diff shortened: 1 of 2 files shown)


return42 (dismissed):
I haven't tested k8s, but the 'doc' option LGTM

dockerfiles/docker-entrypoint.sh
```diff
@@ -19,25 +19,30 @@ help() {
...
-export SEARX_VERSION=$(su searx -c 'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' 2>/dev/null)
-printf 'searx version %s\n\n' "${SEARX_VERSION}"
-
 export UWSGI_SETTINGS_PATH=/etc/searx/uwsgi.ini
 export SEARX_SETTINGS_PATH=/etc/searx/settings.yml
 
+# Parse special command line
+# see docs/admin/installation-docker.rst
+# display the help message without the version
+if [ "$1" = "docs" ]; then
```
SuperSandro2000: why is this called docs and not help?

dalf: Because the output will be included into the documentation which is produce using ```make docs```.

But ```help``` makes sense.

dalf: Updated

Issue #2067, opened 2020-07-13
New architecture proposal


dalf:
This [wiki page](https://github.com/asciimoo/searx/wiki/New-architecture-proposal) describes a new architecture.

Feel free to add comment here or send me a email (see my github profile)

Poke:
* everyone
* @nfk since you maintain a fork of searx.
* @unixfox as a maintainer of searx.be.

The purpose is to be able to rewrite some part of searx in golang:
* the front-end:
  * morty and filtron could be a library
  * the front-end would embed these two libraries and call the workers.
* the workers:
  * the XPath engines for example.

return42:
Thank you for your article, I read it with interest and we had already talked a little about it in a PN.

Here are my 5cent: If I forget technology and all we already have, how should a process model looks like?

I will start with the HTTP client & server communication; for a maximum of performance and a smooth UX this communication should be asynchronous.

The user sends the search term to the front-end, the first thing the front-end does, is to send a response to the client which instructs the client to load some resources (e.g. static resources like CSS files or even the JS).

In the next step, the front-end parses user's query.  A user query can be different things, mostly it is a search term but sometimes it might be a [algebra expression](https://github.com/asciimoo/searx/pull/3849) asking sympy or whatever application the future brings.  What I mean: we can have workers for HTTP queries (our engines) or even workers which are more an application.

For me _plugins_ in searx are more like filters and  https://github.com/asciimoo/searx/pull/3849 is more like a self-contained application.

Back to the client/front-end communication in which we have different tasks/subjects, e.g we have a "info-box" which could be loaded asynchronous .. or a "instant-answer" box .. the _main-box_ is the "result-list" .. all these "boxes" can be load asynchronous and not everything has to wait until the last engine has answered ..

That was very unsorted, I hope it is understandable what I mean: due to the asynchronous appearance of the page, the timeouts for the engines can be extended by 0,5-1 sec. without the user noticing.

return42:
@asciimoo  What do you think about such fundamental changes? Would you be interested in this or would you rather reject it because you might be pursuing a different goal. Both would be fine for me, I just don't have time to bet on a dead horse.

dalf:
As I understand, you are describing a progressive rendering from a searx instance:
* as soon as there is a request, send the HTML ```<head>``` (or HTTP/2 push or whatever), so the browser can download the CSS/JS.
* as soon as infoboxes are ready send them to the browser.
* as soon as answers are ready send them to the browser.

[...]

PR #2189, opened 2020-09-10
Architecture clean up

dalf:
## What does this PR do?

Architecture clean up.

### Overview

Some notes:
* There is no functional changes.
* Add ```__slots__``` to some classes to make clear the list of the attributes.

#### web
* searx.query
* searx.webadapter
* searx.webapp
* searx.autocomplete
* searx.preferences

#### core
* searx.search
* searx.external_bang
* searx.poolrequests
* searx.languages (engines call searx.utils.is_valid_lang)
* searx.results
	
The entry point of the core is searx.search.Search
* input: searx.search.SearchQuery
* output: searx.results.ResultContainer

#### both:
* searx.utils
* searx.brand
* searx.exceptions
	
#### other:
* searx.testing

### searx/search.py

* [SearchQuery](https://github.com/searx/searx/blob/fd8a63bd26e5890e2a125b9db69e5720cdec61d7/searx/search.py#L65-L92)
	* was previously in query.py
	* this object is the entry point to the Search class, it doesn't contain any reference to RawTextQuery
	* time_range parameter: 'None' or '' are not accepted anymore (see [searx.webadapter.parse_time_range](https://github.com/searx/searx/blob/fd8a63bd26e5890e2a125b9db69e5720cdec61d7/searx/webadapter.py#L87)). Previously: https://github.com/searx/searx/blob/ae07f4a211ecba0331bcab5903e3263c646f8bdb/searx/query.py#L194
	* remove reference to the [preferences parameters](https://github.com/searx/searx/blob/ae07f4a211ecba0331bcab5903e3263c646f8bdb/searx/query.py#L196) (see below searx.search.Search)

* [EngineRef](https://github.com/searx/searx/blob/fd8a63bd26e5890e2a125b9db69e5720cdec61d7/searx/search.py#L52-L62)
	* was previously a [Dict](https://github.com/searx/searx/blob/ae07f4a211ecba0331bcab5903e3263c646f8bdb/searx/query.py#L138-L140) with two or three keys: name, category, from_bang
	* make clear that this is a engine reference (see tests/unit/test_search.py for example)
	* all variables using this class are renamed accordingly.

* Search class:
	* important change in the search method: the token is not more checked in this method. The check is done before (see [searx.webadapter](https://github.com/searx/searx/blob/fd8a63bd26e5890e2a125b9db69e5720cdec61d7/searx/webadapter.py#L34-L36))

### searx/query.py

* RawTextQuery
	* engines attribute is replace by enginerefs

### searx/webadapter.py

* get_search_query_from_webapp
	* previously in webapp.py
	* in charge of checking the engine tokens (actually done by the validate_engineref_list function)
	* returns a SearchQuery, RawTextQuery, the list of unknown engine, the list of engine without token (might be useful to display them in result page)
* each parameter has a dedicated function (parse_pageno, page_lang,....).
	* for example, it makes clear that query_pageno depend only of the form, not from the preferences, not from the cookies.
	* easier to add test about each of them.
* parse_category_form
	* parse the ```category=general,music``` and ```category_music=on```
	* get_search_query_from_webapp and get_selected_categories functions call this function
* get_selected_categories
	* from a form (optional) and preferences, return the selected categories
	* was https://github.com/searx/searx/blob/272158944bf13503e2597018fc60a00baddec660/searx/webapp.py#L371-L385
	* note: searx.webapp.render iterates over request.args and searx.webapdater.get_selected_categories iterates over request.form (related to https://gitlab.e.foundation/e/cloud/my-spot/commit/812bac9d9649b5ba53a97a43c15b739692706afa )

### searx/webapp.py

* render function:
	* remove selected_categories, set in preferences() and index() using the new searx.webadapter.get_selected_categories function.
	* remove all_categories: set in the preferences() function.

* It makes the code more readable (at least I hope it is the case)
* without huge changes, it makes https://github.com/searx/searx/issues/2067 easier to implement.

N/A. This section has to be written.

## Related issues

* https://github.com/searx/searx/issues/2067

### DIFF
--- a/searx/search.py
+++ b/searx/search.py
@@ -16,26 +16,20 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 '''
 
 import gc
-import sys
 import threading
 from time import time
 from uuid import uuid4
 from _thread import start_new_thread
 
-from flask_babel import gettext
 import requests.exceptions
 import searx.poolrequests as requests_lib
-from searx.engines import (
-    categories, engines, settings
-)
+from searx.engines import engines, settings
 from searx.answerers import ask
 from searx.external_bang import get_bang_url
 from searx.utils import gen_useragent
-from searx.query import RawTextQuery, SearchQuery, VALID_LANGUAGE_CODE
 from searx.results import ResultContainer
 from searx import logger
 from searx.plugins import plugins
-from searx.exceptions import SearxParameterException
 
 
 logger = logger.getChild('search')
@@ -51,6 +45,26 @@ else:
         from sys import exit
 
         exit(1)
+
+
+class SearchQuery:
+    """container for all the search parameters (query, language, etc...)"""
+
+    def __init__(self, query, engines, categories, lang, safesearch, pageno, time_range,
+                 timeout_limit=None, preferences=None, external_bang=None):
+        self.query = query
+        self.engines = engines
+        self.categories = categories
+        self.lang = lang
+        self.safesearch = safesearch
+        self.pageno = pageno
+        self.time_range = None if time_range in ('', 'None', None) else time_range
+        self.timeout_limit = timeout_limit
+        self.preferences = preferences
+        self.external_bang = external_bang
+
+    def __str__(self):
+        return self.query + ";" + str(self.engines)
 
 
 def send_http_request(engine, request_params):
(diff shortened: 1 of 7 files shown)


searx/search.py
```diff
@@ -53,6 +48,49 @@
...
+    def __init__(self, name: str, category: str, from_bang: bool=False):
+        self.name = name
+        self.category = category
+        self.from_bang = from_bang
+
+    def __str__(self):
+        return "(" + self.name + "," + self.category + "," + str(self.from_bang) + ")"
+
+
+class SearchQuery:
```
asciimoo, line 64: What is the benefit of moving `SearchQuery` from `query.py`?

searx/webadapter.py
```diff
@@ -0,0 +1,234 @@
...
+        engine = engines[engineref.name]
+        if not preferences.validate_token(engine):
+            no_token.append(engineref)
+            continue
+
+        valid.append(engineref)
+    return valid, unknown, no_token
+
+
+def parse_pageno(form: typing.Dict[str, str]) -> int:
```
asciimoo: Wouldn't it be better to add the functionality of the below functions to `preferences.py` somehow?

asciimoo (commented):
LGTM overall, please check my notes.

searx/search.py
dalf: The purpose is to have one simple object to run a query: the searx.search.Search object.
It takes a SearchQuery object as input and returns a ResultContainer object.
SearchQuery contains only simple Python types, it doesn't depend on the web / front-end.

The front-end parses the user inputs using webadapter.py.

And webadapter.py uses:
* query.py to parse the query & bangs.
* preferences.py to parse the preferences from the cookies.

With SearchQuery in search.py, it creates a clean split between front-end and back-end.

searx/webadapter.py
dalf: TLDR: 
* Yes, it would make sense to avoid parsing the parameters at two different locations in the code.
* currently webadapter.py is more or less cut / paste from search.py (with the exception of get_selected_categories).
* the parsing from webadapter/preferences/webapp is not easy to follow.
* Perhaps it would be easier to do that in another PR ?

----

There are some differences between webadapter.py and preferences.py: 
* The parameters in the preferences.py are stored in the cookies, and at least one parameter doesn't make sense here: ```pageno```.
* These two parameters could be set in the preferences:
  * ```time_range``` with a default of ```None```
  * ```timeout``` from query (see https://github.com/searx/searx/pull/1640 ), with a default of ```None```

webadapter.py or preferences.py could convert the ```categorie_*=on|off``` parameters into ```categories=...,...```.

preferences.py and webadapter.py parses safesearch in the same way.

------

Example of complicated parsing: the language parameter.

* preferences.py checks the language value [using the searx.utils.match_language function](https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/preferences.py#L164-L168)
* webapp.py checks the language value from the URL using a regex. This is on purpose, see https://github.com/searx/searx/commit/fd65c1292179fb082e965a1ee6a88b9298a54fc1

Knowing that, let's follow how the language value is set.

https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/preferences.py#L323-L326

The default value is ```settings['search'].get('default_lang', '')```

searx.webapp.pre_request:
1. initialize a Preferences object.
1. then parses [the cookies](https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/webapp.py#L458)
1. then parses [the ```preferences``` query parameter](https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/webapp.py#L470)
1. then parses [the query parameters](https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/webapp.py#L473)
1. then if the ```language``` value is empty or None, the value is set to the result of ```_get_browser_or_settings_language```:
    * the Accept-Language header value
	* or ```settings['search'].get('default_lang', 'en')``` (second time usage of the settings)

searx.webapp.index calls get_search_query_from_webapp:
  1. it gets the value from ```raw_text_query.languages[-1]```
  1. if the value is empty/not set, it uses ```form.get('language')```: actually the value should already be in the ```preferences``` (```searx.webapp.pre_request```), however the checking is less restrictive here (```VALID_LANGUAGE_CODE.match(query_lang)``` instead of ```searx.utils.match_language```).
  1. if not, from the preferences

So, at least, the second step can be removed in get_search_query_from_webapp, but there is no more exception, so an invalid value is just ignored.

asciimoo (approved):
Alright, let's test the current state and we can iterate later.
Thanks!

PR #2190, opened 2020-09-11
[fix] searx.utils.HTMLTextExtractor: invalid HTML don't raise an Exception

### LINKED ISSUE #2188: 500 internal error for json data from some searches
For some searches with searx I receive an 500 internal error when trying to access the json data (or csv or rss). To reproduce this try the following steps:

1. Search for `Maya_sculpture-71.893.4` e.g. https://searx.xyz/?q=Maya_sculpture-71.893.4
2. Click on the `json` button. to download the results as json data.

This leads to the 500 error. However, most other searches work well and also the json data can be downloaded. Only with this very specific search and few others I have experienced this problem.

Does anyone see why this causes a problem? How can this been solved?

dalf:
## What does this PR do?

searx.utils.HTMLTextExtractor :
* without this PR, invalid HTML raise an ```Exception```: https://github.com/searx/searx/blob/272158944bf13503e2597018fc60a00baddec660/searx/utils.py#L112
* with this PR, invalid HTML raises a dedicated exception, then ```html_to_text``` can return the partial content.

The ```Exception``` ends up to a HTTP Error 500 for the user.

* See the new tests in tests/unit/test_utils.py

## Related issues

Close #2188

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -77,6 +77,10 @@ def highlight_content(content, query):
     return content
 
 
+class HTMLTextExtractorException(Exception):
+    pass
+
+
 class HTMLTextExtractor(HTMLParser):
 
     def __init__(self):
@@ -92,7 +96,7 @@ class HTMLTextExtractor(HTMLParser):
             return
 
         if tag != self.tags[-1]:
-            raise Exception("invalid html")
+            raise HTMLTextExtractorException()
 
         self.tags.pop()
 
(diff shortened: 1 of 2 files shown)


kvch:
Why swallow the error completely? Is it not worth logging the exception so instance admins can learn about the error?

dalf:
How do you see the logging ?

instead of ```pass```:

```python
logger.warn("HTMLTextExtractor : invalid HTML\n%s", html)
```

dalf:
Updated with ```logger.debug```.

PR #2195, opened 2020-09-14
Small changes

dalf:
## What does this PR do?

1. searx.query.RawTextQuery: rename the methods getSearchQuery and changeSearchQuery
1. searx.search: remove the unused global variable number_of_searches
1. searx.search: change the function declaration order.

There is no functional change.

1. searx.query.RawTextQuery: getSearchQuery is confusing, the method returns a ```str``` not a [SearchQuery](https://github.com/searx/searx/blob/ae07f4a211ecba0331bcab5903e3263c646f8bdb/searx/query.py#L183) object
1. number_of_searches : if there is some statistics, it should go into the ```searx/engines/__init__.py```
1. searx.search: function declaration order: it make the code more readable (perhaps not a shared opinion).

* Make sure there is no usage of getSearchQuery, changeSearchQuery, number_of_searches

## Related issues

Partial implementation of #2189

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -38,22 +38,22 @@ def get(*args, **kwargs):
 def searx_bang(full_query):
     '''check if the searchQuery contain a bang, and create fitting autocompleter results'''
     # check if there is a query which can be parsed
-    if len(full_query.getSearchQuery()) == 0:
+    if len(full_query.getQuery()) == 0:
         return []
 
     results = []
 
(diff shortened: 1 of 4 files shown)


return42:
Sorry for nitpicking / for the future:

> searx.query.RawTextQuery: getSearchQuery is confusing, the method returns a str not a SearchQuery object

shouldn't this info better placed in the commit messages?

PR #2199, opened 2020-09-15
[mod] upgrade requests to version 2.24.0. pyopenssl becomes optional.

dalf:
from [requests 2.24.0 changelog](https://github.com/psf/requests/blob/master/HISTORY.md#2240-2020-06-17): 
> pyOpenSSL TLS implementation is now only used if Python either doesn’t have an ssl module or doesn’t support SNI.

## What does this PR do?

* upgrade requests to version 2.24.0
* requirements.txt: pyopenssl is still required to install searx as a fallback.
* searx log a critical message if ssl modules doesn't support SNI, and pyOpenSSL is not installed.
* The message is logged from searx.poolrequests instead of searx.webapp.

See 
* https://docs.python.org/3/library/ssl.html#ssl.HAS_SNI
* https://github.com/psf/requests/pull/5443

SNI is supported by Python version 3.2.

check HTTPS connection are working in the following scenarios:
* upgrade with requests and pyopenssl
* requests 2.24.0, and without pyopenssl

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,12 @@
-certifi==2020.4.5.1
+certifi==2020.6.20
 babel==2.7.0
 flask-babel==1.0.0
 flask==1.1.2
-idna==2.9
+idna==2.10
 jinja2==2.11.1
 lxml==4.5.0
 pygments==2.1.3
 pyopenssl==19.1.0
 python-dateutil==2.8.0
 pyyaml==5.3.1
-requests[socks]==2.23.0
+requests[socks]==2.24.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,12 +1,12 @@
...
 lxml==4.5.0
 pygments==2.1.3
-pyopenssl==19.1.0
 python-dateutil==2.8.0
 pyyaml==5.3.1
-requests[socks]==2.23.0
+requests[socks]==2.24.0
+idna==2.10
+pyopenssl==19.1.0
+certifi==2020.6.20
```
SuperSandro2000: If we keep this sorted version updates are easier to spot.

dalf: The sort order in this file is basically random, the version above tries to group by theme (flask, http,...).
But anyway this is small file.
I've have pushed force a new commit that keep the same sort order.

return42: > The sort order in this file is basically random

Is it? In f9186344 there was an alphabetic order and in a3368c07a babel was added at the wrong position.

Issue #2204, opened 2020-09-19
Documentation build is broken


return42:
The build of the docs has some errors (and warnings) ..

````
$ make clean docs-live
CLEAN     docs-clean
PYENV     usage: source ./local/py3/bin/activate
build searx/brand.py
build utils/brand.env
PYENV     install .
...
````
1. Error comes from pip ...
````
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
````
2. is a dependency error ..
````
readme-renderer 26.0 requires Pygments>=2.5.1, but you'll have pygments 2.1.3 which is incompatible.
````

3. I guess comes from #2112

````
ERROR: centos-: uWSGI not yet implemented
````
4. error is from sphinx-autobuild, which has changed the command line (solved by #2210)

````
PYENV     installing Sphinx
SPHINX    docs-live --> file:///share/test123/searx/dist/docs
usage: sphinx-autobuild [-h] [--port PORT] [--host HOST]
                        [--re-ignore RE_IGNORE] [--ignore IGNORE]
                        [--no-initial] [--open-browser] [--delay DELAY]
                        [--watch DIR] [--pre-build COMMAND] [--version]
                        sourcedir outdir [filenames [filenames ...]]
sphinx-autobuild: error: unrecognized arguments: --poll -B
make: *** [Makefile:88: docs-live] Fehler 2
````

dalf:
Maybe not related: see the PR https://github.com/searx/searx/pull/2154

(also, WIP here to upgrade pygments : https://github.com/dalf/searx/tree/upgrade-pygments )

dalf:
Since commit fd5fe369844e481aecc0d731b08ee8b29c9b47e6, ```make gh-pages``` works despite the warning about the pygments version.

See https://github.com/searx/searx/tree/gh-pages and https://searx.github.io/searx/

return42:
Strange, ATM I get:

    ERROR:searx.engines:Engine config error: ambigious shortcut: apkm

If I revert #2230 and #2232 everything works fine ... including a well formed admin/engines.html  .. I can't reproduce the issue of the both PRs.

dalf:
Which Python version are you using ? ( = Which Sphinx version are you using ?)

[...]

PR #2206, opened 2020-09-19
[mod] upgrade pygments

dalf:
## What does this PR do?

* upgrade pygments to the last version
* add `searx_data/update/update_pygments.py` to update the css styles of the oscar and simple themes.
    * the script creates a `pygments.less` file for each theme
    * `code.less` includes `pygments.less`
* oscar theme: fix the `.highlight` CSS class. In the master branch:
    * one usage hightlights the words from the user query (using bold font)
    * another usage highlights the source code. This PR renames this CSS class to `.highlight-code`, so the source code don't use a bold font (but perhaps it was intended ?)
* the pygments styles are nearly identical.

The version 2.1.3 is 5 years old, and starts to create incompatibility problem. See https://github.com/searx/searx/issues/2204

Most probably it is possible to the old versions of the different dependencies, but it would be good to be able to upgrade easily.
Currently it is difficult because the pygments CSS are hard-coded into the simple and oscar themes:
* https://github.com/searx/searx/blob/master/searx/static/themes/simple/less/code.less
* https://github.com/searx/searx/blob/master/searx/static/themes/oscar/less/logicodev/code.less
* https://github.com/searx/searx/blob/master/searx/static/themes/oscar/less/pointhi/code.less

The ```.highlight *``` CSS are actually the result of a Python method call: ```HtmlFormatter(style=style).get_style_defs(cssclass)```.

This PR adds a script to easily update the themes ( utils/update-pygments.py ).

With all the themes & style, search for ```!scc searx```.
Compare visually the result with a searx instance without this PR.
The results should be identical:
* mouse cursor on code, line number
* being able to select the code without the line number.

## Additional note

This PR commits the `pygments.less` files, so `searx_data/update/update_pygments.py` must be called after each upgrade of pygments.

Another PR can improve that:
* the `pygments.less` files are not commited (in .gitignore)
* there are generated each time when the themes are built.
* since the built css are commited too, it doesn't solve the issue.

## Related issues

Related to:
* #2204
* #2034


searx/static/themes/oscar/less/logicodev/results.less
```diff
@@ -33,12 +33,10 @@
...
     font-size: 13px;
+}
 
-
-    .highlight {
-        font-weight:bold;
-    }
-
+.result .highlight {
+    font-weight: bold;
```
dalf: Highlight the words from the query for all result types (include source code).

kvch (dismissed):
LGTM. Nit: is it possible to leave a bit more space between line numbers and code lines?

dalf:
Rebased #2346

Updated to leave 8px instead of 5px between line numbers and code lines:

image

PR #2209, opened 2020-09-20
Allow !engine to be at any position in search query

### LINKED ISSUE #363: Parse !engine or :language anywhere in the search query

0xhtml:
## What does this PR do?

With this PR bangs (e.g. !engine or :language) can be at any position inside of the search query.

For many users it might be convenient to place the !engine at the end of the query.

Try searching using the engine bang at different positions in the query.

## Related issues

Fixes #363

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -39,7 +39,8 @@ class RawTextQuery:
         if disabled_engines:
             self.disabled_engines = disabled_engines
 
-        self.query_parts = []
+        self.search_query = ''
+        self.bang_query = ''
         self.engines = []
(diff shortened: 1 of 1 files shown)


0xhtml:
Seems like there already is an implementation: #1915
Closing in favor of other PR.

Issue #2176, opened 2020-09-07
ERROR: URL http://127.0.0.1:8888 not available, check searx & uwsgi setup!


Volence:
I seem to be having an issue with the autoinstall scripts on an Ubuntu 20.04 server. Before the end the only error I see is `  |searx| ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /usr/local/searx`, not sure if that's the cause of the issue. The output looks like the following:

```shell
====================
searx@192.168.1.251 
====================

Install searx@192.168.1.251  (service)
======================================

installation of packages
------------------------

package(s)::

   virtualenv python3-dev python3-babel python3-venv uwsgi
   uwsgi-plugin-python3 git build-essential libxslt-dev zlib1g-dev
   libffi-dev libssl-dev shellcheck python-is-python3

Should packages be installed? [YES/no] 
Yes
Hit:1 http://us.archive.ubuntu.com/ubuntu focal InRelease
Get:2 http://us.archive.ubuntu.com/ubuntu focal-updates InRelease [111 kB]
Get:3 http://us.archive.ubuntu.com/ubuntu focal-backports InRelease [98.3 kB]
Get:4 http://us.archive.ubuntu.com/ubuntu focal-security InRelease [107 kB]
Fetched 317 kB in 0s (811 kB/s)    
Reading package lists... Done
Building dependency tree       
Reading state information... Done
All packages are up to date.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'libxslt1-dev' instead of 'libxslt-dev'
build-essential is already the newest version (12.8ubuntu1).
git is already the newest version (1:2.25.1-1ubuntu3).
libffi-dev is already the newest version (3.3-4).
libssl-dev is already the newest version (1.1.1f-1ubuntu2).
libxslt1-dev is already the newest version (1.1.34-4).
python-is-python3 is already the newest version (3.8.2-4).
python3-dev is already the newest version (3.8.2-0ubuntu2).
zlib1g-dev is already the newest version (1:1.2.11.dfsg-2ubuntu1).
python3-venv is already the newest version (3.8.2-0ubuntu2).
shellcheck is already the newest version (0.7.0-2build2).
uwsgi is already the newest version (2.0.18-11ubuntu1).
uwsgi-plugin-python3 is already the newest version (2.0.18-11ubuntu1).
virtualenv is already the newest version (20.0.17-1).
[...]

PR #2211, opened 2020-09-20
Bugfix of:  ERROR: centos-: uWSGI not yet implemented

return42:
Since there is no centos-7 documentation 'searx.sh doc' produce an error message::

    ERROR: centos-: uWSGI not yet implemented

This patch removes CentOS from  the docs build. 

OT: 

- CentOS 7 support was added by  #2176 
- CentOS 8 is the same as RHEL. 
- Adding support for CentOS 7 was only needed to distinguish between (the old) yum and dnf package managers
- The "yum vs dnf" topic is the same to Fedora & RHEL
- Adding a CentOS 8 container might be a little more up-to-date

To test run:

    make clean make docs

Partial patch of  #2204

### DIFF
--- a/utils/searx.sh
+++ b/utils/searx.sh
@@ -826,7 +826,7 @@ rst-doc() {
     # I use ubuntu-20.04 here to demonstrate that versions are also suported,
     # normaly debian-* and ubuntu-* are most the same.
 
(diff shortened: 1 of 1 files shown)


dalf:
Thank you!

PR #2212, opened 2020-09-21
Some fixes in fetch languages script

MarcAbonce:
## What does this PR do?

I was late to answer the comment in https://github.com/searx/searx/pull/2137#discussion_r484901563, but this PR tries to fix that issue and a couple other things I found on the way:

* Add language names in Qwant's fetch languages function.
* Add supported language list for Startpage.
* Small fix in Google to fetch search languages rather than UI languages.
* Fix bug where language name was not set.
    In this case, the error was caused by the Corsican language, which is not known by the babel library. If this happens, I'm checking Wikipedia which has both the local and English language names for practically every language that could ever be added to `languages.py`.
* Increase the engine counter's threshold a little bit.
    After fixing the above issue, Corsican was added to the language list, but when I tried to run a couple of queries in Qwant (which claims to support Corsican), I found that almost all the results were in French. I've also noticed the same with some engines like Dailymotion and Peertube that allow a lot of languages in their APIs even if they have no videos in them yet.
* Fix small issue when selecting default country for a language.
    Previous version of the script would sometimes return unusual country choices like French Basque but no Spanish Basque or Montenegrin Serbian but no Serbian Serbian. This happens because very few engines contain language country pairs, so with less common languages one engine can significantly skew the results.
* Make `write_languages_file` function use `pprint` instead of ugly string concatenation.

In the end, nothing has really changed. The new `languages.py` contains exactly the same languages as the old one. There's only a few differences in a few language's country code and the change in format caused by `pprint`.

However, the `fetch_languages` script doesn't break or ignore any language anymore, even if Corsican were to be added in the future.

Pick a language and run some queries on it.

## Related issues
#2137


dalf (approved):

dalf:
Thank you !

PR #2219, opened 2020-09-23
Fix small inconsistencies in about page

resynth1943:
See diff.

### DIFF
--- a/searx/templates/__common__/about.html
+++ b/searx/templates/__common__/about.html
@@ -7,7 +7,7 @@
     while not storing information about its users.
   </p>
 
-  <p>More about searx ...</p>
+  <p>More about searx...</p>
 
   <ul>
     <li><a href="https://github.com/asciimoo/searx">github</a></li>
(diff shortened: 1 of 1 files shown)


searx/templates/__common__/about.html
```diff
@@ -49,9 +49,9 @@ <h2>Technical details - How does it work?</h2>
 
     It provides basic privacy by mixing your queries with searches on other
     platforms without storing search data. Queries are made using a POST request
-    on every browser (except chrome*).  Therefore they show up in neither our
-    logs, nor your url history.  In case of Chrome* users there is an exception,
-    searx uses the search bar to perform GET requests.
+    on every browser (except Chromium-based browsers*).  Therefore they show up 
+    in neither our logs, nor your url history.  In case of Chrome* users there
```
dalf: There is a Chrome* reference here

resynth1943: Fixed in b46bded.

dalf:
Thank you !

PR #2224, opened 2020-09-28
[enh] update infobox engines

dalf:
### duckduckgo_definition engine

The duckduckgo definition results are user friendly in many cases.

This PR fixes this issue.

### wikidata engine

The wikidata is rewritten to use one SPARQL request instead of two HTTP requests. It should be faster (bellow 2 seconds, usually around 1 second instead of time out).

The generated request : https://gist.github.com/dalf/9d17bcb57b817c609111102e7d1c25b0
The parameters are on the lines 7, 8, 9 and 127
To execute the query: https://query.wikidata.org/
Notes:
* the SPARQL query won't fit in the query parameter, so the query is sent using a POST request. The consequence is wikidata doesn't cache the results which may be good for the privacy.
* it is possible to see how the query is optimized using the `debug_explain_wikidata_query` function (it is dead code on runtime, but it may provide some help to debug / understand why the query is slow)

The query returns only the values but not the property names.

The properties names are loaded when the engine is initialized: 
https://github.com/searx/searx/blob/43613f35c564ba7185009dd2a9928258d7e18c2f/searx/engines/wikidata.py#L80-L92

The result is stored into WIKIDATA_PROPERTIES, but there are some hard-coded values (and the unit symbol see bellow):
https://github.com/searx/searx/blob/43613f35c564ba7185009dd2a9928258d7e18c2f/searx/engines/wikidata.py#L34-L45

### new file searx/data/external_urls.json

This PR adds a new file, and new a module searx.external_urls.
This module provides a the function get_external_url : 
```python
def get_external_url(url_id, item_id, alternative="default")
```

These URL are store in different places:
* the URL are ignored in the master branch implementation. The new implementation (note: _external_urls.json uses the ddg id_, it should be a problem):
https://github.com/searx/searx/blob/43613f35c564ba7185009dd2a9928258d7e18c2f/searx/engines/duckduckgo_definitions.py#L178-L185
* the wikidata engine in the master branch : https://github.com/searx/searx/blob/5707fba5618f3e1e31d9b0656e02eca594cd3131/searx/engines/wikidata.py#L151-L168

This file and module allow to have one central to update these URL.

More over, it is easy to add alternative implementation (like nitter, individious, etc..).

Out of topic note: DOI resolver URL may fit into this file :
https://github.com/searx/searx/blob/5707fba5618f3e1e31d9b0656e02eca594cd3131/searx/settings.yml#L1059-L1063

### new file searx/data/wikidata_units.json

Both engines return references to unit using wikidata entities.

For example in ddd:
```{'unit': 'http://www.wikidata.org/entity/Q712226', 'amount': '+20.99'}```

This new file stores the symbol for each wikidata unit.

It can be updated using ```utils/fetch_wikidata_units.py```.

Out of topic note: in the same spirit than the currency engine, another offline unit conversion engine can implemented using this file (the wikidata entities contains, see https://www.wikidata.org/wiki/Property:P2370 https://www.wikidata.org/wiki/Property:P2442 and https://www.wikidata.org/wiki/Q712226 )

### change in searx/results.py

About infoboxes merge: 
* In the master branch,  the attributes are merged when the label is equals, and the URL are merged when there more or less equals ( compare_urls ).
* In this PR, in additition to the existing conditions, the attributes and URLs when the wikidata entities are the same. 

See:
* https://github.com/searx/searx/blob/43613f35c564ba7185009dd2a9928258d7e18c2f/searx/engines/duckduckgo_definitions.py#L221
* https://github.com/searx/searx/blob/43613f35c564ba7185009dd2a9928258d7e18c2f/searx/engines/duckduckgo_definitions.py#L197

Without this additional merge condition, when a user request `:es !wd !ddd new york`, the infobox will contain the area twice: one with a English label, one with a Spanish label.

The wikidata engine is often the slowest engine in the general category. Users may want to use a faster infobox engine. The alternative is the duckduckgo_engine. Unfortunately, the ddd engine is only really helpful with the english language.

Moreover the version in the master branch returns un-parsed result. This PR fixes that and try to improve the response time of the wikidata engine.

* ```!ddd new york city```: 
  * the open street map opens OSM with the right zoom
  * the ```area``` field is readable
* ```!ddd pink floyd```
  * the twitter/ facebook/... links are links
* ```!ddd ip```
  * no answers (the server IP). related to  #2197 (it does NOT fix the issue)
* ```!wd !ddd bubalus``` : https://github.com/searx/searx/issues/669
* ```!wd !ddd we bare bears``` : https://github.com/searx/searx/issues/1097

* Oscar theme / official website link: 
   * check that the official link is displayed just bellow the infobox title.
   * the full URL is displayed according this commit: https://github.com/searx/searx/commit/74270a3fc2db9637d0f997282f092ddfdd3a72c7 (see  searx/templates/oscar/infobox.html )
   * it was initially the domain name: https://github.com/searx/searx/commit/29960aa1d912f365846731c9aca166b5b7879d02
   * so the `domain` is not set : https://github.com/searx/searx/blob/29960aa1d912f365846731c9aca166b5b7879d02/searx/engines/wikidata.py#L420

## Related issues

N/A

### CODE REVIEW

**Comment by kvch** [comment]
Are your Draft PRs supposed to be reviewed? Or are these really just drafts?

**Comment by dalf** [comment]
I didn't removed the "Draft" tag because I will most probably rewrite the commit history (but I can do at the end after the review) and it lacks some explanations.

But a first review can be done: the code won't change a lot.

Infoboxes are basically an infinite task: there are always some glitches here and there, but I think it is better.

**Comment by return42** [comment]
> I didn't removed the "Draft" tag because I will most probably  rewrite the commit history

Really does It? .. I do not have any experience, but as far I know, the tag is from github not git itself. How did you add it? .. do you have a link from github describing the draft feature? (excuse my naive interference)

**Comment by dalf** [comment]
@return42 : when I have started the PR I wasn't sure if I was going to split the PR into different PR (like the Drop Python 2 PR). Yes, this is github tag which blocks a merge into the master ( https://github.blog/2019-02-14-introducing-draft-pull-requests/ ) : it is not related to the git history / rebase / history rewrite.

**Comment by dalf** [comment]
It should be ready except:
* Wikidata: date are not always formatted.
* Some comments are missing.
* The commit history is messy.

The big question: the wikidata engine is faster according to my tests, but is it the case for everyone ?

File: searx/engines/duckduckgo_definitions.py [hunk 1]
```diff
@@ -10,30 +10,57 @@
 """
 
 import json
+import re
+import math
```
kvch: This import is unused.

searx/engines/duckduckgo_definitions.py
```diff
@@ -10,30 +10,57 @@
 """
 
 import json
+import re
```
kvch: This import is unused.

searx/engines/duckduckgo_definitions.py
```diff
@@ -10,30 +10,57 @@
 """
```
kvch: Could you please clean up the indentation in this file? The way dictionaries are indented are inconsistent and ugly. I know it is old code, but please take care of this. :)

```python
 results.append({
                'infobox': heading,
                'id': infobox_id,
                'content': content,
                'img_src': image,
                'attributes': attributes,
                'urls': urls,
                'relatedTopics': relatedTopics
                })
```
could be
```python
results.append({'infobox': heading,
                'id': infobox_id,
                'content': content,
                'img_src': image,
                'attributes': attributes,
                'urls': urls,
                'relatedTopics': relatedTopics})
```

dalf: 💯

searx/engines/wikidata.py
```diff
@@ -3,501 +3,791 @@
...
+        attributes.append(WDImageAttribute(name, url_id, priority))
 
-    # INFOBOX ATTRIBUTES (ROWS)
+    def add_date(name):
+        attributes.append(WDDateAttribute(name))
 
     # DATES
     # inception date
-    add_attribute(attributes, id_cache, 'P571', date=True)
+    add_date('P571')
```
kvch: These could be refactored to a for loop:
```python
date_properties = [
    'P571',  # inception date
    'P576',  # dissolution date
# ....
]

for property in date_properties:
    add_date(property)
```

The same goes for `add_label`, `add_value`, `add_url` and `add_image`.

kvch (commented):
Small comments, overall looks good.

dalf:
@kvch , thank you for the feedback.

May I ask you the response time of the wikidata engine ?

server-timing

kvch:
It is between 1.5-1.8:

wikidata

searx/engines/wikidata.py
dalf: Fixed in the last commit

searx/engines/duckduckgo_definitions.py
dalf: Done

dalf:
I have updated the commit history: based on the last commit of the master branch, more meaningful commit messages.

kvch (approved):
Awesome PR! I tested it and everything seems to work as intended.

Issue #1559, opened 2019-04-11
Embedded searx-checker


dalf:
I've updated searx-checker to produce an html output, see https://searx.space
Source code based on work of @kvch : https://github.com/searx/searx-checker

It shows which engines currently doesn't work (warning : some engine may work / not work from time to time).

The thing that would be nice is to be able test the different searx instances.
Example : the first instance in https://stats.searx.xyz/ is https://searx.at
This instance answers quickly with the last version searx and good security configuration. 
But this instance doesn't work : whatever the engine is, whatever the query is, it always shows the same error message. It would be nice to show this information on stats.searx.xyz

Using searx-checker doesn't work : most the searx engines disable and severly limit json output.

A solution is to embed searx-checker inside searx. 
Every day searx makes a self test of all the configured engines. It can be done easily if each engine contains a list of queries that are supposed to provide answers (with a default one is not configured).
The result would be available on a /status URL : see https://searx.space/output/localhost.json for an example.

Then stats.searx.xyz can display a sum up of this information : 
* on the current view : number of working engine per instance (perhaps with a detail per category).
* another view (to create) : number of working instance per engine.

unixfox:
That's a good idea! I often want to know which engine is not working on my instance. 
Meanwhile why not making instead a program written with a browser engine framework like selenium or puppeteer to check the availability of each engine on a searx instance?
I know it would take much longer and consume more ram and processor but that's the only way currently to avoid the json output restriction and other restrictions as well.

kvch:
@dalf I really like the interface you have put together. Are you planning to add it to upstream?

@unixfox I think "integrating" searx-checker is a better idea. There is no need to use the JSON output of searx. All you need is a module which calls the `request` functions of every engine and check the results. Searx-checker only uses the JSON output, as it's an external tool. But once it's integrated, enabling JSON output is not required.

dalf:
@kvch Thanks! 
Simple answer: I don't know when, but yes it would be great (no more searx-checker project, only searx).

For now, I've made a quick an dirty integration as you may have seen : 
* https://github.com/searx/searx-docker/blob/master/docker-compose.yaml
* https://github.com/searx/searx-checker/blob/master/loop.sh 
* result here: https://myinstance/status (access even by external javascript).

Nice to have : an API endpoint to get by engine
* the status (give response or not)
* the error rate (the engine may work but not always)
[...]

PR #2225, opened 2020-09-28
Processors

dalf:
## What does this PR do?

Create one "processor" (*) per one class per engine type:
* offline
* online
  * online_currency
  * online_dictionnary

Each processor has two methods:
```python
   def get_params(self, search_query, category):
      pass

   def search(self, query, params, result_container, start_time, timeout_limit):
      pass  
```

[online_currency](https://github.com/searx/searx/blob/f938072b5451cdcbea506707c4d4af1ccbcc333c/searx/search/processors/online_currency.py) and [online_dictionnary](https://github.com/searx/searx/blob/f938072b5451cdcbea506707c4d4af1ccbcc333c/searx/search/processors/online_dictionary.py) checks if the query match the excepted format ( equivalent of #2159 ).

There is [one processor per engine](https://github.com/searx/searx/blob/f938072b5451cdcbea506707c4d4af1ccbcc333c/searx/search/processors/__init__.py).

Once merged, it will be easier to implement
* https://github.com/searx/searx/issues/1559
* https://github.com/searx/searx/pull/477
* and most probably https://github.com/searx/searx/issues/2067

* activate `` and `` engines, then search for `en-es house`
* search for `10 eur in jpy`
* make a query using the `time_range` parameter
* make a query using the `language` parameter
* activate the `command` engine with the following configuration, and make sure it is working:
```yaml
  - name: locate
    engine: command
    command: ['locate', '--existing', '--ignore-case', '{{QUERY}}']
    shortcut: locate
    tokens: []
    disabled: False
    delimiter:
        chars: ' '
        keys: ['line']
```


searx/search/processors/online.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # add params from search_query (or engine)
+        if hasattr(self.engine, 'language') and self.engine.language:
+            params['language'] = self.engine.language
+        else:
+            params['language'] = search_query.lang
+
+        params['safesearch'] = search_query.safesearch
+        params['time_range'] = search_query.time_range
+        return params
```
dalf, line 53: ```language```, ```safesearch``` and ```time_range``` could be set in the ```abstract.py```, so ```params``` would have these values even for the offline engines.

dalf: Done with code pushed force (based on the master branch).

searx/search/processors/online.py
```diff
@@ -0,0 +1,165 @@
...
+from searx.search.processors.abstract import EngineProcessor
+
+
+logger = logger.getChild('search.processor.online')
+
+DEFAULT_PARAMS = {
+    'method': 'GET',
+    'headers': {},
+    'data': {},
+    'url': '',
```
dalf, line 26: Could be `None`

searx/search/processors/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+    engine_type = getattr(engine, 'engine_type', 'online')
+    processor_class = get_processor_class(engine_type)
+    if processor_class:
+        return processor_class(engine, engine_name)
+    else:
+        return None
+
+
+def initialize(engine_list):
+    engines.initialize_engines(engine_list)
```
dalf, line 35: initialize_engines uses threading, it would be nice to move the threading here.

so:
* searx.engines is about engine description
* searx.search.processors is about how to process the search queries (initialization & run).

searx/search/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+    def _get_requests(self):
+        # init vars
+        requests = []
+
+        # max of all selected engine timeout
+        default_timeout = 0
+
+        # start search-reqest for all selected engines
+        for engineref in self.search_query.engineref_list:
+            processor = processors[engineref.name]
```
dalf, line 162: Raise an exception if there is processor for one engine : it should not happen.

searx/search/processors/online_dictionary.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+from searx.utils import is_valid_lang
+from .online import OnlineProcessor
+
+
+parser_re = re.compile('.*?([a-z]+)-([a-z]+) ([^ ]+)$', re.I)
+
+
+class OnlineDictionaryProcessor(OnlineProcessor):
```
dalf, line 12: is OnlineTranslatorProcessor a better name since two languages are required ?

dalf:
Once merge, it will be easier to implement https://github.com/searx/searx/issues/1559 and https://github.com/searx/searx/pull/477

PR #2239, opened 2020-10-02
[mod] preferences.py: check language setting with a regex instead of match_language

### LINKED ISSUE #2220: language parameter: validity
From https://github.com/searx/searx/pull/2189#discussion_r494123787

When the language parameter is set from the cookies, the validity is checked using the match_language function:
https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/preferences.py#L165-L168

When the language parameter is set from the URL or from the query, the validity is checked using a regex:
https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/search.py#L281-L290

The regex is on purpose, see commit https://github.com/searx/searx/commit/fd65c1292179fb082e965a1ee6a88b9298a54fc1

Should we changed preferences to check the validity using a regex instead of match_language ?

ping @MarcAbonce

dalf:
## What does this PR do?

 searx/preferences.py : check the value of the `language` cookie using a regex instead of the [match_language](https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/utils.py#L219) function.

See issue  #2220 and the commit https://github.com/searx/searx/commit/fd65c1292179fb082e965a1ee6a88b9298a54fc1

Without this PR, when the language parameter is set from the cookies, the validity is checked using the match_language function:
https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/preferences.py#L165-L168

When the language parameter is set from the URL or from the query, the validity is checked using a regex:
https://github.com/searx/searx/blob/21dbc7e852a5c48097c0067e78f0600972e54823/searx/search.py#L281-L290

With this PR, the `language` value check is the same (cookie, form value). See note below.

* `make test`
* check different values for the `language` cookie.

## Related issues

Close #2220

## Note

https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/webadapter.py#L49-L64

With this PR, `query_lang = form.get('language')` is redundant since `request.form` is already parsed: https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/webapp.py#L448 (it was already the case for `language` value recognized by the match_language function). The code could be:

```python
def parse_lang(preferences: Preferences, form: Dict[str, str], raw_text_query: RawTextQuery) -> str:
    # get language
    # set specific language if set on request, query or preferences
    # TODO support search with multible languages
    if len(raw_text_query.languages):
        query_lang = raw_text_query.languages[-1]
    else:
        query_lang = preferences.get_value('language')

    return query_lang
```

I think it is better to have another PR to deal with that since the `form` parameter can be removed the `get_search_query_from_webapp` function.

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -10,7 +10,7 @@ from urllib.parse import parse_qs, urlencode
 
 from searx import settings, autocomplete
 from searx.languages import language_codes as languages
-from searx.utils import match_language
+from searx.webutils import VALID_LANGUAGE_CODE
 
 
 COOKIE_MAX_AGE = 60 * 60 * 24 * 365 * 5  # 5 years
@@ -162,9 +162,7 @@ class SearchLanguageSetting(EnumStringSetting):
     """Available choices may change, so user's value may not be in choices anymore"""
 
     def _validate_selection(self, selection):
-        if selection != "" and not match_language(
-                # pylint: disable=no-member
-                selection, self.choices, fallback=None):
+        if selection != '' and not VALID_LANGUAGE_CODE.match(selection):
             raise ValidationException('Invalid language code: "{0}"'.format(selection))
 
     def parse(self, data):
@@ -181,6 +179,7 @@ class SearchLanguageSetting(EnumStringSetting):
                 data = lang
             else:
                 data = self.value
+        self._validate_selection(data)
         self.value = data
 
 
--- a/searx/query.py
+++ b/searx/query.py
@@ -22,9 +22,7 @@ import re
 from searx.languages import language_codes
 from searx.engines import categories, engines, engine_shortcuts
 from searx.search import EngineRef
-
-
-VALID_LANGUAGE_CODE = re.compile(r'^[a-z]{2,3}(-[a-zA-Z]{2})?$')
+from searx.webutils import VALID_LANGUAGE_CODE
 
 
 class RawTextQuery:
--- a/searx/webadapter.py
+++ b/searx/webadapter.py
@@ -1,6 +1,7 @@
 from typing import Dict, List, Optional, Tuple
 from searx.exceptions import SearxParameterException
-from searx.query import RawTextQuery, VALID_LANGUAGE_CODE
+from searx.webutils import VALID_LANGUAGE_CODE
+from searx.query import RawTextQuery
 from searx.engines import categories, engines
 from searx.search import SearchQuery, EngineRef
 from searx.preferences import Preferences
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -11,6 +11,8 @@ from codecs import getincrementalencoder
 from searx import logger
 
 
+VALID_LANGUAGE_CODE = re.compile(r'^[a-z]{2,3}(-[a-zA-Z]{2})?$')
+
 logger = logger.getChild('webutils')
 
 


searx/preferences.py
```diff
@@ -162,9 +162,7 @@ class SearchLanguageSetting(EnumStringSetting):
     """Available choices may change, so user's value may not be in choices anymore"""
 
     def _validate_selection(self, selection):
-        if selection != "" and not match_language(
-                # pylint: disable=no-member
-                selection, self.choices, fallback=None):
+        if selection != '' and not VALID_LANGUAGE_CODE.match(selection):
```
MarcAbonce, line 165: The first validation is not really necessary anymore because the regex should not match with an empty string. But it's also fine if you want to keep it just to be extra sure.

MarcAbonce (approved):
It looks good to me.
The redundant form thing could be included here or in a separate PR. Either way would be ok for me.

searx/preferences.py
dalf: In the default configuration `selection == ''`:
https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/settings.yml#L8
and
https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/preferences.py#L323-L326

So with only `if not VALID_LANGUAGE_CODE.match(selection):`
* `make test` fails (test_webadapter.py fails)
* the web server crashes with:
```
  File "/home/alexandre/code/searx/local/py3/lib/python3.8/site-packages/flask/app.py", line 2242, in preprocess_request
    rv = func()
  File "/home/alexandre/code/searx/searx/webapp.py", line 427, in pre_request
    preferences = Preferences(themes, list(categories.keys()), engines, plugins)
  File "/home/alexandre/code/searx/searx/preferences.py", line 322, in __init__
    'language': SearchLanguageSetting(
  File "/home/alexandre/code/searx/searx/preferences.py", line 44, in __init__
    self._post_init()
  File "/home/alexandre/code/searx/searx/preferences.py", line 80, in _post_init
    self._validate_selection(self.value)
  File "/home/alexandre/code/searx/searx/preferences.py", line 166, in _validate_selection
    raise ValidationException('Invalid language code: "{0}"'.format(selection))
searx.preferences.ValidationException: Invalid language code: ""
```

---

When language can be '' in the Preferences, there is a fallback in the webapp :
https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/webapp.py#L454-L455

And _get_browser_or_settings_language falls back to `en` (on the first and the last lines):
https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/webapp.py#L157-L163

So.... in this commit I would keep the `selection != ''`, but another PR can:
* Preferences takes the default value for locale and language from the constructor.
* webapp.pre_request
    * get the defaut values for locale and language from the HTTP headers
    * fallback to `en`
    * initialize the Preferences object with these default values.
* webapp.get_locale relies only on `request.preferences.get_value('locale')`

MarcAbonce: Yeah, you're right. My mistake.

Issue #715, opened 2016-10-02
HTTP headers


dalf:
searx includes contents from others websites. The contents are escaped, trimed and all, but there is always a risk. The security could be improve using HTTP headers. 

Some headers that could be added : 

`X-Frame-Options: SAMEORIGIN`
when there is the image proxy (break the included videos, in this case a more precise).
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

`X-Content-Security-Policy "default-src 'self'; img-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self'; script-src 'self' 'unsafe-inline'; connect-src 'self';"`
to be adjust when there are images, videos
https://content-security-policy.com/
https://report-uri.io/

`X-Content-Type-Options: nosniff`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options

`X-XSS-Protection "1; mode=block"`
https://blogs.msdn.microsoft.com/ieinternals/2011/01/31/controlling-the-xss-filter/

`X-Download-Options: noopen`
https://msdn.microsoft.com/fr-fr/library/jj542450(v=vs.85).aspx#noOpenDirective

more or less out of topic
`X-Robots-Tag: noindex, nofollow`
http://robots-txt.com/x-robots-tag/
avoid to be indexed and then spam (useless for that purpose ?)

HSTS header ( see https://github.com/asciimoo/searx/issues/62 )
https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security
https://hstspreload.appspot.com/

Should nginx / apache send the headers or should the python code should send them ? in case of X-Content-Security-Policy and X-Frame-Options the values can change according to the search options.

cy8aer:
IMHO it would be better to have the web servers create the headers if you want to have additional headers (like key pinning stuff).  But because Apache/mod_uwsgi does not work with header add I hard code my headers in webapp.py by now which is not very easy because I need to edit webapp.py with every git pull. Is there an mod_proxy_uwsgi description for searx? mod_proxy_uwsgi makes web server side headers possible.

dalf:
A solution about hard coded headers in webapp is settings.yml. But it would be messy to have header inside settings.yml and inside the HTTP server.
I agree it should be in the web server, so it is related to the installation.

cy8aer:
[...]

PR #2242, opened 2020-10-03
[mod] URL for the static file contains the sha1

dalf:
## What does this PR do?

for example, instead of 
```html
<link rel="stylesheet" href="/static/css/bootstrap.min.css" type="text/css" />
```

the HTML becomes:
```html
<link rel="stylesheet" href="/static/css/bootstrap.min.css?0d0624d4eb8096f13cc9f981ed648b3a6624e532" type="text/css" />
```

* allow to cache the static files forever
* avoid bugs when the static files are updated but not reloaded

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -295,12 +295,19 @@ def get_result_template(theme, template_name):
 
 
 def url_for_theme(endpoint, override_theme=None, **values):
+    suffix = ""
     if endpoint == 'static' and values.get('filename'):
+        filename = values['filename']
+        # check if this is a static within the theme
         theme_name = get_current_theme_name(override=override_theme)
-        filename_with_theme = "themes/{}/{}".format(theme_name, values['filename'])
+        filename_with_theme = "themes/{}/{}".format(theme_name, filename)
         if filename_with_theme in static_files:
-            values['filename'] = filename_with_theme
-    url = url_for(endpoint, **values)
+            values['filename'] = filename = filename_with_theme
+        # add sha1
+        sha1 = static_files.get(filename)
+        if sha1:
+            suffix = "?" + sha1
+    url = url_for(endpoint, **values) + suffix
     if settings['server']['base_url']:
         if url.startswith('/'):
             url = url[1:]
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import io
 import os
 import csv
 import hashlib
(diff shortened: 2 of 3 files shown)


SuperSandro2000:
> allow to cache the static files forever

Can we set the headers for that in python? We should also recommend that in some doc so that people can advice their reverse proxies to add them.

dalf:
I agree that searx should returns good HTTP headers by default.

There is an issue about this : https://github.com/searx/searx/issues/715 part of [v1.0.0](https://github.com/searx/searx/projects/5)

About the docker image there is
https://github.com/searx/searx/blob/53c8d945b489b1185d89e02bbc6b6ea6b60a91da/dockerfiles/uwsgi.ini#L42
so the expires header is easy to modify.

Cache control header is less easy: https://uwsgi-docs.readthedocs.io/en/latest/InternalRouting.html?highlight=route-uri#the-internal-routing-table

At least it can be modified here https://github.com/searx/searx-docker/blob/7ad4d4903d0ce41de4094d5fe0fecda156b41eaa/Caddyfile#L65-L69

---

Note with https://github.com/searx/searx/issues/2067 : 
* flask is replaced by [starlette](https://www.starlette.io/templates/) ( [Jinja2Template](https://github.com/encode/starlette/blob/4d5708e4688ff4d89888224668d9435b3e9e7000/starlette/templating.py#L55-L64) )
* uwsgi is replaced by [uvicorn](https://www.uvicorn.org/)

asciimoo:
Setting http headers for static file caching isn't really the task of searx, I think. In production there is a webserver which handles static file serving, it should handle caching as well.

asciimoo:
@dalf what is the benefit of using this method over the traditional web server/header based caching?

dalf:
When a static file is updated, the browser will cache the file for an undefined time, especially on a mobile phone. It can be a problem when the static files fix a bug.

This is exactly what I have experienced with https://github.com/searx/searx/pull/2205 : the server was up to date, but on my mobile the feature wasn't working. After I cleaned the cache, it worked.

With the content identifier in the URL, there is no need to clean the cache, no need to care for the HTTP headers, the updated files are downloaded as soon there is new version. Moreover the HTTP server can be configured with a very long cache time.

asciimoo:
> When a static file is updated, the browser will cache the file for an undefined time, especially on a mobile phone. It can be a problem when the static files fix a bug.

With proper cache/etag headers it shouldn't be a problem.

dalf:
Chrome is not easy to deal with: https://stackoverflow.com/questions/43659756/chrome-ignores-the-etag-header-and-just-uses-the-in-memory-cache-disk-cache

SuperSandro2000:
> In production there is a webserver which handles static file serving, it should handle caching as well.

IIRC the docker images don't have that and caching improves performance by a lot.

I recently noticed that the official nextcloud docker images didn't have caching enabled and puting another proxy before it is not a great option just to set some headers.

Issue #2245, opened 2020-10-05
Request method preference not respected when addin Search Engine


michaelschufi:
1. I've changed the request method to GET in the preferences.
1. I've added searx to my Firefox browser as a search engine

Yet, it still searches using POST when using the address bar. What am I doing wrong?

dalf:
Once you added searx, the opensearch.xml (the POST/GET choice) is cached:
* try to remove searx from the search engine
* re-add the searx

---
Probably related to https://github.com/searx/searx/issues/2069#issuecomment-671099077

michaelschufi:
Yeah, I've tried that. Even tried the "forget about this site" feature. But it still didn't work :/

michaelschufi:
Seems to be working now. Tested it on a.searx.space.

PR #2247, opened 2020-10-05
[fix] opensearch.xml URL contains method and autocomplete parameters

dalf:
## What does this PR do?

When the user add searx as a search engine, the browser loads the /opensearch.xml URL without the cookies.
Without the query parameters, the user preferences are ignored (method and autocomplete).

In addition, opensearch.xml is modified to support automatic updates,
see https://developer.mozilla.org/en-US/docs/Web/OpenSearch

The user choice between the GET and POST method is ignored in opensearch.xml
This PR fixes this issue.

With Firefox, with Chrome (with Safari ?):
* Add the instance without preferences
* Check the browser uses the 'POST' method (if default) 
* Remove the instance
* Change to the 'GET' method
* Add the instance again
* Check the browser uses the 'GET' method 

## Related issues

* https://github.com/searx/searx/issues/2245
* https://github.com/searx/searx/issues/2069#issuecomment-671099077

### DIFF
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -15,4 +15,8 @@
   {% if autocomplete %}
     <Url rel="suggestions" type="application/x-suggestions+json" template="{{ host }}autocompleter?q={searchTerms}"/>
   {% endif %}
+
+  <Url type="application/opensearchdescription+xml"
+      rel="self"
+      template="{{ opensearch_url }}" />
 </OpenSearchDescription>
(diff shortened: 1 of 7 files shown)


dalf:
* Firefox: ok
* Chrome always use the GET method. Related to https://github.com/searx/searx/issues/1666

I guess this has to be implemented (from https://github.com/searx/searx/issues/1666#issuecomment-657271726 ) :
> Perhaps we could switch the GET search url to / and use /search for doing POST requests. In this case the opensearch should work and we could still do POST requests if the search isn't performed from the URL bar.

dalf:
I merge the PR: the Chrome issue is about the content of opensearch.xml not about URL of opensearch.xml

michaelschufi:
@dalf 
When will this be released?

dalf:
Searx is more or less using rolling release.

On searx.space, you can check for the version 0.17.0-164 or later (but it might not work on instances with a custom theme).

Example including this PR:
* https://searx.lelux.fi/
* https://searx.likkle.monster/
* https://searx.deepak.pro/
* https://searx.be/
* https://metasearch.nl/

PR #2251, opened 2020-10-08
Add minimum safe search option based on #1392

kvch:
## What does this PR do?

This PR is the renovated version of #1392. It adds a new options to `settings.yml` called `minimum_safe_search`. Thus, instance administrators can control the minimum level of safe search on their instances. Possible values are `0`, `1` and `2`.

This PR is also superior to the previous because if an admin configures this option, only the engines which meet the criteria are displayed on the preferences page.

This PR gives more control to instance admins by letting them enforce minimal safe search requirements.

## Related issues

Closes #1392

### DIFF
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -194,6 +194,7 @@
                                 </tr>
                         {% for search_engine in engines_by_category[categ] %}
                             {% if not search_engine.private %}
+                            {% if minimum_safesearch > 0 and search_engine.safesearch or minimum_safesearch == 0 %}
                                 <tr>
                                     {% if not rtl %}
                                     <td class="onoff-checkbox">
@@ -219,6 +220,7 @@
                                     </td>
                                     {% endif %}
                                 </tr>
+                            {% endif %}
                             {% endif %}
                         {% endfor %}
                           </table>
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -4,6 +4,7 @@ general:
 
 search:
     safe_search : 0 # Filter results. 0: None, 1: Moderate, 2: Strict
+    minimum_safe_search : 0 # Enforce a minimum safe_search
     autocomplete : "" # Existing autocomplete backends: "dbpedia", "duckduckgo", "google", "startpage", "swisscows", "qwant", "wikipedia" - leave blank to turn it off by default
     default_lang : "" # Default search language - leave blank to detect from browser information or use codes from 'languages.py'
     ban_time_on_fail : 5 # ban time in seconds after engine errors
--- a/searx/webadapter.py
+++ b/searx/webadapter.py
@@ -76,7 +76,8 @@ def parse_safesearch(preferences: Preferences, form: Dict[str, str]) -> int:
         query_safesearch = preferences.get_value('safesearch')
 
     # safesearch : second check
-    if query_safesearch < 0 or query_safesearch > 2:
+    minimum_safe_search = int(settings['search'].get('minimum_safe_search', 0))
+    if query_safesearch < 0 or query_safesearch > 2 or query_safesearch < minimum_safe_search:
         raise SearxParameterException('safesearch', query_safesearch)
 
     return query_safesearch
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -350,10 +350,12 @@ def image_proxify(url):
 
 def render(template_name, override_theme=None, **kwargs):
     disabled_engines = request.preferences.engines.get_disabled()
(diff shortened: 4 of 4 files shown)


kvch:
Come to think about the feature, the instance admin already configures `safe_search`. Instead of configuring an "enforced" value, we should consider adding an option for admins where they could provide a list of options which cannot be overwritten by user preferences.

dalf:
> we should consider adding an option for admins where they could provide a list of options which cannot be overwritten by user preferences.

👍 

How to write that information ?


#### value and boolean
```yaml
    safe_search : 0 # Value not enforced
```

```yaml
    safe_search : [0, False] # Value not enforced
```

```yaml
    safe_search : [0, True] # Enforced value
```

#### value and choices
```yaml
    safe_search : 0 # No enforced value
```

```yaml
    safe_search : 0, [0,1,2] # No enforced value, equivalent
```

```yaml
    safe_search : 1, [1,2] # Limit the possible value to 1 or 2 with a default value of 1
```

searx/templates/oscar/preferences.html
```diff
@@ -194,6 +194,7 @@ <h3>{{ _('Engines') }}</h3>
                                 </tr>
                         {% for search_engine in engines_by_category[categ] %}
                             {% if not search_engine.private %}
+                            {% if minimum_safesearch > 0 and search_engine.safesearch or minimum_safesearch == 0 %}
```
dalf, line 197: To ease template implementation and for consistency, why not move these conditions just bellow :
https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/webapp.py#L811-L812
?

(line 196 can move too, but it is not related to this PR).

kvch: Agreed. I am closing this PR once we make a decision about the configuration.

kvch:
The options you described seem too intrusive to me. I would not change the existing settings at all. I would rather add a new option which expects a list of option names (from the Preferences page) which are locked.

For example:
```yaml
general:
  locked_preferences:
    - safesearch
    - method
```

dalf:
| cookie name / preference name  / query parameter name | YAML path |
| ------------------------ | --------- |
| locale			| ui.default_locale |
| theme			| ui.default_theme |
| oscar-style		| ui.theme_args.oscar_style |
| results_on_new_tab	| ui.results_on_new_tab |
| language		| search.default_lang |
| safesearch		| search.safe_search |
| autocomplete	| search.autocomplete |
| image_proxy		| server.image_proxy |
| method			| server.method |
|  doi_resolver	| hard coded to 'oadoi.org' |

Without some renaming it will super confusing (?). For example:
```yaml
ui:
    default_theme: "legacy"
general:
  locked_preferences:
    - theme
```

Perhaps:
```yaml
ui:
    default_theme: "legacy"
locked_preferences:
  - ui.default_theme
```
?

But then:
```yaml
search:
    safe_search : 0 # Filter results. 0: None, 1: Moderate, 2: Strict
    autocomplete : "" # Existing autocomplete backends: "dbpedia", "duckduckgo", "google", "startpage", "swisscows", "qwant", "wikipedia" - leave blank to turn it off by default
    default_lang : "" # Default search language - leave blank to detect from browser information or use codes from 'languages.py'
    ban_time_on_fail : 5 # ban time in seconds after engine errors
    max_ban_time_on_fail : 120 # max ban time in seconds after engine errors
```

Why ```search.safe_search``` is allowed and not ```search.ban_time_on_fail``` ? Yes, the later one is not in the preferences.

kvch:
I do not mind how we refer to locked options; by their name in settings.yml or in preferences. Searx can "translate" the options to preferences when these are initialized. But I think this option would not be widely used by admins, so I would rather just create a documentation page with the mapping of the options than add the translation functionality of searx.

> Why search.safe_search is allowed and not search.ban_time_on_fail ? Yes, the later one is not in the preferences.

I am not sure I understand this question. The whole point of this feature is to limit the options that can be configured using the Preferences page. `ban_time_on_fail` is not exposed, thus, there is no need to lock it.

dalf:
I've tried this scenario: _"I'm new to Searx, I would like to set and lock a default language and a default safe_search, how can I do that ?"_

If the language setting is named in two different ways inside the settings.yml, the admin will have to the read the documentation at the same time (even if it makes sense because on one side this the default value, on the other side this is the preferences). With the `search.default_lang` notation, once understood, there is no need to read the documentation at the same.

About ```ban_time_on_fail ``` : _"Ok I can lock safe_search, oh can I lock ban_time_on_fail  ? Ah yes it doesn't make sense, this is not a preference"_ (so just ignore this paragraph, sorry).

PR #2254, opened 2020-10-09
[fix] fix duckduckgo engine

dalf:
## What does this PR do?

- remove paging support: a "vqd" parameter is required between each request. This parameter is uniq for each request
- update the URL (no redirect), use the POST method
- language support: works if there is no **more than one request per minute**, otherwise it is ignored ! but it seems to work with curl, so most probably I'm missing something ?

* Fix the duckduckgo engine

* various requests with different languages on the ddg engine.


asciimoo (approved):
Great, thanks!

PR #2256, opened 2020-10-09
Force admins to set secret_key if debug mode is disabled

### LINKED ISSUE #819: Make secret_key default one that will fail if not set to a custom value
Currently, the `secret_key` default value is `ultrasecretkey` which is a valid value. Would it not be better to let the default value of this setting be one that will make searx fail to start? This will force the user to conciously change this setting to a secure value instead of accidentally forgetting to set this to something random and secure.

kvch:
## What does this PR do?

This PR checks if the `server.secret_key` is changed from the default value when searx is not running in debug mode.

From now on if you do not change the secret_key, the following error message is returned:

```
$ SEARX_DEBUG=0 python3 searx/webapp.py
server.secret_key have to be changed if not in debug mode
```

This check forces instance admins to configure a secret key for their instances.

## Related issues

Closes #819

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -85,3 +85,7 @@ if 'SEARX_SECRET' in environ:
     settings['server']['secret_key'] = environ['SEARX_SECRET']
 if 'SEARX_BIND_ADDRESS' in environ:
     settings['server']['bind_address'] = environ['SEARX_BIND_ADDRESS']
+
+if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
+    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
+    exit(1)
--- a/searx/settings_robot.yml
+++ b/searx/settings_robot.yml
@@ -8,7 +8,7 @@ search:
 server:
     port : 11111
     bind_address : 127.0.0.1
-    secret_key : "ultrasecretkey" # change this!
+    secret_key : "changedultrasecretkey"
     base_url : False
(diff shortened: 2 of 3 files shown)


asciimoo (approved):
Great, thanks!

return42:
Breaks the docs build

````
$ make docs
build searx/brand.py
build utils/brand.env
PYENV     install .
SPHINX    docs --> file:///800GBPCIex4/share/searx/dist/docs
Running Sphinx v3.2.1

Configuration error:
The configuration file (or one of the modules it imports) called sys.exit()
make: *** [Makefile:84: docs] Error 2
````

**Comment by NightMachinery** [comment]
I just installed searx according to the docs:
```
git clone https://github.com/searx/searx searx
cd searx

sudo -H ./utils/searx.sh install all
sudo -H ./utils/filtron.sh install all

# Service accessible at port 4004
```

And now:
```
~/code/misc/searx master
❯ ./utils/standalone_searx.py -h
```

The service at port 4004 works fine though. (BTW, any way to disable the frontend and just have the bare essentials for the CLI?)

return42:
I guess there was an old settings file in your /etc/searx ...

https://github.com/searx/searx/blob/e78bfd4d8b62a521e08029bb0e708b2bcec9958d/utils/searx.sh#L413-L419

https://github.com/searx/searx/blob/e78bfd4d8b62a521e08029bb0e708b2bcec9958d/utils/searx.sh#L514-L523

remove the settings.yml or just edit your secret key by hand should do the trick.

PR #2270, opened 2020-10-23
Let admins lock user preferences

kvch:
## What does this PR do?

This PR adds a new attribute to `Setting` in `Preferences` called `locked`. If a `Setting` is `locked`, users cannot change the default value configured in `settings.yml` using the `/preferences` page. Locked settings does not even show up there.

### How to configure it from `settings.yml`?

A new setting is added to the configuration file: `preferences.lock`. It expects a list of preferences form IDs from the `/preferences` page.

#### Example configuration

```yaml
preferences:
    lock: ['autocomplete', 'safesearch', 'method']
```

We had a long conversation about how to refer to the locked settings in #2251. I decided to go with the IDs from the preferences page not only because it is simpler to implement. But rather I think the "administrator journey" is the following when deciding to which setting to lock:
1. Look at the `/preferences` page to see what users can configure
2. Decide if they want to let users configure that specific setting
3. If not, right click on the element and copy their ID to `settings.yml`. If yes, move on to the next setting.

Furthermore, this approach avoids the possible confusion over which settings can be overwritten by users or not.

It is possible that admins do not want to let users customize specific user settings. For example they might lock the search method to POST to help users with their privacy by disabling riskier options.

1. Lock a user setting from `settings.yml`
2. Check if it shows up in the Preferences page
3. Check if you can change the setting from the search URL

## Related issues

Closes #1392
Closes #2251

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -35,9 +35,10 @@ class ValidationException(Exception):
 class Setting:
     """Base class of user settings"""
 
-    def __init__(self, default_value, **kwargs):
+    def __init__(self, default_value, locked=False, **kwargs):
         super().__init__()
         self.value = default_value
+        self.locked = locked
         for key, value in kwargs.items():
             setattr(self, key, value)
 
@@ -115,6 +116,9 @@ class MultipleChoiceSetting(EnumStringSetting):
         self.value = elements
 
     def parse_form(self, data):  # pylint: disable=missing-function-docstring
+        if self.locked:
+            return
+
         self.value = []
         for choice in data:
             if choice in self.choices and choice not in self.value:  # pylint: disable=no-member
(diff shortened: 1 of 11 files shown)


dalf:
Note : in searx/preferences.py :  these method doesn't check the `.locked` values:
* `parse_form(self, input_data)`
* `save(self, resp)`

So the user can hack the HTML and save some disallowed values in the cookies (see the [preferences function webapp.py](https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/webapp.py#L784-L796) ).

These value will be ignored by the `parse_dict(self, input_data)` method, so no big deal.

----
I guess (?) the purpose is to enforce some preferences ? (and not to enforce some default preferences that can be modified in some requests).

Even if the categories and language are hidden in the preferences, there is still an UI in the index page.

Moreover,  webadapter.py reads the safesearch, language and the categories directly parameters from the HTTP form.

See https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/webadapter.py#L223-L224

This is related to https://github.com/searx/searx/pull/2189#discussion_r493056581

kvch:
I've addressed the review notes.

asciimoo (approved):
LGTM, thanks!

Issue #2052, opened 2020-07-08
What is the data lifecycle ?


dalf:
Maybe I'm overthinking.

#### Which data ?

* DOI
  * https://github.com/asciimoo/searx/blob/52eba0c7210c53230fb574176d4bf1af771bf0b4/searx/settings.yml#L905-L911
  * Update script: No
  * Data source: various, https://www.wikidata.org/wiki/Q21980377 (the "official website" property) may help.
  * When it should be updated ? check every day (?)
  * When it is updated: never.
  * Is it a problem to not update ? ❌ outdated URL, disappointing user experience.
  * Is it a problem to update ? No.
* ```searx/data/bangs.json```:
  * Update script: none (can be updated automatically and then manually with some clean up)
  * Data source: [jivesearch](https://github.com/asciimoo/searx/pull/2027) (useful sources are DuckDuckGo bangs and Wikidata)
  * When it should be updated ? require manual checking / perhaps automatic checking (automatically extract the opensearch.xml, check the URL, etc...). Related to my out of topic comment https://github.com/asciimoo/searx/issues/2045#issuecomment-654043502
  * When is it updated ? N/A
  * Is it a problem to not update ? ❌ outdated URL, disappointing user experience.
  * Is it a problem to update ? No.
* ```searx/data/currencies.json``` : 
  * Update script: https://github.com/asciimoo/searx/blob/master/utils/fetch_currencies.py
  * Data source: https://github.com/asciimoo/searx/pull/993/files
  * When it should be updated: ?? check every month / year.
  * When it is updated: never
  * Is it a problem to not update ? ❔ should not be a problem.
  * Is it a problem to update ? No.
* ```searx/data/useragents.json```: 
  * Update script: https://github.com/asciimoo/searx/blob/master/utils/fetch_firefox_version.py
  * Data source: is https://ftp.mozilla.org/pub/firefox/releases/
  * When it should be updated: as soon there is a new Firefox version, but engines compatibility must be check before.
  * When it is updated: sometimes.
  * Is it a problem to not update ? ❔ An old Firefox version may be a problem with some engines.
  * Is it a problem to update ? Some engine may stop working.
* ```searx/data/engines_languages.json```: 
  * Update script: https://github.com/asciimoo/searx/blob/master/utils/fetch_languages.py
  * Data source: the source is the results of the ```fetch_supported_languages``` / ```_fetch_supported_languages``` functions.
  * When it should be updated: ?? check every week / month / year ??
[...]

PR #2275, opened 2020-10-26
[mod] ahmia_filter.py: minor changes

dalf:
## What does this PR do?

minor changes in ahmia_filter.py:
- use result['parsed_url']
- load ahmia_blacklist.txt in searx.data

- minor speed improvement
- load data in one place (related to https://github.com/searx/searx/issues/2052 )

- check that ahmia_filter works as before

### DIFF
--- a/searx/plugins/ahmia_filter.py
+++ b/searx/plugins/ahmia_filter.py
@@ -3,9 +3,7 @@
 '''
 
 from hashlib import md5
-from os.path import join
-from urllib.parse import urlparse
-from searx import searx_dir
+from searx.data import ahmia_blacklist_loader
 
 name = "Ahmia blacklist"
 description = "Filter out onion results that appear in Ahmia's blacklist. (See https://ahmia.fi/blacklist)"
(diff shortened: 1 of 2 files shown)


searx/plugins/ahmia_filter.py
```diff
@@ -18,15 +16,14 @@
...
-            ahmia_blacklist = f.read().split()
+        ahmia_blacklist = ahmia_blacklist_loader()
     return ahmia_blacklist
 
 
 def not_blacklisted(result):
     if not result.get('is_onion'):
         return True
-    result_hash = md5(urlparse(result.get('url')).hostname.encode()).hexdigest()
+    result_hash = md5(result['parsed_url'].hostname.encode()).hexdigest()
```
asciimoo, line 26: Are you sure that `parsed_url` exists every time, even in case of offline engines?

dalf: You are right, there is no parsed_url / url for offline engines according to :
https://github.com/searx/searx/blob/db703a0283ee169381aeea97c678e666ae508348/searx/results.py#L202-L202

What do you think of this : replace the line 27 by
```
if not result.get('is_onion') or not result.get('parsed_url'):
```
?

dalf: Code updated

asciimoo (approved):
Great, thanks!
(Travis is so slow indeed, the job has started more than an hour ago and still running)

PR #2276, opened 2020-10-27
Doc installation fix

pierrechtux:
## What does this PR do?

Fixes a few details in the ['Step by step installation'](https://searx.github.io/searx/admin/installation-searx.html#installation-basic) documentation page.

## Why is this change important?

There are a few details that can make the install go wrong:

 * before doing `sudo -H cp "$SEARX_SRC/searx/settings.yml" "${SEARX_SETTINGS_PATH}"` , the destination path should have been created.  Added a line with `sudo -H mkdir -p "${SEARX_SETTINGS_PATH}"`.
 * In searx.sh , uncommented the definition of  `SEARX_INTERNAL_URL` environment variable in .config.sh , hoping that this will make the documentation more clear by pointing to a correct URL.

This fix should hopefully ease a bit the install process for newcomers.

PR #2273, opened 2020-10-25
Doc installation corrections

pierrechtux:
## What does this PR do?

Just a few fixes in the 'Step by step installation' documentation page.

## Why is this change important?
There are a few details that can make the install go wrong:
<ul>
<li>before doing `sudo -H cp "$SEARX_SRC/searx/settings.yml" "${SEARX_SETTINGS_PATH}"` , the destination path should have been created.  Added a line with `sudo -H mkdir -p "${SEARX_SETTINGS_PATH}"`.
<li>In searx.sh , uncommented the SEARX_INTERNAL_URL environment variable in .config.sh , hoping that this will make the documentation more clear by pointing to a correct URL.
</ul>

PR #2277, opened 2020-10-27
Doc installation corrections

return42:
## What does this PR do?

- add missing quotation marks in docs/build-templates
- searx configuration - add 'mkdir /etc/searx' before copy
- searx configuration - quote bash calls in sed replacements

This PR was inspired by #2273 & #2276 / thanks @pierrechtux

    make docs

### DIFF
--- a/docs/build-templates/filtron.rst
+++ b/docs/build-templates/filtron.rst
@@ -7,11 +7,11 @@
     .. code-block:: sh
 
       $ sudo -H useradd --shell /bin/bash --system \\
-          --home-dir "$SERVICE_HOME" \\
-          --comment "Privacy-respecting metasearch engine" $SERVICE_USER
+          --home-dir \"$SERVICE_HOME\" \\
+          --comment \"Privacy-respecting metasearch engine\" $SERVICE_USER
 
-      $ sudo -H mkdir "$SERVICE_HOME"
-      $ sudo -H chown -R "$SERVICE_GROUP:$SERVICE_GROUP" "$SERVICE_HOME"
+      $ sudo -H mkdir \"$SERVICE_HOME\"
+      $ sudo -H chown -R \"$SERVICE_GROUP:$SERVICE_GROUP\" \"$SERVICE_HOME\"
 
 .. END create user
 
@@ -23,16 +23,16 @@
 
     .. code-block:: bash
 
-       $ cat > "$GO_ENV" <<EOF
+       $ cat > \"$GO_ENV\" <<EOF
        export GOPATH=${SERVICE_HOME}/go-apps
        export PATH=\$PATH:${SERVICE_HOME}/local/go/bin:\$GOPATH/bin
        EOF
-       $ sudo -i -u "${SERVICE_USER}"
+       $ sudo -i -u \"${SERVICE_USER}\"
        (${SERVICE_USER}) $ echo 'source $GO_ENV' >> ~/.profile
        (${SERVICE_USER}) $ mkdir ${SERVICE_HOME}/local
-       (${SERVICE_USER}) $ wget --progress=bar -O "${GO_TAR}" \\
-                   "${GO_PKG_URL}"
-       (${SERVICE_USER}) $ tar -C ${SERVICE_HOME}/local -xzf "${GO_TAR}"
+       (${SERVICE_USER}) $ wget --progress=bar -O \"${GO_TAR}\" \\
+                   \"${GO_PKG_URL}\"
+       (${SERVICE_USER}) $ tar -C ${SERVICE_HOME}/local -xzf \"${GO_TAR}\"
        (${SERVICE_USER}) $ which go
        ${SERVICE_HOME}/local/go/bin/go
 
@@ -46,7 +46,7 @@
 
     .. code-block:: bash
 
-       $ sudo -i -u "${SERVICE_USER}"
+       $ sudo -i -u \"${SERVICE_USER}\"
        (${SERVICE_USER}) $ go get -v -u github.com/asciimoo/filtron
 
 .. END install filtron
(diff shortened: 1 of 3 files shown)


pierrechtux (commented):
For some reason, when I run the
`make docs`
, the generated page
`searx/dist/docs/admin/installation-searx.html`
doesn't display correctly all the frames labeled 'bash'.

Therefore I cannot test the generated documentation.

I'm going to try to find out why that webpage is not correctly rendered, maybe something is wrong on my side.

return42:
> doesn't display correctly all the frames labeled 'bash'.

Except the frame in chapter "Install packages" nearly all frames on this html-page do only have one tab and it is named "bash".

pierrechtux:
> > doesn't display correctly all the frames labeled 'bash'.
> 
> Except the frame in chapter "Install packages" nearly all frames on this html-page do only have one tab and it is named "bash".

Yes, fine.  But in the html generated file on my machine, these frames do not appear at all.

Here is the output with the error messages when I try to regenerate the doc by making `make docs-clean docs`:

[searx_try_regen_documentation_2020_10_28__18_10.txt](https://github.com/searx/searx/files/5453688/searx_try_regen_documentation_2020_10_28__18_10.txt)

There's an error at the beginning concerning "missing command systemctl": this is normal, my machine runs on Devuan, which is basically a Debian without systemd.  This doesn't worry me much.

The two lines with "WARNING: Unexpected return code 42 from command " concerning "../utils/lxc.sh --help" and "../utils/searx.sh --help" worry me a bit more.

Sorry for the French in the output messages.

I'm trying to figure out how that system of documentation generation is working.

return42:
> Sorry for the French in the output messages.

You can use `LANG=C` in front of your command to get `en` localized messages.

> Here is the output with the error messages when I try to regenerate the doc by making make docs-clean docs:

There is no need for `. ./searx-ve/bin/activate` when using Makefile targets, since they already build their own pyenv, see https://searx.github.io/searx/dev/makefile.html#python-environment

> my machine runs on Devuan, which is basically a Debian 

`make docs` should work

> a Debian without systemd

Our [installation scripts](https://searx.github.io/searx/admin/installation.html#installation-scripts) won't work.  I am not a fan boy of systemd but since all major distributors decided to pack systemd as default init system, all our installation scripts are using systemd.

pierrechtux:
Here's the output, with `LANG=C`, of `make docs`.
[searx_try_regen_documentation_2020_10_30__15_12.txt](https://github.com/searx/searx/files/5465664/searx_try_regen_documentation_2020_10_30__15_12.txt)

Actually, the problem of those tabbed areas containing the code not appearing in the html files on my side is certainly related to my own setup, and doesn't have anything to do with your PR, so for the moment, I'll just leave it as it is.

(I initially thought that this would be a super-easy contribution... oh well... ;o))

pierrechtux (approved):
Seems good to me.  Too bad I cannot test, but this is most certainly my fault.

return42:
> Too bad I cannot test, but this is most certainly my fault.

@pierrechtux thanks for taking the time!  _"don't worry, be happy"_ :smile:  ... if you wan't to have a last try I suggest to start from the beginning, very simply by::

````
$ cd ~/Downloads/
$ git clone https://github.com/searx/searx.git
$ cd searx/
$ git pull origin refs/pull/2277/head
$ SEARX_DEBUG=1 make docs-live
````

**Review by pierrechtux — APPROVED** [review]

**Comment by dalf** [comment]
Thank you !

**Comment by pierrechtux** [comment]
> @pierrechtux thanks for taking the time! _"don't worry, be happy"_ smile ... if you wan't to have a last try I suggest to start from the beginning, very simply by::
> 
> ```
> $ cd ~/Downloads/
> $ git clone https://github.com/searx/searx.git
> $ cd searx/
> $ git pull origin refs/pull/2277/head
> $ SEARX_DEBUG=1 make docs-live
> ```

I followed your advice, letter for letter (except I went to my favourite ~/tmp/ directory instead of Downloads), and still, I get the same results: no tabbed frames in the resulting webpage.

(not worth pasting here the long output from the make process, unless you wish to get it)

**Comment by return42** [comment]
> I followed your advice, letter for letter ... and still, I get the same results: no tabbed frames

That's really a pity / I do not have experience with Devuan so I can't say what's going wrong. I even do not know if it is related to Devuan or  what the cause is. Sorry that I can't help you solving the issue you have.

**Comment by pierrechtux** [comment]
No worries. Actually, Devuan is extremely similar to Debian, and my current setup is an actual Debian, which I switched to Devuan (that took just a few seconds).  I did that when the systemd made my machine boot in 30 minutes, instead of under one minute.

But I have another Debian machine closeby (which supports systemd), so I'll wake up that machine and I'll work on it, to check if this is related to Devuan or not.

**Comment by pierrechtux** [comment]
I've just tried that on a machine with Debian 10; results are even worse.

Some requirements are obviously missing: latex, something with pygments.

Here's a (long) record of my terminals when doing that:
[searx_try_regen_documentation_2020_11_05_20_09.txt](https://github.com/searx/searx/files/5496469/searx_try_regen_documentation_2020_11_05_20_09.txt)

Note that the text is structured with {{{ and }}} which are default delimiters for folding in vim.

In vim, setting:
```
set foldmethod=marker
set fdc=5
set foldclose=all
```
and playing with z commands helps a lot.

return42:
> and playing with z commands helps a lot.

Use the terminal.

> get pull request from return42

This PR is already merged, no longer need to pull from my (old) branch.

> WARNING: Unexpected return code 42 from command '../utils/lxc.sh --help'

Its very hard to read this log from vim.  lxc is not needed to build the docs.

> ! LaTeX Error: File `anyfontsize.sty' not found.

Seems you have latex installed but something is missed .. if latex is installed, it is used to render math expressions.

> OSError: [Errno 98] Address already in use

The port docs-live uses is already used by another process.

Over all: the LOG looks very strange to me, never seen before.

Sorry, I can't help.

PR #2285, opened 2020-10-30
bugfix & refactor digg engine 

return42:
## What does this PR do?

_from the commit messages .._ 

1. The ISO time stamp of published date has been changed

   Error pattern:

       Engines cannot retrieve results:
       digg (unexpected crash time data '2020-10-16T14:09:55Z' does not match format '%Y-%m-%d %H:%M:%S')

2. digg - improve results and clean up source code

- strip html tags and superfluous quotation marks from content
- remove not needed cookie from request
- remove superfluous imports
- add searx/engines/digg.py to pylint check (test.pylint)

## Why is this change important?

fix errors and remove obsolete source code

## How to test this PR locally?

Use search query `!digg wheels`

PR #2286, opened 2020-10-30
[refactor] deviantart - improve results and clean up source code

return42:
## What does this PR do?

Devian's request and response forms has been changed.

- fixed title
- fixed time_range_dict to 'popular-*-***'
- use image from <noscript> if exists
- drop obsolete "http to https, remove domain sharding"
- use query URL https://www.deviantart.com/search/deviations?page=5&q=foo
- add searx/engines/deviantart.py to pylint check (test.pylint)

Error pattern::

    There DEBUG:searx:result: invalid title: {'url': 'https://www.deviantart.com/  ...

Old one throws out no results but errors 

Query:  `!da  millennium falcon`

### DIFF
--- a/searx/engines/deviantart.py
+++ b/searx/engines/deviantart.py
@@ -7,75 +7,70 @@
  @using-api   no (TODO, rewrite to api)
  @results     HTML
  @stable      no (HTML can change)
- @parse       url, title, thumbnail_src, img_src
+ @parse       url, title, img_src
 
  @todo        rewrite to api
 """
+# pylint: disable=missing-function-docstring
 
-from lxml import html
-import re
 from urllib.parse import urlencode
-
+from lxml import html
 
 # engine dependent config
 categories = ['images']
 paging = True
 time_range_support = True
 
+time_range_dict = {
+    'day': 'popular-24-hours',
+    'week': 'popular-1-week',
+    'month': 'popular-1-month',
+    'year': 'most-recent',
+}
+
 # search-url
-base_url = 'https://www.deviantart.com/'
-search_url = base_url + 'search?page={page}&{query}'
-time_range_url = '&order={range}'
+base_url = 'https://www.deviantart.com'
 
-time_range_dict = {'day': 11,
-                   'week': 14,
-                   'month': 15}
-
-
-# do search-request
 def request(query, params):
-    if params['time_range'] and params['time_range'] not in time_range_dict:
-        return params
 
-    params['url'] = search_url.format(page=params['pageno'],
-                                      query=urlencode({'q': query}))
+    # https://www.deviantart.com/search/deviations?page=5&q=foo
+
+    query =  {
+        'page' : params['pageno'],
+        'q'    : query,
+    }
     if params['time_range'] in time_range_dict:
-        params['url'] += time_range_url.format(range=time_range_dict[params['time_range']])
+        query['order'] = time_range_dict[params['time_range']]
+
+    params['url'] = base_url + '/search/deviations?' + urlencode(query)
 
     return params
 
-
-# get response from search-request
 def response(resp):
-    results = []
 
-    # return empty array if a redirection code is returned
-    if resp.status_code == 302:
-        return []
+    results = []
 
     dom = html.fromstring(resp.text)
 
-    # parse results
     for row in dom.xpath('//div[contains(@data-hook, "content_row")]'):
         for result in row.xpath('./div'):
-            link = result.xpath('.//a[@data-hook="deviation_link"]')[0]
-            url = link.attrib.get('href')
-            title = link.attrib.get('title')
-            thumbnail_src = result.xpath('.//img')[0].attrib.get('src')
-            img_src = thumbnail_src
 
-            # http to https, remove domain sharding
-            thumbnail_src = re.sub(r"https?://(th|fc)\d+.", "https://th01.", thumbnail_src)
-            thumbnail_src = re.sub(r"http://", "https://", thumbnail_src)
+            a_tag = result.xpath('.//a[@data-hook="deviation_link"]')[0]
+            noscript_tag = a_tag.xpath('.//noscript')
 
-            url = re.sub(r"http://(.*)\.deviantart\.com/", "https://\\1.deviantart.com/", url)
+            if noscript_tag:
+                img_tag = noscript_tag[0].xpath('.//img')
+            else:
+                img_tag = a_tag.xpath('.//img')
+            if not img_tag:
+                continue
+            img_tag = img_tag[0]
 
-            # append result
-            results.append({'url': url,
-                            'title': title,
-                            'img_src': img_src,
-                            'thumbnail_src': thumbnail_src,
-                            'template': 'images.html'})
+            results.append({
+                'template': 'images.html',
+                'url': a_tag.attrib.get('href'),
+                'img_src': img_tag.attrib.get('src'),
+                'title': img_tag.attrib.get('alt'),
+            })
 
-    # return results
     return results
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,7 @@ test.pylint: pyenvinstall
 		searx/preferences.py \
 		searx/testing.py \
 		searx/engines/gigablast.py \
+		searx/engines/deviantart.py \
 	)
 
 # ignored rules:
@@ -236,7 +237,7 @@ test.sh:
 
 test.pep8: pyenvinstall
 	@echo "TEST      pycodestyle (formerly pep8)"
-	$(Q)$(PY_ENV_ACT); pycodestyle --exclude='searx/static, searx/languages.py, searx/engines/gigablast.py' \
+	$(Q)$(PY_ENV_ACT); pycodestyle --exclude='searx/static, searx/languages.py, searx/engines/gigablast.py, searx/engines/deviantart.py' \
         --max-line-length=120 --ignore "E117,E252,E402,E722,E741,W503,W504,W605" searx tests
 
 test.unit: pyenvinstall


searx/engines/deviantart.py
```diff
@@ -7,76 +7,76 @@
...
-# get response from search-request
 def response(resp):
+    """get response from search-request"""
+
     results = []
 
-    # return empty array if a redirection code is returned
-    if resp.status_code == 302:
-        return []
+    with open("/tmp/dump.html", "w") as f:
```
kvch: Leftover from debugging?

searx/engines/deviantart.py
```diff
@@ -7,76 +7,76 @@
...
+            else:
+                img_tag = a_tag.xpath('.//img')
+            if not img_tag:
+                continue
+            img_tag = img_tag[0]
+
+            item = dict()
+
+            item['template'] = 'images.html'
+            item['url']      = a_tag.attrib.get('href')
```
kvch: Why do these lines pass on pep8/pylint?

Why don't you follow the convention in python and searx where we set dict values like this:

```python
results.append({'url': url,
                'title': title,
                'img_src': img_src,
                'thumbnail_src': thumbnail_src,
                'template': 'images.html'})
```

searx/engines/deviantart.py
```diff
@@ -7,76 +7,76 @@
...
 def request(query, params):
-    if params['time_range'] and params['time_range'] not in time_range_dict:
-        return params
+    """do search-request"""
 
-    params['url'] = search_url.format(page=params['pageno'],
-                                      query=urlencode({'q': query}))
+    # https://www.deviantart.com/search/deviations?page=5&q=foo
+
+    query =  dict(
```
kvch: This indentation also seems weird. Why aren't you using the conventional indentation?

searx/engines/deviantart.py
```diff
@@ -7,76 +7,76 @@
...
+)
 
+# search-url
+base_url = 'https://www.deviantart.com'
 
-# do search-request
 def request(query, params):
-    if params['time_range'] and params['time_range'] not in time_range_dict:
-        return params
+    """do search-request"""
```
kvch: This comment seems unnecessary.

searx/engines/deviantart.py
```diff
@@ -7,76 +7,76 @@
...
-base_url = 'https://www.deviantart.com/'
-search_url = base_url + 'search?page={page}&{query}'
-time_range_url = '&order={range}'
-
-time_range_dict = {'day': 11,
-                   'week': 14,
-                   'month': 15}
+time_range_dict = dict(
+    day     = 'popular-24-hours'
+    , week  = 'popular-1-week'
```
kvch: Why are you putting the commas to the beginning of the line? We do not use this kind of format in searx.

kvch (changes requested):
The fix look good, however, the indentation is comletely different from the convetions we use in searx. I am not sure why our style checker accepts this format. But we definitely have to enforce the style used in searx.

return42:
@kvch thanks for having an eye on.

>  the indentation is comletely different from the convetions we use in searx.

I rebased and force pushed a patch.  

> Why aren't you using the conventional indentation?

[For new code pep8 suggests Donald E. Knuth's style](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

kvch:
> > Why aren't you using the conventional indentation?
> 
> [For new code pep8 suggests Donald E. Knuth's style](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

I agree with those guidelines and use it where possible. However, it only applies to binary operators. Comma is not a binary operator. Ref: https://www.cs.auckland.ac.nz/references/unix/digital/AQTLTBTE/DOCU_062.HTM

searx/engines/deviantart.py
```diff
@@ -7,76 +7,70 @@
...
+
+            if noscript_tag:
+                img_tag = noscript_tag[0].xpath('.//img')
+            else:
+                img_tag = a_tag.xpath('.//img')
+            if not img_tag:
+                continue
+            img_tag = img_tag[0]
+
+            results.append({
```
kvch, line 69: Why have you changed the indentation here?

return42:
>  I agree with those guidelines ... Comma is not a binary operator. 

.. this is not correct. The definition of a binary operation is very simple: A binary operation is a operation that combines two elements (called operands) to produce another element.

    a = b + c
    a = [b, c] 

In python and many other languages, a comma is the operator to concatenate two (or more) elements into a list (or tuple).

But this is only a academical view, the question is; what is common used (in searx) ..

> the indentation is comletely different from the convetions we use in searx

so I fixed to meet searx needs.

searx/engines/deviantart.py
return42: This is a `[refactor]` patch. So I decided to switch from the [_"Aligned with opening delimiter"_](https://www.python.org/dev/peps/pep-0008/#indentation) style to the [ _"Add 4 spaces (an extra level of indentation) to distinguish arguments from the rest"_](https://www.python.org/dev/peps/pep-0008/#indentation) style.

With an eye on the [Maximum Line Length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) I've got used to add an extra level of indentation (the latter style).

IMO: The growing diff is a common drawback of the patch, when switching from one code style to another. But I think this additional diff is acceptable in a *rewrite/refactoring* patch.

searx/engines/deviantart.py
dalf: pylint in the current configuration requires a docstring for each function, module and class.

`disable=bad-whitespace, duplicate-code, missing-function-docstring, missing-class-docstring, missing-module-docstring`
in `.pylintrc` remove this need.

return42: > in .pylintrc remove this need.

I vote to disable only where it is needed to be disabled. E.g. I added

````python
# pylint: disable=missing-function-docstring
````

on top of the py-file.

**Comment by kvch** [comment]
> > I agree with those guidelines ... Comma is not a binary operator.
> 
> .. this is not correct. The definition of a binary operation is very simple: A binary operation is a operation that combines two elements (called operands) to produce another element.
> 
> ```
> a = b + c
> a = [b, c] 
> ```
> 
> In python and many other languages, a comma is the operator to concatenate two (or more) elements into a list (or tuple).
> 
> But this is only a academical view, the question is; what is common used (in searx) ..
> 

The academic view is not this. Binary means something has two parts. Binary operation is the operation that combines *exactly two* things:
__Let  S  be a non-empty set, and  ⋆  said to be a binary operation on  S  , if  a⋆b  is defined for all  a,b∈S . In other words,  ⋆  is a rule for any two elements in the set  S.__

Binary operators e.g. `+` can be rewritten like this `+(a, b)`. But in programming we use the format `a + b`. In your example `[a, b]` comma is not the operator. The operator is `list` and can be called like `list(a, b)`. But as Python comes with lots of syntactical sugar including `[` and `]` to represent lists, we do not have to write out the operator. But still it is just an "alias" for `list`.

File: searx/engines/deviantart.py (hunk 6 above)
Comment (kvch) [reply searx/engines/deviantart.py:69]: I am sorry if I was not clear. Personally I prefer this indentation, as if you add a new element to the end of the line, you do not have to add extra comma, thus decreasing the size of the diff. By changing the indentation I meant why did you put `:` at the same position in every line? 

If you add a new key which is longer than any of the existing keys, the whole dictionary has to be realigned, so the `:`s can be at the same vertical position again.

But I would be happy with this format:

```python
results.append({
    'template': 'images.html',
    'url': a_tag.attrib.get('href'),
    'img_src': img_tag.attrib.get('src'),
    'title': img_tag.attrib.get('alt'),
})
```

return42: > I meant why did you put : at the same position in every line?

It is a (uneasy) compromise I often use for readability reasons.   Its a weak example, but in long lines and more key/value pairs, it has its value:

````python
{
    'img_src': img_tag.attrib.get('src'),
    'title': img_tag.attrib.get('alt'),
}
````

IMO: aligned its more _eye catching_ that merely different attributes from the same element are used as values

````python
{
    'img_src': img_tag.attrib.get('src'),
    'title':   img_tag.attrib.get('alt'),
}
````

The big drawback of is ..

> If you add a new key which is longer 

Yes, hoping there is enough space left does not scale very good in the diff.

I rebased my branch, removed the alignment (also in the `time_range_dict`)  and force-pushed.

return42:
**update**: [commas are not operators](https://github.com/python/peps/issues/1706)

@kvch thanks for correcting me!

---

TL:DR;

> The academic view is not this. 
> But as Python comes with lots of syntactical sugar including 

In C exists [the comma operator (binary operator)](https://en.wikipedia.org/wiki/Comma_operator) but I think this one is off topic here. I refer this one here, just to distinguish "binary operator" in between a mathematical context and a programming context.

> Binary means something has two parts.  ....

I have to admit that I've never thought about it much.  I see the point in your explanation (thanks) and now I have a doubt how to read _"binary operators"_ in this sentence:

[For decades the recommended style was to break after binary operators. But this can hurt readability in two ways: the operators tend to get scattered across different columns on the screen, and each operator is moved away from its operand and onto the previous line. ](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

Since this section is about _readability_  I didn't distinguished between a binary operator in mathematics and a comma in a programming context (e.g. to separating the items of an argument list).  May be my fail .. 

With your argumentation in mind, I don't have any longer a strong opinion.

**update**: I added an issue to the python/peps repo ["PEP-8: Should a line break before or after a comma?"](https://github.com/python/peps/issues/1706)

kvch:
> **update**: [commas are not operators](https://github.com/python/peps/issues/1706)
> 
> @kvch thanks for correcting me!
> 
> TL:DR;
> 
> > The academic view is not this.
> > But as Python comes with lots of syntactical sugar including
> 
> In C exists [the comma operator (binary operator)](https://en.wikipedia.org/wiki/Comma_operator) but I think this one is off topic here. I refer this one here, just to distinguish "binary operator" in between a mathematical context and a programming context.

I was not aware of this operator in C. I had coded in C for a few years, but for some reason never encountered it. I learnt something new today. Thanks for that. :)

kvch (dismissed):

return42:
> I learnt something new today. Thanks for that. :)

Same on my side. Thanks for the discussion that helps me to correct a long standing mistake!

I rebased and force pushed .. if you have time to merge .. afterwards I will fix [2285](https://github.com/searx/searx/pull/2285#issuecomment-720967342) according to your remarks from this PR.

return42:
Sorry for being impatient / this engine is already approved by @kvch, can we please merge so I can continue with the next engine (#2285) / thanks!

dalf:
Thank you !

PR #2289, opened 2020-11-02
[mod] pylint: add extension-pkg-whitelist=lxml.etree

dalf:
## What does this PR do?

lxml.etree uses C extensions, as described by the pylint documention, this PR add the module name lxml.etree to extension-pkg-whitelist

There is a second commit: https://github.com/searx/searx/pull/2289#issuecomment-721046822

## Why is this change important?

Without this change, pylint displays error about lxml.

Important note: these errors are not displayed currently since pylint run only a few files.

## How to test this PR locally?

Run:
```sh
#!/bin/bash
export PYLINT_RC=.pylintrc
pylint --rcfile ${PYLINT_RC} searx/preferences.py searx/testing.py searx/engines/gigablast.py
pylint --rcfile ${PYLINT_RC} --disable=I,C,R,W --additional-builtins=supported_languages,language_aliases searx/engines
pylint --rcfile ${PYLINT_RC} --disable=I,C,R,W --ignore=searx/engines searx
````

## Notes

The current output  of ```pylint --rcfile=.pylintrc --disable=I,C,R,W searx``` :

```
************* Module searx.webadapter
searx/webadapter.py:126:15: E0602: Undefined variable 'preferences' (undefined-variable)
************* Module searx.webapp
searx/webapp.py:47:0: E0611: No name 'HtmlFormatter' in module 'pygments.formatters' (no-name-in-module)
searx/webapp.py:114:0: E1101: Method 'jinja_env' has no 'add_extension' member (no-member)
searx/webapp.py:574:46: E1101: Instance of 'Exception' has no 'message' member (no-member)
************* Module searx.engines.yahoo
searx/engines/yahoo.py:79:50: E0602: Undefined variable 'supported_languages' (undefined-variable)
************* Module searx.engines.scanr_structures
searx/engines/scanr_structures.py:14:0: E0611: No name 'html_to_text' in module 'urllib.parse' (no-name-in-module)
************* Module searx.engines.qwant
searx/engines/qwant.py:38:7: E1136: Value 'categories' is unsubscriptable (unsubscriptable-object)
searx/engines/qwant.py:38:25: E1136: Value 'categories' is unsubscriptable (unsubscriptable-object)
searx/engines/qwant.py:40:63: E1136: Value 'categories' is unsubscriptable (unsubscriptable-object)
searx/engines/qwant.py:50:54: E0602: Undefined variable 'supported_languages' (undefined-variable)
[...]

PR #2290, opened 2020-11-03
Various bug fixes

dalf:
## What does this PR do?

This PR fixes:
- the command engine
- the scanr_structures engine. Note: the certificate is valid valid, but the API has changed : .../api/v2/structures/search )
- the locked categories (in settings.yml). Note: when locked, the categories is set to general without choice :
https://github.com/searx/searx/blob/45f58a4a2a0b89f4b416c28ea769139b16f6436d/searx/preferences.py#L330

Bug fixes.

- command engine: check the example in settings.yml works as expected.
- scanr_structures: impossible to test since the API has changed (related to #2289 )
- locked categories: add in settings.yml
```yml
preferences:
    lock:
      - categories
```
and check the categories are locked (using a bang, a URL parameter, using the UI).

### DIFF
--- a/searx/engines/command.py
+++ b/searx/engines/command.py
@@ -14,8 +14,8 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 '''
 
 
+import re
 from os.path import expanduser, isabs, realpath, commonprefix
-from re import MULTILINE, search as re_search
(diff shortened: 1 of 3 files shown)


asciimoo (approved):
LGTM, thanks!

PR #2291, opened 2020-11-03
[enh] user settings can relied on the default settings

dalf:
## What does this PR do?

This PR is backward compatible with the existing configurations.

It add a new option in settings.yml:
```
use_default_settings: true
```

This new option works only when settings.yml path is loaded from `/etc/searx/settings.yml` or the path defined by the SEARX_SETTINGS_PATH environment variable.

When `True`, the settings.yml is merged with default one (define in the git repository).

For example:

```yaml
use_default_settings: true
server:
    secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
server:
    bind_address: "0.0.0.0"
```

The actual settings are the default settings defined in the git repository with the exception of the secret_key and the bind_address.

There is one exception: if the engine section is defined, only the declare engine while be available.

For example:
```yaml
use_default_settings: true
server:
    secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
engines:
  - name: arch linux wiki
    tokens: ['$ecretValue']
  - name: wikipedia
  - name: wikidata
  - name: ddg definitions
```

In the above example, searx will load only the four declared engines. For each engine, the configuration is merged with the engine configuration defined in settings.yml.

There is two new make target. One of them:
```sh
make SEARX_SETTINGS_PATH=/etc/searx/settings.yml user-settings.update.engines
```
Here /etc/searx/settings.yml is updated:
* a secret_key is defined if not set
* unknown engine (according to [settings.yml](https://github.com/searx/searx/blob/master/searx/settings.yml) ) are removed the engine list
* new engine (according to [settings.yml](https://github.com/searx/searx/blob/master/searx/settings.yml) ) are added the engine list.

Note: to update the yaml file, the python script uses [`ruamel.yaml`](https://pypi.org/project/ruamel.yaml/) instead of pyyaml to keep the comments (for reference see https://github.com/yaml/pyyaml/issues/90 and https://web.archive.org/web/20160812090911/https://pyyaml.org/ticket/114 ). `ruamel.yaml` is installed when the user calls the `user-settings.update.engines` target.

When searx is not installed from a git repository, it can be not easy to keep an up to date settings.yml. See https://github.com/searx/searx/issues/1209

## Related issues

* Related to https://github.com/searx/searx/issues/1209
* Alternative / superset https://github.com/searx/searx/pull/1278


asciimoo:
Why is this a draft PR? Looks good to me and it's super useful. Let me know when I can start the review.

dalf:
It is ready for ready, but it is weird if the PR is still shown as a draft : I remembered to have removed the draft status and I don't see it anymore.

[EDIT] side note: the missing part is the docker image (it would be nice to upgrade the settings.yml automatically if there is `use_default_settings: true` in the user settings). It would be the subject of another PR.

return42:
FWIW: for me, the DRAFT tag is removed (press reload button in the browser?)

dalf:
Yes sometimes I have to reload a page from Github: when I press the back button (and some other scenarios) I see obviously an old content, then if I reload the page the right content is shown ( ... and sometimes is not obvious and the old content puzzles me).

Makefile
```diff
@@ -265,4 +265,19 @@ test.clean:
...
+# -------------
+
+PHONY += user-settings.create user-settings.update
+
+user-settings.update:  pyenvinstall
+	$(Q)$(PY_ENV_ACT); pip install ruamel.yaml
+	$(Q)$(PY_ENV_ACT); python utils/update_user_settings.py ${SEARX_SETTINGS_PATH}
+
+user-settings.update.engines:  pyenvinstall
+	$(Q)$(PY_ENV_ACT); pip install ruamel.yaml
```
return42: Does it make sense to add `ruamel.yaml` to `requirements-dev.txt`?

docs/admin/settings.rst
```diff
@@ -188,3 +188,97 @@ Engine settings
...
+The following comand creates or updates a minimal user settings (a secret key is defined if it is not already the case):
+
+.. code-block:: sh
+
+  make SEARX_SETTINGS_PATH=/etc/searx/settings.yml user-settings.update
+
+Set ``SEARX_SETTINGS_PATH`` to your user settings path.
+
+As soon the user settings contains an ``engines`` section, it becomes difficult to keep the engine list updated.
+The following command creates or updates the user settings including the ``engines`` section:
```
return42: Do we really need a update tool for the engines? I guess it is hard to patch the engines in the user settings without having a `diff` of the origin engine settings.yml patch (I haven't look right now into the implementation, theses are just my first thoughts).

Makefile
dalf: The purpose is not to develop / test code, but to update the configuration.

`requirements.txt` fits better, so it would make sense to:
* remove the `pyyaml` dependency
* rely only on `ruamel.yaml`

But `ruamel.yaml` is required in this specific use case and nowhere else, so I went for this hybrid approach.

docs/admin/settings.rst
dalf: If the administrator want to specify one token for one engine, all the engines have to be written down from this:
```yaml
use_default_settings: true
server:
    secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
engines:
  - name: bing
    tokens: ['$ecretValue']
```

...painful.


An alternative approach is to merge the above configuration with the default settings, but in this case it becomes impossible to remove an engine.

Also, may be this tool can be helpful a searx packaged (Debian, Arch, Docker, etc...) is updated : the configuration can be updated automatically with the new engines. Currently the same thing can be done with the full settings.yml but it doesn't seem to be the case of the Debian package for example (see https://github.com/searx/searx/issues/1209 )

dalf: Another idea:
```yaml
use_default_settings: true
server:
    secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
engines:
  - name: bing
    tokens: ['$ecretValue']
remove_engines:  # valid only with use_default_settings: true
  - bing news
  - btdigg
```

In this example, all the engines in the default settings.yml are included except `bing news` and `btdigg`. The `bing` engine has a token.

In this case there is no need to list all the engines, so there is no need to have tool to update the engine list.

return42: > In this case there is no need to list all the engines, so there is no need to have tool to update the engine list.

I would really prefer such a solution, making maintenance less complicated and from searx-admin's view it is easy to understand.  Thanks!

dalf:
I have updated the PR with a new commit (the old one is kept for reference).

`searx/settings.py` -> `searx/settings_loader.py` otherwise `import searx.settings` imports the settings and not the module so it is not possible to test the module.

New syntax which doesn't require a tool to update the user settings. The code is more complicated, `test_settings_loader.py` tests the different use cases (different test files have been added in `tests/unit/settings/` )

-----

```yaml
  use_default_settings: True
  server:
      secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
  engines:
    - name: arch linux wiki
      tokens: ['$ecretValue']
```
In this example, searx loads _all_ the engines, and adds a tokens to the `arch linux wiki`

----

```yaml
  use_default_settings:
      engines:
         remove:
           - google
  server:
      secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
  engines:
    - name: arch linux wiki
      tokens: ['$ecretValue']
```
In this example, searx loads all the engine except `google`.  As the previous example, there is token for the `arch linux wiki` engine.

----

```yaml
  use_default_settings:
      engines:
         keep_only:
           - google
           - duckduckgo
  server:
      secret_key: "uvys6bRhKHUdFF5CqbJonSDSRN8H0sCBziNSrDGNVdpz7IeZhveVart3yvghoKHA"
  engines:
    - name: google
      tokens: ['$ecretValue']
    - name: duckduckgo
      tokens: ['$ecretValue']
```
In this example, searx loads only two engines: google and duckduckgo.

PR #2292, opened 2020-11-03
New engine: Elasticsearch

kvch:
## What does this PR do?

This PR adds a new engine to let users search in Elasticsearch instances. It is able to execute multiple types of queries. Also, it lets instance administrators to provide their custom query.

### Configuration

`base_url`: This is the URL of the Elasticsearch instance. Default: `http://localhost:9200`
`index`: Name of the index or indices. It also accepts patterns. E.g. `my-index-*`. Required.
`username`: Name of the user. Not required.
`password`: Password of the user. Not required.
`query_type`: Type of the query to execute. Available: `match`, `simple_query_string`, `term`, `terms`, `custom`
`custom_query_json`: If `query_type` is set to `custom`, you can set the custom query here. `{{KEY}}` and `{{VALUE}}` are reserved for the key and value part of from the query submitted by the user.

###  Search syntax

Some queries require key value pairs. To configure those you have to use the following syntax: `key:value`.

Example to get the document where user is kvch:
```
!es user:kvch
```

This is a step toward letting admins to add local search engines to their searx instances.

### DIFF
--- a/searx/engines/elasticsearch.py
+++ b/searx/engines/elasticsearch.py
@@ -1,0 +1,142 @@
+from json import loads, dumps
+from lxml import html
+from urllib.parse import quote, urljoin
+from requests.auth import HTTPBasicAuth
+from searx.utils import extract_text, get_torrent_size
+
+
+base_url = 'http://localhost:9200'
+username = ''
+password = ''
+index = ''
+search_url = base_url + '/' + index + '/_search'
+query_type = 'match'
+custom_query_json = {}
+show_metadata = False
+categories = ['general']
+
+
+def init(engine_settings):
+    if 'query_type' in engine_settings and engine_settings['query_type'] not in _available_query_types:
+        raise ValueError('unsupported query type', engine_settings['query_type'])
+
+    if index == '':
+        raise ValueError('index cannot be empty')
+
+
+def request(query, params):
+    if query_type not in _available_query_types:
+        return params
+
+    if username and password:
+        params['auth'] = HTTPBasicAuth(username, password)
+
+    params['url'] = search_url
+    params['method'] = 'GET'
+    params['data'] = dumps(_available_query_types[query_type](query))
+    params['headers']['Content-Type'] = 'application/json'
+
+    return params
+
+
+def _match_query(query):
+    """
+    The standard for full text queries.
+    searx format: "key:value" e.g. city:berlin
+    REF: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html
+    """
+
+    try:
+        key, value = query.split(':')
+    except:
+        raise ValueError('query format must be "key:value"')
+
+    return {"query": {"match": {key: {'query': value}}}}
+
+
+def _simple_query_string_query(query):
+    """
+    Accepts query strings, but it is less strict than query_string
+    The field used can be specified in index.query.default_field in Elasticsearch.
+    REF: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html
+    """
+
+    return {'query': {'simple_query_string': {'query': query}}}
+
+
+def _term_query(query):
+    """
+    Accepts one term and the name of the field.
+    searx format: "key:value" e.g. city:berlin
+    REF: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html
+    """
+
+    try:
+        key, value = query.split(':')
+    except:
+        raise ValueError('query format must be key:value')
+
+    return {'query': {'term': {key: value}}}
+
+
+def _terms_query(query):
+    """
+    Accepts multiple terms and the name of the field.
+    searx format: "key:value1,value2" e.g. city:berlin,paris
+    REF: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html
+    """
+
+    try:
+        key, values = query.split(':')
+    except:
+        raise ValueError('query format must be key:value1,value2')
+
+    return {'query': {'terms': {key: values.split(',')}}}
+
+
+def _custom_query(query):
+    key, value = query.split(':')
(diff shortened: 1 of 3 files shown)


searx/engines/elasticsearch.py
```diff
@@ -0,0 +1,146 @@
...
+            custom_query[key] = custom_query.pop(query_key)
+        if query_value == '{{VALUE}}':
+            custom_query[query_key] = value
+    return custom_query
+
+
+def response(resp):
+    results = []
+
+    print(resp.text)
```
dalf: A debug leftover

searx/engines/elasticsearch.py
```diff
@@ -0,0 +1,142 @@
+from json import loads, dumps
+from lxml import html
+from urllib.parse import quote, urljoin
+from requests import HTTPBasicAuth
```
dalf: I have import error, fixed with : `from requests.auth import HTTPBasicAuth`

searx/engines/elasticsearch.py
```diff
@@ -0,0 +1,142 @@
...
+    searx format: "key:value" e.g. city:berlin
+    REF: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html
+    """
+
+    try:
+        key, values = query.split(':')
+    except:
+        raise ValueError('query format must be "key:value"')
+
+    return {"query": {"match": {key: {'query': value}}}}
```
dalf, line 54: Should be `values` instead of `value`

searx/engines/elasticsearch.py
```diff
@@ -0,0 +1,142 @@
...
+def response(resp):
+    results = []
+
+    resp_json = loads(resp.text)
+    if 'error' in resp_json:
+        raise Exception(resp_json['error'])
+
+    for result in resp_json['hits']['hits']:
+        r = result['_source']
+        r['template'] = 'key-value.html'
```
dalf, line 117: If some of the field are not string, the template will raise an exception here:
https://github.com/searx/searx/blob/c03e4c86bc49d6ef4664c038066d9f1c16e7dafc/searx/templates/oscar/result_templates/key-value.html#L9

Possible fix, just after line 117, add:
`r = {key: str(value) if not key.startswith('_') else value for key, value in r.items() }`

searx/settings.yml
```diff
@@ -225,6 +225,19 @@ engines:
...
+#    engine : elasticsearch
+#    base_url : http://localhost:9200
+#    username : elastic
+#    password : changeme
+#    index : my-index
+#    # available options: match, simple_query_string, term, terms, custom
+#    query_type : match
+#    # if query_type is set to custom, provide your query here
+#    #custom_query_json: {"query":{"match_all": {}}}
+#    disabled : True
```
dalf, line 246: I would add `show_metadata: True` here.

dalf:
I don't know much about elasticsearch. 

How I have tested:
* download title.basics.tsv.gz from https://datasets.imdbws.com/ (see https://www.imdb.com/interfaces/ for the legal notice)
* start elasticsearch using docker : `docker run -d --name elasticsearch -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" elasticsearch:7.9.3`
* then import imdb data, it crashes after 1 million entries, but at least there are some data (checked using https://elasticvue.com/ ). Some field are converted to bool, or list, some values are None.

<details><summary>import code</summary>

```python
import gzip
import csv
import uuid
from datetime import datetime
from multiprocessing import Pool
from elasticsearch import Elasticsearch, helpers

es = Elasticsearch()
es.indices.create(index='imdb', ignore=400)

identity = lambda x: x if x != '\\N' else None
convert = {
    'startYear': lambda x : int(x) if x != '\\N' else None,
    'endYear': lambda x : int(x) if x != '\\N' else None,
    'runtimeMinutes': lambda x : int(x) if x != '\\N' else None,
    'isAdult': lambda x: True if x == '1' else False,
    'genres': lambda x: x.split(',') if x != '\\N' else None,
}
now = datetime.now()


def data_generator(headers, reader, **kwargs):
    row_count = 0
    for row in reader:
        # prepare data
        data = {}
        for i, field in enumerate(row):
            field_name = headers[i]
            converter = convert.get(field_name, identity)
            try:
                data[field_name] = converter(field)
            except Exception as e:
                print('Error', field_name, '=', field, e)
                data[field_name] = field

        yield {
            '_index': 'imdb',
            '_id': data['tconst'],
            '_source': data
        }

        row_count += 1
        if row_count % 1000 == 0:
            print(row_count)


with gzip.open('title.basics.tsv.gz', mode='rt') as fp:
    reader = csv.reader(fp, delimiter='\t')
    headers = next(reader)
    row_count = 0
    bulk_actions = []
    helpers.bulk(es, data_generator(headers, reader))
```

</details>

The query_types `match` and `simple_query_string` works with minor fixes (see comments).
I'm not sure about `term`, `terms` : it seems to works, but sometimes I don't a result : 

  * `tconst:tt0334159` : ok
  * `startYear:1924`  : ok 
  * `primaryTitle:<value>`: whatever `value` is never returns a result.

I don't know about `custom`.

I guess the engine supposes all the field are string, otherwise some specific code is required to parse  the queries (convert value from str to int for example), and format the results (bool -> Yes, No). This becomes tricky because it involves the i18n (Yes -> True, etc...).

The engine raise an exception with the expected query format doesn't match the user input. This is different from the currency, dictzone and translated engines.

Otherwise LGTM.

searx/engines/elasticsearch.py
dalf: Fixed in commit 7a73b41

searx/engines/elasticsearch.py
dalf: Fixed in commit 7a73b41

dalf (dismissed):

kvch:
@dalf thanks

PR #2295, opened 2020-11-04
[enh] add default http headers - closes #715

### LINKED ISSUE #715: HTTP headers

asciimoo:
## What does this PR do?

Adds default HTTP headers to every response. It is configurable from `settings.yml`.

This PR provides better security/privacy defaults for searx instances.

Easiest way is to call `curl -I http://127.0.0.1:8888/`. It can be also tested in browser with the HTTP inspector.

## Related issues

Closes #715

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -17,6 +17,12 @@ server:
     image_proxy : False # Proxying image results through searx
     http_protocol_version : "1.0"  # 1.0 and 1.1 are supported
     method: "POST" # POST queries are more secure as they don't show up in history but may cause problems when using Firefox containers
+    default_http_headers:
+        X-Content-Type-Options : nosniff
+        X-XSS-Protection : 1; mode=block
+        X-Download-Options : noopen
+        X-Robots-Tag : noindex, nofollow
+        Referrer-Policy : no-referrer
 
 ui:
     static_path : "" # Custom static path - leave it blank if you didn't change
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -488,6 +488,16 @@ def pre_request():
 
 
 @app.after_request
+def add_default_headers(response):
+    # set default http headers
+    for header, value in settings['server'].get('default_http_headers', {}).items():
+        if header in response.headers:
+            continue
+        response.headers[header] = value
+    return response
+
+
+@app.after_request
 def post_request(response):
     total_time = time() - request.start_time
     timings_all = ['total;dur=' + str(round(total_time * 1000, 3))]


dalf:
About X-Content-Security-Policy, with 
```default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self'; script-src 'self' 'unsafe-inline'; connect-src 'self';```

the music / video iframes and OSM maps won't work.


[The value](https://github.com/searx/searx-docker/blob/7ad4d4903d0ce41de4094d5fe0fecda156b41eaa/Caddyfile#L83) in searx-docker:

```X-Content-Security-Policy : "upgrade-insecure-requests; default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; form-action 'self'; font-src 'self'; frame-ancestors 'self'; base-uri 'self'; connect-src 'self' https://overpass-api.de; img-src 'self' data: https://*.tile.openstreetmap.org; frame-src https://www.youtube-nocookie.com https://player.vimeo.com https://www.dailymotion.com https://www.deezer.com https://www.mixcloud.com https://w.soundcloud.com https://embed.spotify.com"```

* [`upgrade-insecure-requests`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/upgrade-insecure-requests) makes sure there is no HTTP request.
* [`default-src 'none'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src) everything is deny except what is explicitly allowed (no worker, no manifest, etc...)
* [`script-src 'self'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) : there is no inline JS.
* [`style-src 'self' 'unsafe-inline'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src) because there are some inline CSS without a nonce.
* `font-src 'self'`: font icons.
* [`frame-ancestors 'self'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) : searx can't included in another URL (debatable).
----

I would add:
```Referrer-Policy : no-referrer```

In case an HTML template has a mistake, the HTTP header is here (personal view point, debatable).

Note: the meta tag is supported by slightly old browser than the HTTP header:
* HTTP header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy (table at the bottom, line `Referrer-Policy`)
* meta tag: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link (table at the bottom, line `referrerpolicy`)

**Comment by return42** [comment]
FWIW: I do not have rich experience with CSP, but on my http instance I needed to set `data: darmarit.org;` to get the base64 encoded `<img src='data:image/....` (from the google-image engine). Here is my V-host entry for CSP:

    Header set Content-Security-Policy "default-src 'self'; img-src 'self' data: darmarit.org;"

(I haven't tested OSM and others)

**Comment by asciimoo** [comment]
@dalf, @return42  you are right, thanks for the review. I've updated the CSP and added Referrer-Policy.

**Comment by dalf** [comment]
I just discovered this extension https://addons.mozilla.org/en-US/firefox/addon/laboratory-by-mozilla/ (_not actively monitored for security by Mozilla_ by developed by April King, the author of https://observatory.mozilla.org/ ).

I confirm that without `darmarit.org` there are some missing images in https://darmarit.org/searx/, I don't understand why. It works with the updated value from this PR (using the FF extension).

It is possible to click "Record this site (will disable existing CSP while recording)" and the CSP header will dynamically updated to allow all requests from the website : after visiting OSM the extension added `img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org`

Using the "Custom CSP policy" option, it is possible to set the same CSP than in searx-docker: 
everything seems to work except on error in the console log triggered by  jquery-1.11.1.min.js:3:4518  that can't load an inline javascript (I have the same error on my instance).

But:
```
$ git grep "<iframe" searx/engines/*.py
searx/engines/dailymotion.py:embedded_url = '<iframe frameborder="0" width="540" height="304" ' +\
searx/engines/deezer.py:embedded_url = '<iframe scrolling="no" frameborder="0" allowTransparency="true" ' +\
searx/engines/invidious.py:        '<iframe width="540" height="304" '
searx/engines/mixcloud.py:embedded_url = '<iframe scrolling="no" frameborder="0" allowTransparency="true" ' +\
searx/engines/peertube.py:        '<iframe width="560" height="315" sandbox="allow-same-origin allow-scripts allow-popups" '
searx/engines/sepiasearch.py:embedded_url = '<iframe width="540" height="304" src="{url}" frameborder="0" allowfullscreen></iframe>'
searx/engines/soundcloud.py:embedded_url = '<iframe width="100%" height="166" ' +\
searx/engines/spotify.py:embedded_url = '<iframe data-src="https://embed.spotify.com/?uri=spotify:track:{audioid}"\
searx/engines/vimeo.py:embedded_url = '<iframe data-src="https://player.vimeo.com/video/{videoid}" ' +\
searx/engines/youtube_api.py:embedded_url = '<iframe width="540" height="304" ' +\
searx/engines/youtube_noapi.py:embedded_url = '<iframe width="540" height="304" ' +\
```

peertube and sepiasearch are missing from [searx-docker configuration](https://github.com/searx/searx-docker/blob/7ad4d4903d0ce41de4094d5fe0fecda156b41eaa/Caddyfile#L83):
* searx/engines/peertube.py : add `https://peer.tube/` to `frame-src`, easy.
* searx/engines/sepiasearch.py: the value can be anything ( search for ```!sep test```  and look at the console).

Either we:
* ignore the sepiasearch engine for now.
* ~investigate the sepiasearch engine results, and every possible website~ I don't think it is possible since the number of peertube instance will change over time.
* open the door to anything / rely on the reverse proxy configuration as it is now (but the recommended configuration in the documentation and searx-docker will have the issue).
* dynamically generate the CSP header according to:
    * the results.
    * the engine configuration like `categories`, `paging`, `safe_search`.... It is more safe but it doesn't solve the problem for the sepiasearch engine. For this engine we can have an option wildcard option to generate the CSP value from the results (I think this is the safest option).

... not simple...

[EDIT] Out of topic in this PR: it shows that the CSP in related to the code. Moreover, if the CSP is dynamically generated, we could allow inline javascript : searx render the HTML template using a random nonce, and then add this nonce to the CSP header.

asciimoo:
@dalf I agree, CSP is quite a complex question. What about removing the CSP from this PR for now and finalize it in a dedicated PR?

asciimoo:
PR is updated/force-pushed.

dalf (approved):
The `X-Frame-Options "SAMEORIGIN"` header can be added at the same time than the CSP header (since the CSP header replace this deprecated header).

About the static files, in the case they are managed by searx and not by a reverse proxy, another entry can be added later in the `sever` section (to be discuss later).

Tested with / without searx-docker setup: LGTM.

PR #2299, opened 2020-11-06
pep8 to pycodestyle

rachmadaniHaryono:
## What does this PR do?

change pep8 dev package requirement to latest pycodestyle

based on the readme from pycodestyle

> This package used to be called pep8 but was renamed to pycodestyle to reduce confusion

https://github.com/PyCQA/pycodestyle

as result makefile will use pycodestyle but test name will still be `test.pep8`

e: some error and warning will be ignored for now

updated package is better

```console
$ make clean test.pep8
```

e: updated with virtualenv
e2: chg based on suggestion

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ test.sh:
 test.pep8: pyenvinstall
 	@echo "TEST      pycodestyle (formerly pep8)"
 	$(Q)$(PY_ENV_ACT); pycodestyle --exclude='searx/static, searx/languages.py, searx/engines/gigablast.py' \
-        --max-line-length=120 --ignore "E402,W503" searx tests
+        --max-line-length=120 --ignore "E117,E252,E402,E722,E741,W503,W504,W605" searx tests
 
 test.unit: pyenvinstall
 	@echo "TEST      tests/unit"
--- a/Makefile
+++ b/Makefile
@@ -235,8 +235,8 @@ test.sh:
 	shellcheck -x .config.sh
 
 test.pep8: pyenvinstall
-	@echo "TEST      pep8"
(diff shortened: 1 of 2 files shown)


return42:
Thanks for your PR.

FYI ..

> $ pip uninstall pep8
> $ pip install pycodestyle

.. will install into user's HOME. Makefile targets do build a pyenv into `./local/py3/`. Further read https://searx.github.io/searx/dev/makefile.html#python-environment

rachmadaniHaryono:
i updated the instruction with virtualenv activation. i also add more configuration to ignore new error and warning

return42:
> How to test this PR locally?

no need to build & source the pyenv, this is already done by the Makefile targets.

[You have never to think about intermediate targets like pyenv or install, the Makefile chains them as requisites. Just run your main target.](https://searx.github.io/searx/dev/makefile.html#python-environment). Typing ...

     make test.pep8

is enough. Every time the requirements.txt or requirements-dev.txt is patched a  [_clean_](https://searx.github.io/searx/dev/makefile.html#id10) is needed, to drop the old pyenv and build new one with the changed requirements:

    make clean test.pep8

Before investigate more time in pycodestyle (aka pep8) I would recommend to have a look at the existing pylint.  @dalf implemented some PRs around to condition a _liniting_ process. The concept he has is already very well developed. A good starting point to read further is [PR 22899](https://github.com/searx/searx/pull/2289#issuecomment-721046822)

rachmadaniHaryono:
i updated the instruction once again.

> #2289 

~~i already read that. does that meant pylint will remove this linter?~~

dalf:
Thank you !

PR #2313, opened 2020-11-12
[WIP] Starlette

dalf:
## What does this PR do?

This is WIP, it is a first attempt to implement https://github.com/searx/searx/wiki/New-architecture-proposal

* The Flask implementation is untouched (minor some cut / paste of some functions)
* The Starlette implementation can be started using `. ./local/py3/bin/activate; python -m searx.web`

i18n:
* There is no Flask_Babel package: see searx.web.i18n (missing function: format_date). 
* There is no `request` obect in the global scope, so `gettext` is available using `request.state.gettext` (currently there is [no lazy / delayed evaluation](https://github.com/searx/searx/blob/0fb8540d672b693eb19b63aa8126e80286d983a5/searx/web/i18n.py#L56) ).
* About Jinja2: see searx.web.templates.SearxJinja2Templates ( `overlay.install_gettext_translations(get_translations(locale), newstyle=True)`)

image proxy:
* use httpx
* see https://www.python-httpx.org/async/#streaming-responses

`searx.web.middleware` should be a AWSGI middleware (currently it requires to `await pre_request(request)` )

`/search`: call SearchWithPlugins using a thread pool. The current implementation is really inefficient:
   * create / reuse a thread to call [`search.search`](https://github.com/searx/searx/blob/aa6eaf603ff9b403dcc32adadbb1e75445f0348c/searx/search.py#L435-L443)
   * then create a thread per engine as usual.
   * the Search and SearchWithPlugins classes should deal with async.

I think [`get_base_url`](https://github.com/searx/searx/blob/aa6eaf603ff9b403dcc32adadbb1e75445f0348c/searx/webapp.py#L264-L272) is not required with starlette : https://www.starlette.io/release-notes/#01211

This change should NOT be merged, it is a request for comment.

`. ./local/py3/bin/activate; python -m searx.web` with / without `SEARX_DEBUG=1`.


dalf:
I'm afraid a solution like https://networklore.com/start-task-with-flask/ while breaks the initialization (with / without uwsgi, debug / not debug, etc).

the target on the long run is starlette : https://github.com/searx/searx/pull/2313

So, despite the fact the check on searx.webapp import creates issues, I think it is still an improvement.

PR #2318, opened 2020-11-16
Add pylint

dalf:
## What does this PR do?

Activate pylint globally with 3 invocations (see commit 72108a22674e6272ae8dcd1a50eab0b5c0a0fae5 )

* one with the default .pylintrc file
* one for searx/engines with some disabled checks and some additional-builtins
* one for the all the code except the engines with some disabled checks

Various fixes are included in the PR:
* remove dead code
* remove duplicate test
* fix code indentations
* fix a long standing bug in the `/stats` URL:  the error section was reporting the error count instead of the error percentage. (see d7814c9963ff6cf7aa36719b85c54789521a9e48 )

It improves the code quality over time.


searx/__init__.py
```diff
@@ -16,9 +16,9 @@
 '''
 
 import logging
+import io
```
kvch: Why is it better according to pylint? I thought it was always better to import only the functions you are using, not the whole moduel.

dalf: the previous code was ```from io import open``` which overrides `open`, pylint is not happy about that.

Actually I think we can use `open` directly, and just remove ```from io import open```.

dalf: Updated.

return42: > Actually I think we can use `open` directly, and just remove `from io import open`.

AFAIK `io.open` was sometimes needed in py2.  In py3 `open` and `io.open` are the same thing.

PR #2319, opened 2020-11-16
[enh] Add multiple outgoing proxies

### LINKED ISSUE #1512: [Enhancement] Add rotating outgoing proxies

### LINKED ISSUE #1899: Proxy rotation on searx

dalf:
## What does this PR do?

credits go to @bauruine see https://github.com/searx/searx/pull/1958

Add support for multiple outgoing proxies.

Public instances require often use multi proxies.
Without this PR, an additional proxy / dispatch layer is required.

Moreover it provides a way to dispatch the requests to different outgoing sources.

## Author's checklist

(should be part of the `make test`)

In one terminal:
```sh
pip install proxy.py
. ./local/py3/bin/activate 
proxy --hostname 127.0.0.1 --port 9000
```

In another terminal:
```
. ./local/py3/bin/activate 
proxy --hostname 127.0.0.1 --port 9001
```

In settings.yml
```
   outgoing:
       proxies:
           http:
               - http://localhost:9000
               - http://localhost:9001
           https:
               - http://localhost:9000
               - http://localhost:9001
```

Look at the proxy logs.

---

Second test with [ EDIT see [comment bellow](https://github.com/searx/searx/pull/2319#pullrequestreview-536045954) ] :
```
  - name : google
    engine : google
    shortcut : go
    outgoing:
       proxies:
           http:
               - http://localhost:9000
               - http://localhost:9001
           https:
               - http://localhost:9000
               - http://localhost:9001
```

Check that there are new connection for google in the proxy logs (there is some delay).

## Related issues

* Close https://github.com/searx/searx/pull/1958
* Close https://github.com/searx/searx/issues/1899
* Close https://github.com/searx/searx/issues/1512

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -25,7 +25,7 @@ from operator import itemgetter
 from searx import settings
 from searx import logger
 from searx.data import ENGINES_LANGUAGES
-from searx.poolrequests import get
+from searx.poolrequests import get, get_proxy_cycles
 from searx.utils import load_module, match_language, get_engine_from_settings
 
 
@@ -79,16 +79,18 @@ def load_engine(engine_data):
         logger.exception('Cannot load engine "{}"'.format(engine_module))
         return None
 
-    for param_name in engine_data:
+    for param_name, param_value in engine_data.items():
         if param_name == 'engine':
-            continue
-        if param_name == 'categories':
-            if engine_data['categories'] == 'none':
+            pass
+        elif param_name == 'categories':
+            if param_value == 'none':
                 engine.categories = []
             else:
-                engine.categories = list(map(str.strip, engine_data['categories'].split(',')))
-            continue
-        setattr(engine, param_name, engine_data[param_name])
+                engine.categories = list(map(str.strip, param_value.split(',')))
+        elif param_name == 'proxies':
+            engine.proxies = get_proxy_cycles(param_value)
+        else:
+            setattr(engine, param_name, param_value)
 
     for arg_name, arg_value in engine_default_args.items():
         if not hasattr(engine, arg_name):
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -111,6 +111,32 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
+def get_proxy_cycles(proxy_settings):
+    if not proxy_settings:
+        return None
+    # Backwards compatibility for single proxy in settings.yml
+    for protocol, proxy in proxy_settings.items():
+        if isinstance(proxy, str):
+            proxy_settings[protocol] = [proxy]
+
+    for protocol in proxy_settings:
+        proxy_settings[protocol] = cycle(proxy_settings[protocol])
+    return proxy_settings
+
+
+GLOBAL_PROXY_CYCLES = get_proxy_cycles(settings['outgoing'].get('proxies'))
+
+
+def get_proxies(proxy_cycles):
+    if proxy_cycles:
+        return {protocol: next(proxy_cycle) for protocol, proxy_cycle in proxy_cycles.items()}
+    return None
+
+
+def get_global_proxies():
+    return get_proxies(GLOBAL_PROXY_CYCLES)
+
+
 def request(method, url, **kwargs):
     """same as requests/requests/api.py request(...)"""
     time_before_request = time()
@@ -119,8 +145,8 @@ def request(method, url, **kwargs):
     session = SessionSinglePool()
 
     # proxies
-    if kwargs.get('proxies') is None:
-        kwargs['proxies'] = settings['outgoing'].get('proxies')
+    if not kwargs.get('proxies'):
+        kwargs['proxies'] = get_global_proxies()
 
     # timeout
     if 'timeout' in kwargs:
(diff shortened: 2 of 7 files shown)


searx/engines/__init__.py
```diff
@@ -79,16 +79,18 @@ def load_engine(engine_data):
         logger.exception('Cannot load engine "{}"'.format(engine_module))
         return None
 
-    for param_name in engine_data:
+    for param_name, param_value in engine_data.items():
         if param_name == 'engine':
-            continue
-        if param_name == 'categories':
-            if engine_data['categories'] == 'none':
+            pass
```
kvch, line 84: `pass`? Why not just remove this branch of the `if` if nothing has to happen in case of engines?

searx/poolrequests.py
```diff
@@ -111,6 +111,32 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
+def get_proxy_cycles(proxy_settings):
+    if not proxy_settings:
+        return None
+    # Backwards compatibility for single proxy in settings.yml
```
kvch, line 117: :tada:

searx/poolrequests.py
```diff
@@ -111,6 +111,32 @@ def get_time_for_thread():
...
+    for protocol, proxy in proxy_settings.items():
+        if isinstance(proxy, str):
+            proxy_settings[protocol] = [proxy]
+    #
```
kvch: Leftover?

searx/poolrequests.py
```diff
@@ -111,6 +111,32 @@ def get_time_for_thread():
     return threadLocal.total_time
 
 
+def get_proxy_cycles(proxy_settings):
```
kvch, line 114: Maybe add a few unit tests? These functions can be tested quite easily.

searx/engines/__init__.py
dalf: By default, all the attributes are copied to `engine`, see line 92 and 93. 
This line says explicitly that the `engine` attribute name is skipped.
May be I should add a comment about that and / or move this condition just before the check.

Alternative:
```python
if param_name == 'engine':
   continue
if param_name == 'categories':
   ...
```

I prefer `if ... elif ... elif ... else`

searx/poolrequests.py
dalf: See last commit, tests of:
* get_proxy_cycles : old syntax, new syntax.
* get_proxies
* request: with and without the `proxies=` parameter

searx/poolrequests.py
dalf: Removed
The initial intent was too close the backward compatible "section", a blank line is enough.

unixfox:
It works great in the global settings and inside an engine settings but the example for the engine proxies is incorrect.

It should be:
```yml
- name : google
    engine : google
    shortcut : go
    proxies:
       http:
           - http://localhost:9000
           - http://localhost:9001
       https:
           - http://localhost:9000
           - http://localhost:9001
```

Instead of:
```yml
- name : google
    engine : google
    shortcut : go
    outgoing:
       proxies:
           http:
               - http://localhost:9000
               - http://localhost:9001
           https:
               - http://localhost:9000
               - http://localhost:9001
```

Or maybe @dalf you forgot to include `outgoing` in the code?

unixfox (approved):
It works great in the global settings and inside an engine settings but the example for the engine proxies is incorrect.

It should be:
```yml
- name : google
    engine : google
    shortcut : go
    proxies:
       http:
           - http://localhost:9000
           - http://localhost:9001
       https:
           - http://localhost:9000
           - http://localhost:9001
```

Instead of:
```yml
- name : google
    engine : google
    shortcut : go
    outgoing:
       proxies:
           http:
               - http://localhost:9000
               - http://localhost:9001
           https:
               - http://localhost:9000
               - http://localhost:9001
```

Or maybe @dalf you forgot to include `outgoing` in the code?

dalf:
Thank you.

Yes there is a mistake in the top comment, it should be:
```yaml
- name : google
    engine : google
    shortcut : go
    proxies:
       http:
           - http://localhost:9000
           - http://localhost:9001
       https:
           - http://localhost:9000
           - http://localhost:9001
```

it keeps the compatibility with the existing code:
https://github.com/searx/searx/blob/2fc3b17c85512280173bb47f449cc2faa64b0501/searx/search.py#L120-L122

PR #1591, opened 2019-05-17
Feature/standalone searx update

rachmadaniHaryono:
- add standalone_searx to docs using autodoc
- test for module
- example on module docstring include how to run from cli or python console
- split the script to several group: argument parser, dict converter, search query getter

what is not on the pr:

setting SEARX_DEBUG to pass check from #2256 will break ptipython console.
some doc assume user already change secretkey

i recommend searx's secretkey also have env var so user can change it directly without change yaml config or enable debug mode   

<details>
  <summary>old description</summary>
  
- the module is split into ~~two~~ three, argument parser function, result getter function and main function that display how to use searx as standalone
- some change is added because python3 on main function
- i added typing package because i use this to check main function
- change engine shortcut check because on test the check is twice and current check fail to recognize that
- the test is very basic and it don't check the con/tent because on each test on different computer may change

e: example python3 code using standalone_searx

```python
from typing import Any
import importlib.util

import searx
from searx import settings

search_query = 'rain'
module_name = 'utils.standalone_searx'
filename = 'utils/standalone_searx.py'
spec = importlib.util.spec_from_file_location(module_name, filename)
ss = importlib.util.module_from_spec(spec)  # type: Any
spec.loader.exec_module(ss)  # type: ignore

searx.engines.initialize_engines(settings['engines'])  # type: ignore
args = ss.parse_argument([search_query, '--category', 'images'])
s_query, res  = ss.get_result(args)
```</details>

PR #2321, opened 2020-11-17
[fix] fix a test_standalone_searx test case

dalf:
## What does this PR do?

If test_engines_init.py runs before test_standalone_searx.py, the engine list is not empty.
It makes test_get_search_query flaky.

This commit initializes the engline list in test_standalone_searx.py

Fix a flaky test with a well define runtime.

... it is not easy to reproduce:
* @return42 has experienced this error
* the github action branch has the same error : see https://github.com/dalf/searx/runs/1411204212?check_suite_focus=true#step:6:167

## Related issues

Related to : 
* https://github.com/searx/searx/pull/1591#issuecomment-723541549
* https://github.com/searx/searx/pull/2309

### DIFF
--- a/tests/unit/test_standalone_searx.py
+++ b/tests/unit/test_standalone_searx.py
@@ -8,7 +8,8 @@ import sys
 from mock import Mock, patch
 from nose2.tools import params
 
-from searx.search import SearchQuery
+from searx.search import SearchQuery, EngineRef
+from searx.engines import initialize_engines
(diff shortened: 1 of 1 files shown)


return42:
> Fix a flaky test with a well define runtime.

Sorry for the late reply .. I can confirm that my issue mentioned in https://github.com/searx/searx/pull/1591#issuecomment-723541549 is fixed.  Thanks a lot for your investigation!

PR #2325, opened 2020-11-19
Add recoll engine

kvch:
This is the rebased version of https://github.com/searx/searx/pull/1257

Closes #1257

### DIFF
--- a/searx/engines/recoll.py
+++ b/searx/engines/recoll.py
@@ -1,0 +1,106 @@
+"""
+ Recoll (local search engine)
+
+ @using-api   yes
+ @results     JSON
+ @stable      yes
+ @parse       url, content, size, abstract, author, mtype, subtype, time, \
+              filename, label, type, embedded
+"""
+
+from json import loads
+from searx.url_utils import urlencode, quote
+from datetime import date, timedelta
+
+# engine dependent config
+paging = True
+time_range_support = True
+
+# parameters from settings.yml
+base_url = None
+search_dir = ''
+mount_prefix = None
+dl_prefix = None
+
+# embedded
+embedded_url = '<{ttype} controls height="166px" ' +\
+    'src="{url}" type="{mtype}"></{ttype}>'
+
+
+# helper functions
+def get_time_range(time_range):
+    sw = {
+        'day': 1,
+        'week': 7,
+        'month': 30,
+        'year': 365
+    }
+
+    offset = sw.get(time_range, 0)
+    if not offset:
+        return ''
+
+    return (date.today() - timedelta(days=offset)).isoformat()
+
+
+# do search-request
+def request(query, params):
+    search_after = get_time_range(params['time_range'])
+    search_url = base_url + 'json?{query}&highlight=0'
+    params['url'] = search_url.format(query=urlencode({
+        'query': query,
+        'page': params['pageno'],
+        'after': search_after,
+        'dir': search_dir}))
+
+    return params
+
+
+# get response from search-request
+def response(resp):
+    results = []
+
+    response_json = loads(resp.text)
+
+    if not response_json:
+        return []
+
+    for result in response_json.get('results', []):
+        title = result['label']
+        url = result['url'].replace('file://' + mount_prefix, dl_prefix)
+        content = u'{}'.format(result['snippet'])
+
+        # append result
+        item = {'url': url,
+                'title': title,
+                'content': content,
+                'template': 'files.html'}
+
+        if result['size']:
+            item['size'] = int(result['size'])
+
+        for parameter in ['filename', 'abstract', 'author', 'mtype', 'time']:
+            if result[parameter]:
+                item[parameter] = result[parameter]
+
+        # facilitate preview support for known mime types
+        if 'mtype' in result:
+            (mtype, subtype) = result['mtype'].split('/')
+            item['mtype'] = mtype
+            item['subtype'] = subtype
+
+            if mtype in ['audio', 'video']:
+                item['embedded'] = embedded_url.format(
+                    ttype=mtype,
+                    url=quote(url.encode('utf8'), '/:'),
+                    mtype=result['mtype'])
+
+            if mtype in ['image'] and subtype in ['bmp', 'gif', 'jpeg', 'png']:
+                item['img_src'] = url
+
+        results.append(item)
+
+    if 'nres' in response_json:
+        results.append({'number_of_results': response_json['nres']})
+
+    return results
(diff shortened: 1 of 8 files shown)


searx/settings.yml
```diff
@@ -648,6 +648,64 @@ engines:
     shortcut : qws
     categories : social media
 
+    # recoll is a local search engine based on Xapian:
```
dalf: Would it better to add this documentation into docs/admin/engines/recoll.rst ?
So settings.yml stays "light".

kvch: Where should I put it? AFAIK we don't have special documentation for engines.

return42: I vote for adding a subsection to [admin/engines](https://github.com/searx/searx/blob/master/docs/admin/engines.rst)  for each engine which needs more documentation. This can also be done by inserting a `.. toctree::` at the top and adding a file for each engine.  If [reST](https://searx.github.io/searx/dev/reST.html) is not your part, I can prepare a patch on top of this PR.

kvch: @return42 I would be grateful if you would prepare a PR with the proper page layout. Structuring documentation is not my strong suit. I have added a new page with the description of the engine, so you do not have to write that yourself.

return42: I prepared two patches (e9134fa77c6 and 7e69765) you can pull from my [recollengine](https://github.com/return42/searx/tree/recollengine) branch.

> Structuring documentation is not my strong suit.

No problem, we all have our own majors and minors interests .. if you like to see other things documented, ask me.

dalf:
I have install recoll and recoll-webui using https://github.com/dalf/recoll (Docker but it shouldn't be a problem to do the same thing without it).

Important note: 
* I haven't install https://github.com/Yetangitu/recoll-webui/tree/jsonpage because it is Python 2 (and the code doesn't work with Python3).
* but  https://framagit.org/medoc92/recollwebui.git so no paging.

I haven't test the media embedding ( `if mtype in ['audio', 'video']:` )

It works with this patch:
```patch
diff --git a/searx/engines/recoll.py b/searx/engines/recoll.py
index a5d8dd74..019f9aa1 100644
--- a/searx/engines/recoll.py
+++ b/searx/engines/recoll.py
@@ -9,7 +9,7 @@
 """
 
 from json import loads
-from searx.url_utils import urlencode, quote
+from urllib.parse import urlencode, quote
 from datetime import date, timedelta
 
 # engine dependent config
@@ -68,7 +68,7 @@ def response(resp):
     for result in response_json.get('results', []):
         title = result['label']
         url = result['url'].replace('file://' + mount_prefix, dl_prefix)
-        content = u'{}'.format(result['snippet'])
+        content = '{}'.format(result['snippet'])
 
         # append result
         item = {'url': url,
@@ -84,7 +84,7 @@ def response(resp):
                 item[parameter] = result[parameter]
 
         # facilitate preview support for known mime types
-        if 'mtype' in result:
+        if 'mtype' in result and '/' in result['mtype']:
             (mtype, subtype) = result['mtype'].split('/')
             item['mtype'] = mtype
             item['subtype'] = subtype
```

Once https://github.com/searx/searx/issues/2067 is done and looking at [the UI code](https://framagit.org/medoc92/recollwebui/-/blob/master/webui.py#L254-271), IMO, the engine could use directly the python3-recoll package instead of using a deprecated UI.

dalf (changes requested):
I have install recoll and recoll-webui using https://github.com/dalf/recoll (Docker but it shouldn't be a problem to do the same thing without it).

Important note: 
* I haven't install https://github.com/Yetangitu/recoll-webui/tree/jsonpage because it is Python 2 (and the code doesn't work with Python3).
* but  https://framagit.org/medoc92/recollwebui.git so no paging.

I haven't test the media embedding ( `if mtype in ['audio', 'video']:` )

It works with this patch:
```patch
diff --git a/searx/engines/recoll.py b/searx/engines/recoll.py
index a5d8dd74..019f9aa1 100644
--- a/searx/engines/recoll.py
+++ b/searx/engines/recoll.py
@@ -9,7 +9,7 @@
 """
 
 from json import loads
-from searx.url_utils import urlencode, quote
+from urllib.parse import urlencode, quote
 from datetime import date, timedelta
 
 # engine dependent config
@@ -68,7 +68,7 @@ def response(resp):
     for result in response_json.get('results', []):
         title = result['label']
         url = result['url'].replace('file://' + mount_prefix, dl_prefix)
-        content = u'{}'.format(result['snippet'])
+        content = '{}'.format(result['snippet'])
 
         # append result
         item = {'url': url,
@@ -84,7 +84,7 @@ def response(resp):
                 item[parameter] = result[parameter]
 
         # facilitate preview support for known mime types
-        if 'mtype' in result:
+        if 'mtype' in result and '/' in result['mtype']:
             (mtype, subtype) = result['mtype'].split('/')
             item['mtype'] = mtype
             item['subtype'] = subtype
```

Once https://github.com/searx/searx/issues/2067 is done and looking at [the UI code](https://framagit.org/medoc92/recollwebui/-/blob/master/webui.py#L254-271), IMO, the engine could use directly the python3-recoll package instead of using a deprecated UI.

searx/engines/recoll.py
```diff
@@ -0,0 +1,106 @@
...
+
+ @using-api   yes
+ @results     JSON
+ @stable      yes
+ @parse       url, content, size, abstract, author, mtype, subtype, time, \
+              filename, label, type, embedded
+"""
+
+from json import loads
+from searx.url_utils import urlencode, quote
```
dalf: must be `from urllib.parse import urlencode, quote`

searx/engines/recoll.py
```diff
@@ -0,0 +1,106 @@
...
+
+    response_json = loads(resp.text)
+
+    if not response_json:
+        return []
+
+    for result in response_json.get('results', []):
+        title = result['label']
+        url = result['url'].replace('file://' + mount_prefix, dl_prefix)
+        content = u'{}'.format(result['snippet'])
```
dalf: `u'{}'` --> `'{}'`.

searx/engines/recoll.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if result['size']:
+            item['size'] = int(result['size'])
+
+        for parameter in ['filename', 'abstract', 'author', 'mtype', 'time']:
+            if result[parameter]:
+                item[parameter] = result[parameter]
+
+        # facilitate preview support for known mime types
+        if 'mtype' in result:
```
dalf: `result['mtype']` can be empty.
`if 'mtype' in result and '/' in result['mtype']:` avoids an exception line 88.

searx/settings.yml
kvch: I have picked the patches. Thank you!

dalf (approved):
All good

PR #2326, opened 2020-11-21
WIP: feat(engine): support DICT protocol, add dict engine

komachi:
This adds support for [DICT protocol](https://en.wikipedia.org/wiki/DICT).

DICT is very simple to use, so on average linux distro this should work like `sudo apt install dictd dict-freedict-eng-jpn` which allows you to query your localhost dictd server. You can also use remote DICT server, like dict.org.

As DICT is not HTTP, this requires more changes than usual for a new engine. In current state it works for me, but I probably missed something in implementation, so I ask for additional feedback.

searx_dict cleaned

## Checklist

- [ ] : add unit tests

### DIFF
--- a/searx/engines/dict.py
+++ b/searx/engines/dict.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python
+
+"""
+ DICT protocol
+ @website     https://tools.ietf.org/html/rfc2229
+ @provide-api yes
+ @using-api   yes
+ @results     DICT
+ @stable      yes
+ @parse       title, content
+"""
+
+from dictionary_client import DictionaryClient
+
+categories = ['general']
+paging = False
+db = '*'
+host = 'localhost'
+port = 2628
+result_template = 'dict.html'
+DICT = True
+
+
+def request(query, params):
+    params['query'] = query
+    params['method'] = 'define'
+    return params
+
+
+def response(resp):
+    results = []
+    if not resp or not resp['content']:
+        return results
+
+    for definition in resp['content']:
+        results.append({
+            'title': resp['databases'][definition['db']],
+            'content': definition['definition'],
+            'template': 'dict.html'
+        })
+    return results
--- a/searx/poolrequests.py
+++ b/searx/poolrequests.py
@@ -4,6 +4,7 @@ from itertools import cycle
 from threading import RLock, local
 
 import requests
+import socks
 
 from searx import settings
 from searx import logger
@@ -79,7 +80,6 @@ if settings['outgoing'].get('source_ips'):
 else:
     http_adapters = cycle((HTTPAdapterWithConnParams(pool_connections=connect, pool_maxsize=maxsize), ))
     https_adapters = cycle((HTTPAdapterWithConnParams(pool_connections=connect, pool_maxsize=maxsize), ))
-
 
 class SessionSinglePool(requests.Session):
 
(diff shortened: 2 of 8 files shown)


searx/engines/dict.py
```diff
@@ -0,0 +1,41 @@
...
+ @results     DICT
+ @stable      yes
+ @parse       title, content
+"""
+
+from dictionary_client import DictionaryClient
+
+categories = ['general']
+paging = False
+db = '*'
```
dalf, line 17: According to https://pypi.org/project/py-dict-client/ db can be, for example, `'fra-eng'` ?

If yes, there are two existing dictionary engines:
* https://github.com/searx/searx/blob/master/searx/engines/dictzone.py
* https://github.com/searx/searx/blob/master/searx/engines/translated.py

There are 'activated' when the query looks like `[lang1]-[lang2] [query]`, for example `en-fr house` (in the general category).

Do you think something similar can be implemented ? Probably using a mapping between the searx languages and the dict languages ?

searx/poolrequests.py
```diff
@@ -182,3 +191,60 @@ def patch(url, data=None, **kwargs):
 
 def delete(url, **kwargs):
     return request('delete', url, **kwargs)
+
+
+def dict_request(method, query, **kwargs):
```
dalf: `poolrequests.py` could be renamed `https.py` or `http_client.py`, because it focuses on HTTP requests.

It would be better have a different file for the dict protocol.

dalf (commented):
See this draft https://github.com/searx/searx/pull/2225

Especially:
* [abstract.py](https://github.com/searx/searx/blob/c7cd53c5a85085112cf8bace367b7605851e49e9/searx/search/processors/abstract.py)
* [online.py](https://github.com/searx/searx/blob/c7cd53c5a85085112cf8bace367b7605851e49e9/searx/search/processors/online.py)
* [online_dictionary.py](https://github.com/searx/searx/blob/c7cd53c5a85085112cf8bace367b7605851e49e9/searx/search/processors/online_dictionary.py)

The difference here is about the online_dictionary.

A possible change in #2225 : each processor inherits of two classes:
* how to process the request (HTTP, dict protocol, offline, ...)
* how to process / prepare the request (general, dict, currency).

searx/engines/dict.py
komachi: DICT is not limited to multilingual dictionaries, so this couldn't cover every case. Freedict.org dictionaries packaged with Debian follows `fd-[lang1]-[lang2]` pattern, when Mueller's English-Russian dictionary packaged with Debian named `mueller7`, and it could be different with different packaging. This should probably be specified in settings.yml, e.g. if this a multilingual dict, so it could be used as it.

searx/poolrequests.py
komachi: Some engines imports modules from poolrequests, so this requires a lot of changes
```
$ rg poolrequests
searx/autocomplete.py
28:from searx.poolrequests import get as http_get
searx/poolrequests.py
13:logger = logger.getChild('poolrequests')
searx/search.py
26:import searx.poolrequests as requests_lib
searx/engines/soundcloud.py
19:from searx.poolrequests import get as http_get
searx/engines/wolframalpha_noapi.py
15:from searx.poolrequests import get as http_get
searx/engines/wikidata.py
23:from searx.poolrequests import post, get
searx/engines/__init__.py
28:from searx.poolrequests import get
searx/engines/pubmed.py
18:from searx.poolrequests import get
searx/engines/gigablast.py
19:from searx.poolrequests import get
searx/engines/yggtorrent.py
16:from searx.poolrequests import get as http_get
searx/engines/duckduckgo_images.py
20:from searx.poolrequests import get
```

Separate `dict_client.py` shouldn't have any drawback.

PR #2327, opened 2020-11-22
Add preference for displaying advanced settings

### LINKED ISSUE #696: Make different forms of search appear by default without having to click "advanced settings"
One of the really useful part of Searx is being able to search through files or videos. Having that option hidden under "advanced settings" is annoying, especially for mobile where its hard to click the advanced setting without click the first search result. I would like to have to appear by default on my instance. Is this possible? 

it would be nice if the old default theme was kept as a legacy theme

renyhp:
## What does this PR do?

Fix #696, as suggested by a [comment](https://github.com/searx/searx/issues/696#issuecomment-713047361) there

See #696 

Just pull, then open preferences, set Show Advanced Settings to On, then reopen the index page

## Author's checklist

I simply executed the changes as stated in the comment, but I don't fully understand them, so there needs to be a review.

Also, if I add a `advanced_search : True` line under `ui:` in `/etc/searx/settings.yml`, whenever I make a search I get a banner saying "Error: invalid preferences". I don't know if that's intentional; it would be nice anyway if the default setting could be set in `/etc/searx/settings.yml` (which is what I was trying to do), and if this PR does not do that, or does in a broken way, I am willing to make the required changes, provided that I get a little bit of help on how to do that :)

## Related issues

Closes #696

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -545,6 +545,9 @@ def index_error(output_format, error_message):
 def index():
     """Render index page."""
 
+    # UI
+    advanced_search = request.form.get('advanced_search', request.preferences.get_value('advanced_search'))
+
     # redirect to search if there's a query in the request
     if request.form.get('q'):
         query = ('?' + request.query_string.decode()) if request.query_string else ''
@@ -553,6 +556,7 @@ def index():
     return render(
         'index.html',
         selected_categories=get_selected_categories(request.preferences, request.form),
+        advanced_search=advanced_search,
     )
 
 
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -546,7 +546,7 @@ def index():
     """Render index page."""
 
     # UI
-    advanced_search = request.form.get('advanced_search', request.preferences.get_value('advanced_search'))
+    advanced_search = request.preferences.get_value('advanced_search')
 
     # redirect to search if there's a query in the request
     if request.form.get('q'):
@@ -611,7 +611,7 @@ def search():
         return redirect(result_container.redirect_url)
 
     # UI
-    advanced_search = request.form.get('advanced_search', None)
+    advanced_search = request.preferences.get_value('advanced_search')
 
     # Server-Timing header
     request.timings = result_container.get_timings()
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -577,6 +577,7 @@ def search():
         if output_format == 'html':
             return render(
                 'index.html',
+                advanced_search=request.preferences.get_value('advanced_search')
                 selected_categories=get_selected_categories(request.preferences, request.form),
             )
         else:
@@ -611,7 +612,10 @@ def search():
         return redirect(result_container.redirect_url)
 
     # UI
-    advanced_search = request.preferences.get_value('advanced_search')
+    # 'q' in request.from, possible value from request.form.get('advanced_search'):
+    # * 'on': the checkbox is checked
+    # * None: the checkbox is unchecked or request is sent from opensearch.xml
+    advanced_search = request.form.get('advanced_search')
 
     # Server-Timing header
     request.timings = result_container.get_timings()
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -577,7 +577,7 @@ def search():
         if output_format == 'html':
             return render(
                 'index.html',
-                advanced_search=request.preferences.get_value('advanced_search')
+                advanced_search=request.preferences.get_value('advanced_search'),
                 selected_categories=get_selected_categories(request.preferences, request.form),
             )
         else:
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -611,12 +611,6 @@ def search():
     if result_container.redirect_url:
         return redirect(result_container.redirect_url)
 
-    # UI
-    # 'q' in request.from, possible value from request.form.get('advanced_search'):
-    # * 'on': the checkbox is checked
-    # * None: the checkbox is unchecked or request is sent from opensearch.xml
-    advanced_search = request.form.get('advanced_search')
-
     # Server-Timing header
     request.timings = result_container.get_timings()
 
@@ -725,7 +719,6 @@ def search():
         pageno=search_query.pageno,
         time_range=search_query.time_range,
         number_of_results=format_decimal(number_of_results),
-        advanced_search=advanced_search,
         suggestions=suggestion_urls,
         answers=result_container.answers,
         corrections=correction_urls,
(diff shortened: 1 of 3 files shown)


searx/webapp.py
```diff
@@ -545,6 +545,9 @@ def index_error(output_format, error_message):
 def index():
     """Render index page."""
 
+    # UI
+    advanced_search = request.form.get('advanced_search', request.preferences.get_value('advanced_search'))
```
dalf: Line 614 (in the search function) should be replaced by the same code.

renyhp: Well, I can commit the change you're suggesting, but the "Invalid settings" banner still appears anyway...

dalf:
A not obvious _partial_ patch:

```diff
diff --git a/searx/preferences.py b/searx/preferences.py
index b37f0104..a9f16ff7 100644
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -397,11 +397,13 @@ class Preferences:
                 choices=['', 'logicodev', 'logicodev-dark', 'pointhi']),
             'advanced_search': MapSetting(
                 settings['ui'].get('advanced_search', False),
+                is_locked('advanced_search'),
                 map={
                     '0': False,
                     '1': True,
                     'False': False,
-                    'True': True
+                    'True': True,
+                    'on': True,
                 }
             ),
         }
diff --git a/searx/webapp.py b/searx/webapp.py
index 83855d65..933c7007 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -545,7 +545,7 @@ def index():
     """Render index page."""
 
     # UI
-    advanced_search = request.form.get('advanced_search', request.preferences.get_value('advanced_search'))
+    advanced_search = request.preferences.get_value('advanced_search')
 
     # redirect to search if there's a query in the request
     if request.form.get('q'):
@@ -610,7 +610,7 @@ def search():
         return redirect(result_container.redirect_url)
 
     # UI
-    advanced_search = request.form.get('advanced_search', None)
+    advanced_search = request.preferences.get_value('advanced_search')
 
     # Server-Timing header
     request.timings = result_container.get_timings()
```

"Because" of this line: https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/templates/oscar/advanced.html#L1

`advanced_search` is set to `on` when the user sends a search.

BUT: if the user unchecks `advanced_search` checkbox then the `advanced_search` value (in the HTML form) is not set, so searx uses the value from the preferences or the settings.

dalf:
A second patch:

```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index 933c7007..f8007710 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -577,6 +577,7 @@ def search():
             return render(
                 'index.html',
                 selected_categories=get_selected_categories(request.preferences, request.form),
+                advanced_search=request.preferences.get_value('advanced_search')
             )
         else:
             return index_error(output_format, 'No query'), 400
@@ -610,7 +611,10 @@ def search():
         return redirect(result_container.redirect_url)
 
     # UI
-    advanced_search = request.preferences.get_value('advanced_search')
+    # 'q' in request.from, possible value from request.form.get('advanced_search'):
+    # * 'on': the checkbox is checked.
+    # * None: the checkbox is unchecked or request is sent from opensearch.xml
+    advanced_search = request.form.get('advanced_search')
 
     # Server-Timing header
     request.timings = result_container.get_timings()
```

Explainations:
* URL `/`: advanced_search is read from the preferences (either from user preferences, either from the settings).
* URL `/search`: advanced_search is read from the form. The value can be either `on` or not defined (HTML checkbox behavior).
* URL `/search`, exception where there is no query: copy the behavior from `/`, read  advanced_search from the preferences.

After that it should be okay.

renyhp:
Ok, it has been done. One thing I don't understand: what is this checkbox that you speak of, when there is a query? I can't see any.

dalf:
There is a missing coma at the end of this line:
https://github.com/renyhp/searx/blob/4979b4f9d98186d0a5a8d10c096673c596462eb5/searx/webapp.py#L580

---

When you click "Advanced settings", you actually click on a checkbox:
https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/templates/oscar/advanced.html#L1

That's why you have the map is:
```python
{
                   '0': False,
                   '1': True,
                   'False': False,
                   'True': True
                   'True': True,
                   'on': True,      # 'on' when the checkbox is checked = when "Advanced settings" are shown.
}
```

return42:
To prevent committing [syntax or other errors](https://github.com/searx/searx/pull/2327/checks?check_run_id=1459514948#step:6:162) use [make test](https://searx.github.io/searx/dev/makefile.html#make-test) before commit.

renyhp:
> When you click "Advanced settings", you actually click on a checkbox:

Yes, I figured that, but is that checkbox also present in `/search?q=query` ? If yes, where is it? If not, what is line 618 doing?

dalf:
> If yes, where is it? If not, what is line 618 doing?

TLDR: nothing, it can be removed, the [line 719](https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/webapp.py#L719) can be removed too.

According to `git grep` the only templates to use the `advanced_search` variable are:
* `searx/templates/oscar/preferences.html` (in this PR).
* `searx/templates/oscar/advanced.html`: https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/templates/oscar/advanced.html#L1


`searx/templates/oscar/search_full.html` is the only template to reference `searx/templates/oscar/advanced.html`.
https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/templates/oscar/search_full.html#L16


`searx/templates/oscar/index.html` is the only template to reference `searx/templates/oscar/search_full.html`.
https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/templates/oscar/index.html#L18

Since the `/search` URL uses the `oscar/results.html` template, this line is useless:
https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/webapp.py#L719

So this line is useless too:
https://github.com/searx/searx/blob/aec6a2656c4af5aa307914b78cb4f46f4238109a/searx/webapp.py#L609

This line was introduced by the commit 67e11c42b973932c8f568d80a0f25bfd7fc150ab , at that time, the oscar template was using the "advanced_search" (`oscar/results.html` --> `oscar/search.html` --> `oscar/advanced.html`), it is no more the case.

Thank you for to have asked the question !

dalf (approved):

dalf:
Thank you !

Issue #2329, opened 2020-11-23
Theme: discard all the themes except oscar?


dalf:
Currently, the only theme with the required features is the oscar theme.

See the different PR: the only updated theme is oscar.
So my guess is the oscar is wildly use compare to the other themes

May be discard all the themes is exaggerated, so different approach can be this:
* add some statistics about the theme usage: how many of the requests use the oscar theme in percentage.
* decide what to do after looking at the statistics of the public instances.

(related to https://github.com/searx/searx/issues/1025 )

return42:
> May be discard all the themes is exaggerated, so different approach can be this:

I cannot estimate how complex it is, but would a plug-in infrastructure for themes be a solution?

allendema:
Those mentioned themes are as of Pull [2344](https://github.com/searx/searx/pull/2344) discarded.
So v1.0.0 ships without them.

Issue #1953, opened 2020-05-13
Pix-art is broken


lou-bi:
# Results are not displayed
![image](https://user-images.githubusercontent.com/29218400/81867575-33bba500-9571-11ea-9ba5-a72224e4a5f5.png)
# Results are slow
## With pix-art
![image](https://user-images.githubusercontent.com/29218400/81867843-a0cf3a80-9571-11ea-866a-2d0f318af9dd.png)
## With oscar:
![image](https://user-images.githubusercontent.com/29218400/81868045-f4418880-9571-11ea-9678-31cf014fd535.png)

Plus, looking at the results in the dom, they seems to be different between pix-art and oscar. Is that normal ?

unixfox:
This one I'm able to replicate it on another Searx instances.
I haven't test it on a local instance though.

allendema:
This theme is nowdays removed by default.

PR #2344, opened 2020-12-01
[mod] themes: remove legacy, courgette and pix-art themes

dalf:
## What does this PR do?

Remove deprecated themes:
* legacy
* courgette
* pix-art: see #1953

These themes are not maintained:
* legacy doesn't have [result_templates/files.html](https://github.com/searx/searx/blob/master/searx/templates/oscar/result_templates/files.html) nor [results_key_values.html](https://github.com/searx/searx/blob/master/searx/templates/legacy/result_templates/key-value.html)
* the preferences doesn't allow to provide tokens
* and most probably other things.
* moreover [mootools](https://github.com/searx/searx/blob/master/searx/static/js/mootools-core-1.4.5-min.js) is not maintained anymore, see https://github.com/mootools/mootools-core

Check searx has no issue without these themes.

## Related issues

Related to https://github.com/searx/searx/issues/2329


dalf:
I give few more days, for some feedback.

After that, if there is no comment, I'm going to merge this PR.

return42 (dismissed):
Sorry for the late reply / I made some tests now / LGTM, thank you for cleaning up the code!

Issue #2161, opened 2020-08-30
Invidio.us shutting down, video result replacement options


ertect:
Invidio.us has announced it will soon close, here are the alternative instances they provide for when that happens:
https://instances.invidio.us/
  
- yewtu.be should replace invidio.us as source for video look-ups in searx in my experience of the alternatives.
- yewtu.be has been the best instance when you consider all the comprimises in speed/functionality that you have with other instances.
- other popular instances such as invidious.snopyta.org are too slow, or even faster alternatives like invidious.xyz lack the ability to open channel pages, & instead just return an error.

unixfox:
Hello,

I'm the maintainer of yewtu.be.

Unfortunately right now Searx won't work with the API of yewtu.be because it is blocked by the anti bot protection.

Also I don't really want to have my instance included by default in Searx because this could potentially slow it down if too many API requests are made by a bunch of Searx instances.

I still plan to change the invidious instance to yewtu.be on my searx instance (searx.be) though. And I'll allow anyone that have a Searx instance to manually switch to yewtu.be for the invidious engine.

By the way thank you for your kind of words of saying that it's a good invidious instance. I know it requires to have JavaScript enabled in order to visit the instance but that's unfortunately the only way I found for stopping bots and have a stable instance which can load without difficulties a video with DASH 1080p.

andrew6838:
> Hello,
> 
> I'm the maintainer of yewtu.be.
> 
> Unfortunately right now Searx won't work with the API of yewtu.be because it is blocked by the anti bot protection.
> 
> Also I don't really want to have my instance included by default in Searx because this could potentially slow it down if too many API requests are made by a bunch of Searx instances.
> 
> I still plan to change the invidious instance to yewtu.be on my searx instance (searx.be) though. And I'll allow anyone that have a Searx instance to manually switch to yewtu.be for the invidious engine.
> 
> By the way thank you for your kind of words of saying that it's a good invidious instance. I know it requires to have JavaScript enabled in order to visit the instance but that's unfortunately the only way I found for stopping bots and have a stable instance which can load without difficulties a video with DASH 1080p.

Hi ,

your Searx.be is great, can I have source code of it?


Thanks

unixfox:
> Hi ,
> 
> your Searx.be is great, can I have source code of it?
> 
> 
> Thanks
> 

It's a bit of off topic that you are doing right now...

[...]

PR #2348, opened 2020-12-02
Disable Invidious engine by default

mrwormo:
## What does this PR do?

Disable Invidious engine

https://invidio.us has closed down since September 1 2020. While waiting for another solution, invidious engine should be disabled to no longer have an error message in the search results page.

Make run

## Related issues

#2161

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -480,6 +480,7 @@ engines:
     base_url : 'https://invidio.us/'
     shortcut: iv
     timeout : 5.0
+    disabled : True
 
   - name: kickass
     engine : kickass


unixfox (approved):

return42:
Not need to set disabled=True when the default is already *disabled*, compare https://searx.github.io/searx/admin/engines.html

mrwormo:
> Not need to set disabled=True when the default is already _disabled_, compare https://searx.github.io/searx/admin/engines.html

Add `disable: True` in settings.yml is the only thing which work for me for disable Invidious engine by default.

return42:
> only thing which work for me for disable Invidious engine by default.

Usually this is not needed, I guess you had some settings stored in preferences.

PR #2357, opened 2020-12-06
Invidious engine

mrwormo:
## What does this PR do?

Changes in invidious.py to pick an instance url in a list instead of trying to use invidio.us

To have a working invidious search engine

Make run

## Related issues
#2161


unixfox:
Your commit history is quite messy, could you please fix it?

Also did you test if this won't break the case where someone set base_url to a single URL?

mrwormo:
> Your commit history is quite messy, could you please fix it?

Sorry for the mess, i'm quite newbie with git workflow... May i ask you how i can clean commit history ?

> Also did you test if this won't break the case where someone set base_url to a single URL?

Yes, it works with a single url. But i used `random.choice()` method which not pass the tests. Maybe i should modify the code to use `random.randint()`. I tried and it works.

unixfox:
> > Your commit history is quite messy, could you please fix it?
> 
> Sorry for the mess, i'm quite newbie with git workflow... May i ask you how i can clean commit history ?

Try to clone your forked git repository, switch to the invidious-engine branch and then merge all the commits into one then force push.
Or clone your forked git repository then reset the invidious-engine to the same state as the master branch and then force push.

searx/engines/invidious.py
```diff
@@ -19,8 +20,8 @@
 time_range_support = True
 
 # search-url
-base_url = "https://invidio.us/"
-
+base_url = ["invidious.tube", "invidious.snopyta.org", "invidiou.site"]
+base_url_rand = random.choice(base_url)
```
dalf: As it is written, an instance is picked randomly when searx starts, the other instances are ignored as long as searx is not restarted.

If `base_url` is a string, `base_url_rand` will contains `h` (the first character of `base_url` which most probably start with `https://` ).

mrwormo:
I'm really sorry for the mess, i guess it should be better to close this PR and I go back to working on the code to propose something more successful...

unixfox:
> I'm really sorry for the mess, i guess it should be better to close this PR and I go back to working on the code to propose something more successful...

You can just let this PR open, mark it as a draft and then push your new changes. When you think you are done undraft it and then notify us :).

searx/engines/invidious.py
mrwormo: > As it is written, an instance is picked randomly when searx starts, the other instances are ignored as long as searx is not restarted.

Yeah, i didn't realize this, i though it will pick a different instance for each request...

> If `base_url` is a string, `base_url_rand` will contains `h` (the first character of `base_url` which most probably start with `https://` ).

What do you mean by `h` ?

mrwormo:
> You can just let this PR open, mark it as a draft and then push your new changes. When you think you are done undraft it and then notify us :).

Ok, i did it, now i have to unwind this thing...

searx/engines/invidious.py
dalf: > What do you mean by h ?

Hum, actually, not `h` for sure.

But try this:
```python
base_url = "invidious.tube"
base_url_rand = random.choice(base_url)
```

You can see that `base_url_rand` contains only one character, because `random.choice(base_url)` sees `base_url` as an iterator of character : 

```python
for i in  ["invidious.tube", "invidious.snopyta.org", "invidiou.site"]:
  print(i)
for i in "invidious.tube":
  print(i)
```

#### Why we can `base_url = "invidious.tube"` ?

Because we can have this settings.yml ([EDIT] one admin will want to set one invidious instance instead of many, moreover some admin won't edit settings.yml after the update of invidious.py): 
```yaml
  - name : invidious
    engine : invidious
    base_url : 'https://invidious.tube/'
    shortcut: iv
    timeout : 5.0
    disabled : True
```

#### The solution I see:
* remove `base_url = ["invidious.tube", "invidious.snopyta.org", "invidiou.site"]`
* remove line 24 `base_url_rand = random.choice(base_url)`.
* before line 34 ( `    search_url = "https://" + base_url_rand + "/api/v1/search?q={query}"` ):
```
if isinstance(base_url, list):
  base_url_rand = random.choice(base_url)
else:
  base_url_rand = base_url
```
* change settings.yml this way:
```yaml
  - name : invidious
    engine : invidious
    base_url : 
        - https://invidious.tube
        - https://invidious.snopyta.org
        - https://invidiou.site
    shortcut: iv
    timeout : 5.0
    disabled : True
```

searx uses a different instance on each request. The invidious instances are listed in settings.yml instead of the engine code.

So:
* the admin can send request to .onion ( `base_url: http://wgegijigj.onion` )
*  the admin can optionally do this:
```yaml
  - name : invidious
    engine : invidious
    base_url : https://invidious.tube
    shortcut: iv
    timeout : 5.0
    disabled : True
```

mrwormo: Great ! Thx @dalf i'll take a look soon and follow your advice.
Many thanks, again !

mrwormo: I’ve tried what you propose @dalf, but it doesn't work. 
There is no error in the result page neither in terminal. I just have no results whatever I’m looking for, with the "Sorry, there is no result..." message.

I tried to remove "https://" from `search_url = "https://" + base_url_rand + "/api/v1/search?q={query}"` because the list of url in settings.yml already contains https://, for the same result.

How can i see what search url is used ?

By the way, there is another problem with Invidious in general: even if an instance seems to be alive, the search function on that instance does not return any results...

searx/engines/invidious.py
```diff
@@ -19,8 +20,8 @@
 time_range_support = True
 
 # search-url
-base_url = "https://invidio.us/"
-
+# base_url = ["invidious.tube", "invidious.snopyta.org", "invidiou.site"]
+# base_url_rand = random.choice(base_url)
```
dalf: Can you remove these two lines ?

searx/engines/invidious.py
```diff
@@ -30,7 +31,14 @@ def request(query, params):
...
     }
-    search_url = base_url + "api/v1/search?q={query}"
+
+    if isinstance(base_url, list):
+        base_url_rand = random.choice(base_url)
+    else:
+        base_url_rand = base_url
+
+    search_url = "https://" + base_url_rand + "/api/v1/search?q={query}"
+    print(search_url)
```
dalf: can you remove this line ?

dalf:
Here something that should fix your branch.
Be careful, it will delete / recreate your branch.


First, replace line 23 add 
```python
   # pylint: disable=E0602
```

so, pylint will be okay, so
```sh
make test
```
should okay.

Then, if it works:
```sh
git checkout invidious-engine
# store your change into invidious.patch
git diff f3524455ce1b9ecb55e233b367c6c4792d87eb48~  searx/engines/invidious.py  searx/settings.yml > invidious.patch
# check invidious.patch content
cat invidious.patch
git checkout master
# update your local master branch without the one on github
git pull origin master
# delete your branch (no worries for now: there is a backup on github)
git branch -D invidious-engine
# create the invidious-engine branch again
git checkout -b invidious-engine
# apply your modifications (without all the changes).
git apply invidious.patch
# please make sure searx/settings.yml changes are only about invidious
# check everything is working as expected
make test
# double check the invidious engine is working as expected going to http://127.0.0.1:8888
make run
# add the two modified files to the incoming commit.
git add searx/engines/invidious.py  searx/settings.yml
# show modifications
git diff --cached 
# commit
git commit
# push the commit to github
# WARNING: once done, the previous branch doesn't exist anymore on github.
git push --force --set-origin mrwormo invidious-engine
```

later to update your branch with the changes on master:
```sh
git pull --rebase origin master
git push --force --set-origin mrwormo invidious-engine
```

[EDIT] updated to create a patch for searx/engines/invidious.py  searx/settings.yml instead of searx/engines/qwant.py and searx/engines/command.py.

mrwormo:
Thx for you advice @dalf.

mrwormo:
Hello, nothing is working. I would prefer to remove this PR to start from a healthy base if you don't mind.

dalf:
So, I close this PR. Feel free to open another PR.

PR #2363, opened 2020-12-07
[fix] wikipedia: minor fix: return no result instead of crash in some very few cases.

dalf:
## What does this PR do?

In few cases, the JSON results doesn't contains the key 'type'.

See line:
https://github.com/searx/searx/blob/a458451d20bf45baf62d79a1e4ea2151e176f1d4/searx/engines/wikipedia.py#L55

Return no result instead of crashing.

Detected using `/stats/errors` on public searx instances.

I don't know how to reproduce the bug, nevertheless, the change is small enough to be reviewed.

### DIFF
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -52,7 +52,7 @@ def response(resp):
     api_result = loads(resp.text)
 
     # skip disambiguation pages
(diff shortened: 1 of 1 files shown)


return42 (approved):
LGTM / thanks!

PR #2365, opened 2020-12-07
[fix] soundclound: accept result without content

dalf:
## What does this PR do?

Sometimes, the `result['description']` is `None`: 
https://github.com/searx/searx/blob/a458451d20bf45baf62d79a1e4ea2151e176f1d4/searx/engines/soundcloud.py#L94

The result is still perfectly valid (there is actually no description).

Without this PR, some results are missing.

## Author's checklist

 check for `!soundcloud time` with and without this PR.

### DIFF
--- a/searx/engines/soundcloud.py
+++ b/searx/engines/soundcloud.py
@@ -91,7 +91,7 @@ def response(resp):
     for result in search_res.get('collection', []):
         if result['kind'] in ('track', 'playlist'):
             title = result['title']
-            content = result['description']
+            content = result['description'] or ''
             publishedDate = parser.parse(result['last_modified'])
             uri = quote_plus(result['uri'])
             embedded = embedded_url.format(uri=uri)


return42 (approved):
> check for `!soundcloud time` with and without this PR.

Can't see a change in my results, but the patch makes sense to me.

dalf:
> Can't see a change in my results, but the patch makes sense to me.

Also, in the console I see messages like this one:
`DEBUG:searx:result: invalid content: {'url': 'https://soundcloud.com/skindred/sets/roots-rock-riot', 'title': 'Roots Rock Riot', 'publishedDate': datetime.datetime(2020, 2, 5, 2, 32, 56, tzinfo=tzutc()), 'embedded': '<iframe width="100%" height="166" scrolling="no" frameborder="no" data-src="https://w.soundcloud.com/player/?url=https%3A%2F%2Fapi.soundcloud.com%2Fplaylists%2F981586039"></iframe>', 'content': None, 'engine': 'soundcloud'}`

(here with `!soundcloud rock`)

return42:
> Also, in the console I see messages like this one: ...

Thanks! I can confirm that there are no more `'content': None,` DEBUG messages.

dalf:
Thank you for the review.

PR #2372, opened 2020-12-09
[remove] remove searchcode_doc and twitter

dalf:
## What does this PR do?

Remove two engines:
* twitter: the API has changed. the engine needs to rewritten.
* searchcode_doc: the API about documentation doesn't exist anymore. See https://searchcode.com/api/

About twitter: most probably it is possible to rewrite a twitter engine using the same public API than nitter. Another solution is to add an engine for nitter instead of twitter:
* the nitter API is probably more stable than the twitter API
* nitter deals with the twitter API change.
* on the downside, the setup is more complicated (require to pick an nitter instance or to install nitter).

Currently:
* twitter never a result.
* searchcode_doc return an HTTP error to the user.


return42 (approved):
Lets merge.

PR #2373, opened 2020-12-09
Display if an engine does not support HTTPS requests

### LINKED ISSUE #302: Display which engines don't use HTTPS
The connection between the browser and searx can be encrypted, this visible on the browser.
But the connection between searx and a search engine can be in clear text without encryption. When it's possible searx uses https, but some search engines don't support HTTPS : 
- invalid certificate (not self signed certificate)
- not HTTPS support

In other words, anyone on the wire can intercept the request.

Ideas : 
- Display which search engines don't use HTTPS
- Have an option to disable these engines (a check box, something simple from the user point of view).

kvch:
## What does this PR do?

From now on engines have a new attribute named `https_support`. The engine developer can set it to either `True` or `False` whether the engine is queried using HTTPS. It only has to be set if the `request` function call would take to long. Otherwise, searx sets the attribute automatically.

Now, this information is displayed in the engines list on the Preferences page. Given that we only have 2-4 engines which do not support HTTPS, it might not be the biggest change. :D

## Related issues

Closes #302

### DIFF
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -1,4 +1,4 @@
-{% from 'simple/macros.html' import tabs_open, tabs_close, tab_header, tab_footer, checkbox_onoff, checkbox %}
+{% from 'simple/macros.html' import icon, tabs_open, tabs_close, tab_header, tab_footer, checkbox_onoff, checkbox %}
 
 {% extends "simple/base.html" %}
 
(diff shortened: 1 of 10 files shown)


searx/templates/simple/preferences.html
```diff
@@ -121,7 +121,7 @@ <h2>{{ _('Preferences') }}</h2>
       {% set engine_id = 'engine_' + search_engine.name|replace(' ', '_') + '__' + categ|replace(' ', '_') %}
       <tr>
         <td class="engine_checkbox">{{ checkbox_onoff(engine_id, (search_engine.name, categ) in disabled_engines) }}</td>
-        <th class="name">{{ search_engine.name }}</th>
+        <th class="name">{% if not search_engine.https_support %}{{ icon('exclamation-sign', 'No HTTPS') }}{% endif %} {{ search_engine.name }}</th>
```
dalf: It works with `warning` instead of `exclamation-sign`.

See https://github.com/searx/searx/blob/42a194898baf96e15edeb9268fbec5ecd4423c09/searx/static/themes/simple/gruntfile.js#L134

dalf (commented):
Expect the minor UI issue with the simple theme, LGTM.

searx/templates/simple/preferences.html
kvch: Thanks.

dalf (approved):

Issue #2369, opened 2020-12-08
./manage.sh styles Failed!


jatojoseph:
![searx-error](https://user-images.githubusercontent.com/29528121/101491589-37c37000-3964-11eb-9add-c0d8207bbb52.png)

Hi all,
I have done a few edits in my templates and wish to update the changes but `./manage.sh styles ` seems not to be working, also, `./manage.sh tests` is not working for PRs.
Lastly, how can I remove some of the categories under "Advanced Search"?
Any help pls.?

_Originally posted by @jatojoseph in https://github.com/searx/searx/issues/791#issuecomment-740631608_

dalf:
The documentation needs an update [here](https://searx.github.io/searx/dev/quickstart.html#how-to-compile-styles-and-javascript)

Use ```make themes``` to build the themes.

And ```make test```

Otherwise you can use grunt.

dalf:
I close the issue, feel free to re-open / add a comment.

PR #2380, opened 2020-12-12
[doc] update developer quickstart guide

return42:
See https://github.com/searx/searx/issues/2369#issuecomment-740652878


dalf:
Thank you !

PR #2384, opened 2020-12-13
Improve doc of the installation skripts

return42:
This is a small series to improve the documentation of the installation procedure. For details have a look at the commit messages.


pierrechtux (approved):
In the file docs/admin/installation.rst , instead of:
    "If you install from ``root``"
, I would rather put the following:
    "If you install as the ``root`` user"
, as to clarify the ambiguity between installing from the root directory of a machine, or installing as the root user.

return42:
@pierrechtux sorry, I was to late, PR was merged in the meantime.

> "If you install as the root user"
> , as to clarify the ambiguity between installing from the root directory of a machine, or installing as the root user.

Do you like to send a PR?

PR #2386, opened 2020-12-13
[fix] ERROR:searx:server.secret_key is not changed

### LINKED ISSUE #2278: Makefile target 'make doc' is broken
PR #2256 breaks the docs build

````
$ make docs
build searx/brand.py
build utils/brand.env
PYENV     install .
SPHINX    docs --> file:///800GBPCIex4/share/searx/dist/docs
Running Sphinx v3.2.1

Configuration error:
The configuration file (or one of the modules it imports) called sys.exit()
make: *** [Makefile:84: docs] Error 2
````

FWIW: a workaround is to use debug mode:

      SEARX_DEBUG=1 make docs-clean docs-live

### PR DESCRIPTION
**Pull request by return42**
## What does this PR do?

fix #2256

## Related issues

Closes #2278

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ run:  buildenv pyenvinstall
 	sleep 2 ; \
 	xdg-open http://127.0.0.1:8888/ ; \
 	) &
-	SEARX_DEBUG=1 $(PY_ENV)/bin/python ./searx/webapp.py
+	$(PY_ENV)/bin/python ./searx/webapp.py
 
 # docs
 # ----
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -60,7 +60,3 @@ if 'SEARX_SECRET' in environ:
     settings['server']['secret_key'] = environ['SEARX_SECRET']
 if 'SEARX_BIND_ADDRESS' in environ:
     settings['server']['bind_address'] = environ['SEARX_BIND_ADDRESS']
-
-if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
-    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
-    exit(1)
--- a/searx/settings_loader.py
+++ b/searx/settings_loader.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
-import secrets
+from os import urandom
 
 from os import environ
 from os.path import dirname, join, abspath, isfile
@@ -144,6 +144,6 @@ def load_settings(load_user_setttings=True):
             )
 
     if settings['server'].get('secret_key', 'ultrasecretkey') == 'ultrasecretkey':
-        settings['server']['ultrasecretkey'] = secrets.token_hex(16)
+        settings['server']['ultrasecretkey'] = urandom(16).hex()
 
     return settings, settings_load_message
--- a/searx/settings_loader.py
+++ b/searx/settings_loader.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+import secrets
+
 from os import environ
 from os.path import dirname, join, abspath, isfile
 from collections.abc import Mapping
@@ -112,21 +114,36 @@ def is_use_default_settings(user_settings):
 def load_settings(load_user_setttings=True):
     default_settings_path = get_default_settings_path()
     user_settings_path = get_user_settings_path()
+    settings, settings_load_message = None, None
+
     if user_settings_path is None or not load_user_setttings:
         # no user settings
-        return (load_yaml(default_settings_path),
-                'load the default settings from {}'.format(default_settings_path))
+        load_user_setttings = False
+        settings, settings_load_message = (
+            load_yaml(default_settings_path),
+            'load the default settings from {}'.format(default_settings_path)
+        )
+    else:
+        load_user_setttings = True
+        # user settings
+        user_settings = load_yaml(user_settings_path)
+        if is_use_default_settings(user_settings):
+            # the user settings are merged with the default configuration
+            default_settings = load_yaml(default_settings_path)
+            update_settings(default_settings, user_settings)
+            settings, settings_load_message = (
+                default_settings,
+                'merge the default settings ( {} ) and the user setttings ( {} )'.format(
+                    default_settings_path, user_settings_path)
+            )
+        else:
+            # the user settings, fully replace the default configuration
+            settings, settings_load_message = (
+                user_settings,
+                'load the user settings from {}'.format(user_settings_path)
+            )
 
-    # user settings
-    user_settings = load_yaml(user_settings_path)
-    if is_use_default_settings(user_settings):
-        # the user settings are merged with the default configuration
-        default_settings = load_yaml(default_settings_path)
-        update_settings(default_settings, user_settings)
-        return (default_settings,
-                'merge the default settings ( {} ) and the user setttings ( {} )'
-                .format(default_settings_path, user_settings_path))
+    if settings['server'].get('secret_key', 'ultrasecretkey') == 'ultrasecretkey':
+        settings['server']['ultrasecretkey'] = secrets.token_hex(16)
 
-    # the user settings, fully replace the default configuration
-    return (user_settings,
-            'load the user settings from {}'.format(user_settings_path))
+    return settings, settings_load_message
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,2 +1,0 @@
-import os
-os.environ['SEARX_DEBUG'] = '1'
--- a/utils/standalone_searx.py
+++ b/utils/standalone_searx.py
@@ -15,7 +15,7 @@ Example to use this script:
 
 .. code::  bash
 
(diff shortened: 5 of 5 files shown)

### CODE REVIEW

**Comment by dalf** [comment]
### When do searx use a secret_key ?

The secret_key is not useful for:
* utils/standalone.py
* make test
* make docs

In the master branch, the only use case of secret_key is the webapp.

| Debug/secret_key | Set | Unset |
|-----------|-----|-------|
| True      | Ok  | Ok    |
| False     | Ok  | KO    |


### Why do we need a secret_key ?

According to https://flask.palletsprojects.com/en/1.1.x/config/#SECRET_KEY : 


> A secret key that will be used for securely signing the session cookie and can be used for any other security related needs by extensions or your application.

According to https://github.com/pallets/flask/search?l=Python&q=secret_key , the only use case of secret_key is the flask.session.

searx don't use flask.session on purpose.

According to https://github.com/searx/searx/search?q=secret_key , the only searx usage of secret_key is the image proxy (the image proxy written in Python, not Morty).


### Sum up

1. this line is useless:
https://github.com/searx/searx/blob/36600118fb3e40ebd172245941306943581ef6f8/searx/webapp.py#L117

2. when the key is not set, the Python image proxy doesn't return an unique URL for the searx instance.
   said differently:
   when the key is not set, someone can use a searx an open proxy (at least for the images).

### About this PR

I'm not comfortable adding another step to configure a searx instance ( make sure INSTANCE_ENV='production' ).

### A different solution

move these lines
https://github.com/searx/searx/blob/36600118fb3e40ebd172245941306943581ef6f8/searx/__init__.py#L64-L66

to the run function in webapp.py (just before the logger.debug) :

https://github.com/searx/searx/blob/36600118fb3e40ebd172245941306943581ef6f8/searx/webapp.py#L1090-L1098

The test_webapp still require this line:
https://github.com/searx/searx/blob/36600118fb3e40ebd172245941306943581ef6f8/tests/unit/test_webapp.py#L14

but utils/standalone.py, make test, make docs will be okay.

### A third solution

If the key is None, empty string or 'ultrasecretkey' then searx creates a secure random key.

The key will change after each restart, so the image proxy URL won't be reliable over time.

(I prefer the previous solution).

**Comment by return42** [comment]
@dalf: thanks a lot for your detailed analyses , I didn't know this details before.

I moved the condition to webapp.py .. and renamed this PR.

**Comment by dalf** [comment]
May I ask you to remove `SEARX_DEBUG=1` in this line ?
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/.github/workflows/integration.yml#L69

since it is not more mandatory with this PR.

[EDIT] And there are the tests too : 
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/tests/__init__.py#L2
But here, this test will fail (perhaps another ?):
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/tests/unit/test_webapp.py#L15

A quick solution is not to change the tests for now.

**Comment by return42** [comment]
> May I ask you to remove SEARX_DEBUG=1 in this line ?

good point, I removed all the obsolete SEARX_DEBUG ..

> A quick solution is not to change the tests for now.

for me the test works .. may you like to verify.

**Comment by dalf** [comment]
> for me the test works .. may you like to verify.

This was puzzling me, actually the PR doesn't work: 
* uwsgi call `searx.webapp.app` , the `searx.webapp.run` is never call, see https://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html#the-first-wsgi-application and https://github.com/searx/searx/blob/14c7cc0e118f1d0873b32b34793cdec2c5c9c13e/searx/webapp.py#L1138 
* same about the test 

So the `run` check for the SECRET_KEY when searx is started using `python searx/webapp.py `, most probably in debug mode.

The actual check must be done on wsgi application.

Flask doesn't provide callbacks when the app starts, only on the first request: https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.before_first_request

So the only solution I see is a wsgi middleware, that doesn't nothing except check for the SECRET_KEY when initialize.

**Comment by return42** [comment]
> So the only solution I see is a wsgi middleware, that doesn't nothing except check for the SECRET_KEY when initialize.

Or we bury the idea to test SECRET_KEY,thoughts?  Since #2256 we have a lot of trouble and the only benefit is to catch up sloppy admins.  

> said differently: when the key is not set, someone can use a searx an open proxy (at least for the images).

Admins have to read the documentation, how to [configure searx](https://searx.github.io/searx/admin/installation-searx.html#configuration)

**Comment by dalf** [comment]
The other solution is to create a random key each time: at least it is secure even for a sloppy admin: no open proxy

**Comment by return42** [comment]
>  to create a random key each time:

I will have a look ... WIP

**Comment by dalf** [comment]
Perhaps there is a better version, just for reference:

https://github.com/dalf/searx/blob/1cfe7f2a7543b2994a1afd0d81da1962d04423b0/utils/update_user_settings.py#L19-L24

**Comment by return42** [comment]
@dalf: key is now generated in the settings_loader

https://github.com/searx/searx/blob/f65a4dc4ca5af372ceab84d121747616ee351a26/searx/settings_loader.py#L146-L147

**Comment by kvch** [comment]
@dalf Generating a secret key every time searx starts up is not viable if image proxy is enabled. Once the instance is restarted, it gets a new secret key. Thus, previously working links image result links break on the instance because the secret key is changed. I prefer the initial solution I have submitted. (To me, it is somewhat weird that the documentation requires starting the web application.)

**Comment by return42** [comment]
> To me, it is somewhat weird that the documentation requires starting the web application.

The application is not started, but searx will be imported.  The documentation build itself needs a some values from the searx package like version number ..

https://github.com/searx/searx/blob/59217bb5bed86926b2bbd84f406ed022b0762c6e/docs/conf.py#L5

You can test it by :

```
$ . ./local/py3/bin/activate
(py3) (0) user@c340:/tmp/searx [master|+ 2] 
19:57 $ python
...
>>> import searx
```

see https://github.com/searx/searx/pull/2256/files#diff-5b9df979b3e9792a21e5b0e3786eb590c34fd036a5fce4f79150a5fa0261cbf4

**Comment by return42** [comment]
> Generating a secret key every time searx starts up is not viable ..

key is only generated when it equals to `ultrasecretkey` [see](https://github.com/searx/searx/pull/2386#issuecomment-748607952)

**Comment by dalf** [comment]
@kvch currently `import searx` trigger the secret_key value check.

@kvch  @return42 

Here a patch so the documentation doesn't `import searx.webapp`.

[LAST EDIT] is it still weird written like this ?

It requires some adjustments, at least: 
  * the jinja context is still `webapp`
  * most probably, `if not searx_debug and settings....` should be at the top of the file

```diff
diff --git a/docs/conf.py b/docs/conf.py
index d6fde9be..5f0ab02b 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,9 +27,15 @@ numfig = True
 
 exclude_patterns = ['build-templates/*.rst']
 
-from searx import webapp
+import searx.search
+import searx.engines
+import searx.plugins
+searx.search.initialize()
 jinja_contexts = {
-    'webapp': dict(**webapp.__dict__),
+    'webapp': {
+        'engines': searx.engines.engines,
+        'plugins': searx.plugins.plugins
+    },
 }
 
 # usage::   lorem :patch:`f373169` ipsum
diff --git a/searx/__init__.py b/searx/__init__.py
index 9bbc7c8c..579fcf66 100644
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -61,6 +61,3 @@ if 'SEARX_SECRET' in environ:
 if 'SEARX_BIND_ADDRESS' in environ:
     settings['server']['bind_address'] = environ['SEARX_BIND_ADDRESS']
 
-if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
-    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
-    exit(1)
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1141,3 +1141,7 @@ app.wsgi_app = ReverseProxyPathFix(ProxyFix(application.wsgi_app))
 
 if __name__ == "__main__":
     run()
+
+if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
+    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
+    exit(1)
```

So this fix `make docs` / `make docs-live`. Not the tests (if it makes sense):
https://github.com/searx/searx/blob/59217bb5bed86926b2bbd84f406ed022b0762c6e/tests/__init__.py#L2

My idea about the generated key is to avoid an open proxy, and yes it breaks the links, but only for misconfigured instance.

return42:
What is the problem with this PR?  Every time I implement the solution other solutions are preferred.

If this PR does not match the needs and someone knows a better solution close it

dalf:
Note: the patch just above is a different thing, basically it shows the documentation doesn't depend on the webapp only some other components.

return42:
The  patch just above is totally OK for me, but it is not mine solution. Lets close this PR and send a PR with your solution.

BTW: the main problem is loading setup stuff and decide to exit the whole execution **at import time**.  
The article "[Don’t run code at import time](https://www.benkuhn.net/importtime/)" lists some of the drawbacks with execution at import time. 

Your patch reduces the problem to the webapp, so I would prefer your solution.

PR #2391, opened 2020-12-15
Add link to contact instance maintainer to footer of each page

### LINKED ISSUE #1972: SUGGESTION: Contacting the instance's maintainer(s)
Hello, so I use searx, but I personally think that there should be any way to contact the maintainer(s) of a public instance (email for example). It is harder to trust this awesome service if there is no way to contact the maintainer(s).

BBaoVanC:
## What does this PR do?

Adds a `contact_url` setting to the general section of settings.yml. If set, a hyperlink will be added to the footer (right after the `Public Instances` link) which links to the set url.

This makes it easier for users to report questions, issues, etc to the maintainer of the instance.

Add a key named `contact_url` under the `general` section of settings.yml, with the value being a url (example included in the pull request).

## Related issues

Closes #1972


searx/templates/oscar/base.html
```diff
@@ -89,6 +89,10 @@
                     <a href="{{ brand.GIT_URL }}">{{ _('Source code') }}</a> |
                     <a href="{{ brand.ISSUE_URL }}">{{ _('Issue tracker') }}</a> |
                     <a href="{{ brand.PUBLIC_INSTANCES }}">{{ _('Public instances') }}</a>
+                    {% if contact_url %}
+                    |
+                    <a href="{{ contact_url }}">{{ _('Contact Instance Maintainer') }}</a>
```
unixfox: Why the uppercase for the words "instance" and "maintainer"?

searx/templates/simple/base.html
```diff
@@ -55,6 +55,10 @@
         <a href="{{ brand.GIT_URL }}">{{ _('Source code') }}</a> |
         <a href="{{ brand.ISSUE_URL }}">{{ _('Issue tracker') }}</a> |
         <a href="{{ brand.PUBLIC_INSTANCES }}">{{ _('Public instances') }}</a>
+        {% if contact_url %}
+        |
+        <a href="{{ contact_url }}">{{ _('Contact Instance Maintainer') }}</a>
```
unixfox: Same here

unixfox (commented):
Apart from the uppercase everything is working correctly on my side when testing this PR.

return42:
Does it make sense to add `contact_url` to the *branding*. like we have done for `PUBLIC_INSTANCES`

https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/Makefile#L120-L126

If not, `contatc_url` should also be added to the 

https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/utils/templates/etc/searx/use_default_settings.yml#L3-L5

searx/templates/oscar/base.html
return42: See comment: https://github.com/searx/searx/pull/2391#issuecomment-745955309

`<a href="{{ brand.CONTACT_URL }}">{{ _('Contact Instance Maintainer') }}</a>`

searx/webapp.py
```diff
@@ -413,6 +413,8 @@ def render(template_name, override_theme=None, **kwargs):
 
     kwargs['brand'] = brand
 
+    kwargs['contact_url'] = settings.get('general').get('contact_url')
```
return42: If we ship CONTACT_URL within the `brand`  namespace, this line can be dropped.

BBaoVanC:
> Does it make sense to add `contact_url` to the _branding_. like we have done for `PUBLIC_INSTANCES`
> 
> https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/Makefile#L120-L126
> 
> If not, `contatc_url` should also be added to the
> 
> https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/utils/templates/etc/searx/use_default_settings.yml#L3-L5

Hmm, that actually looks like a good idea.

searx/templates/simple/base.html
BBaoVanC: Fixed in 9a2be39

searx/templates/oscar/base.html
BBaoVanC: 9a2be39

searx/webapp.py
BBaoVanC: Ok, I made 65af3f3 now.

searx/settings.yml
```diff
@@ -1,6 +1,7 @@
 general:
     debug : False # Debug mode, only for development
     instance_name : "searx" # displayed name
+    # contact_url : "mailto:contact@example.com" # contact link on footer
```
return42: We can use only one type of *setting*.  If something is added to the *brand*, it is no longer a setting in the `settings.yml`.  Anyway, the rest of this PR looks good to me.

BBaoVanC: Ok, fixed it in c88b067f.

return42 (dismissed):
LGTM / thanks! .. to achieve a plus point :) .. do you like to squash the patches into one?

BBaoVanC:
> LGTM / thanks! .. to achieve a plus point :) .. do you like to squash the patches into one?

I'm not sure, I haven't really done much squashing before.

return42:
> I'm not sure, I haven't really done much squashing before.

No problem!

BBaoVanC:
> > I'm not sure, I haven't really done much squashing before.
> 
> No problem!

I'd be happy to squash it if I can make it not cause issues with the master branch of my searx fork, since I merged it into my master (twice, once before moving to brand.py, and once after).

BBaoVanC:
Just rebased it, I think it might be ready to merge.

return42 (approved):
LGTM / thanks!

dalf:
It is related to https://github.com/searx/searx/issues/2394 and IMO this contact info should be in settings.yml (also see https://github.com/searx/searx-instances/issues/45 ).

Two notes:
* About the squashed commits : it is also important to keep the searx github history as clean as we can.
* As it is written CONTACT_URL is mandatory and it has a default value `contact@example.com`. It is will misleading for the users, the link `Contact instance maintainer` can be a dead-end if the maintainer don't  replace the default value (I'm nearly sure it will happen often). It would be better to keep ` {% if contact_url %}`)

But I'm sorry for the delay here, I'm going to merge this PR, and make CONTACT_URL optional.

PR #2396, opened 2020-12-16
Fix the StartPage result title is showing the url

### LINKED ISSUE #2395: Startpage: The title result is showing the url
CleanShot 2020-12-16 at 12 00 49
Just updated to the newest version. 0.18.0. The title in StartPage result is showing the url rather than the page title. Same happens to other public instance.

_Originally posted by @lucky13820 in https://github.com/searx/searx/pull/2385#issuecomment-746927618_

lucky13820:
## What does this PR do?

Change the link variable in the startpage.py, so it will contains both the url and the title. Then the result page can show the tile instead of the url in the title section. 

The StartPage will be useful again. 

## Related issues
Closes #2395


BBaoVanC:
This is working for me, merged it into my fork (and my instance).

dalf (approved):
Thank you!

PR #2399, opened 2020-12-18
[fix] determine path to makefile.lxc in a LXC

return42:
The path to the makefile.lxc is not always static and `/share/searx/utils/makefile.lxc` can be wrong.

When target `lxc.activate` is called within the LXC container the path to `makefile.lxc` depends on `LXC_SHARE_FOLDER` and the `basename` of the folder where searx was cloned initial on the HOST system.

This patch determines the path of `makefile.lxc` next to `makefile.include` itself.


dalf (approved):
Thank you !

PR #2405, opened 2020-12-20
[mod] remove emojis from source code

dalf:
## What does this PR do?

Remove emojis from the source code.

See https://github.com/searx/searx/discussions/2403

N/A


return42 (approved):
Thanks!!

Issue #2394, opened 2020-12-16
What searx option is a part of the brand and what is a part of instance settings.yml ?


dalf:
Currently:
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/searx/brand.py#L1-L6

* `SEARX_URL` is very similar to https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/searx/settings.yml#L16
* `GIT_URL` usage:
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/utils/makefile.sphinx#L180
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/utils/searx.sh#L398
* `GIT_BRANCH` usage: utils/searx.sh
https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/utils/searx.sh#L329
* `DOCS_URL` is displayed in the themes, referenced in `setup.py` and `utils/morty.sh`
* `PUBLIC_INSTANCES`,  `DOCS_URL`, `ISSUE_URL` are displayed in the themes and referenced in `setup.py`

----

When a setting should go into brand.py ? (brand.py but it is actually [Makefile](https://github.com/searx/searx/blob/002d53abf35192ba1094d7516da10b3442ca1f16/Makefile#L4-L9) )

return42:
> What is the difference between band.py and settings.yml 

Yes, this needs some discussion .. but I would use a different title simmilar:

What searx option is a part of the brand and what is a part of instance `settings.yml` ..

.. and and what are the intersections.

dalf:
Let's start with `SEARX_URL`:
* in https://searx.github.io/searx/admin/settings.html#global-settings ,  when the admin modifies `base_url`, should we ask to update `SEARX_URL` too ?
* or should we remove `base_url` from settings.yml ? If yes, why not the other settings in the `server` section ? 
  * Note: `utils/morty.sh` and `utils/searx.sh rely` on this value. So should searx.sh creates a settings.yml according to SEARX_URL ? In this case SEARX_URL is a setup parameter for the installation scripts.
* or should we remove `SEARX_URL` from brand ?

return42:
I think the _brand_ is needed for the _buildenv_ while _settings_ are needed for a instance.

Values like `SEARX_URL` and `CONTACT_URL` are needed at build-time and also are needed at instance's run-time.

[...]

PR #2408, opened 2020-12-20
[mod] move brand options from Makefile to settings.yml

return42:
## What does this PR do?

Move brand options from Makefile to settings.yml, see #2394:   _What searx option is a part of the brand and what is a part of instance settings.yml_ 

Options in settings.yml which are a part of the [brand](https://github.com/return42/searx/blob/1d9f26b1bbf937e8a0ae4f40b89a47d9607233de/searx/settings.yml#L4-L11)

create a brand to test .

```diff
diff --git a/Makefile b/Makefile
index a00f9427..2d1995b9 100644
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ export GIT_URL=https://github.com/searx/searx
 export GIT_BRANCH=master
 export DOCS_URL=https://searx.github.io/searx
 export SEARX_URL=https://searx.me
-#export CONTACT_URL=mailto:contact@example.com
+export CONTACT_URL=mailto:contact@example.com
 # END Makefile setup
 
 include utils/makefile.include
```

Then start

     make run

check _contact maintainer_ link at the bottom of searx's intro page.


dalf:
It will take me time to review this PR.

After  a first reading, it seems `brand.py` and `brand.env` can be temporary files.

But IMO, the first thing to answer is what is the purpose of each variable, which problem does it solve.
And only after that, see how and where to implement them.

return42:
> It will take me time to review this PR.

No problem, I do not have much time the next days .. 

> After a first reading, it seems brand.py and brand.env can be temporary files.

I also hoped so, but brand.py is used in setup.py .. and the shell scripts (in contrast to the he makefile targets) do not build the prerequisites before executed.

> But IMO, the first thing to answer is what is the purpose of each variable, which problem does it solve.

This Draft is all I could do yesterday. I wanted to have your assessment of it first. I will probably not have time again until January to deal with it more intensively. But I will read your comments. / Thanks!

dalf:
Currently, would it more simple to: 
* load settings.yml in setup.py (instead of using brand.py)
* read the required values from settings.yml instead of brand.py
?

for example write `settings['general']['contact_url']` instead of  `brand.CONTACT_URL` :
https://github.com/searx/searx/blob/59217bb5bed86926b2bbd84f406ed022b0762c6e/searx/templates/oscar/base.html#L92

return42:
> load settings.yml in setup.py (instead of using brand.py)

Good point, I will give it a try.  But first lets approve & merge #2412, #2420 and  #2421 to have a clean boilerplate.

return42:
@dalf: I implemented commit https://github.com/searx/searx/pull/2408/commits/fa0b560f4b9dba57ee727db5e4ea0dde14719cbc This patch is a first 'proof of concept'.  Later we can decide to remove the brand namespace entirely or not.

ATM in CI we have the problem that setup.py throws [the know exception and exits](https://github.com/searx/searx/pull/2408/checks?check_run_id=1613945057#step:6:91):

I will first have a view at your PR  #2411, before I continue here.

update: #2411 will fix the build issue from above

searx/__init__.py
```diff
@@ -64,3 +64,49 @@
 if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
     logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
     exit(1)
+
+
+class _brand_namespace:
```
dalf, line 65: Technically, this class shouldn't exist: everywhere where there is `brand.CONTACT_URL` it should be `settings['global'].get('contact_url', False)` ?

Or am I missing something ?

But it is convenient way to remove `searx/brand.py` with few changes.

manage.sh
```diff
@@ -123,9 +123,9 @@ docker_build() {
     SEARX_GIT_VERSION=$(git describe --match "v[0-9]*\.[0-9]*\.[0-9]*" HEAD 2>/dev/null | awk -F'-' '{OFS="-"; $1=substr($1, 2); if ($3) { $3=substr($3, 2); }  print}')
 
     # add the suffix "-dirty" if the repository has uncommited change
-    # /!\ HACK for searx/searx: ignore searx/brand.py and utils/brand.env
+    # /!\ HACK for searx/searx: ignore utils/brand.env
     git update-index -q --refresh
-    if [ ! -z "$(git diff-index --name-only HEAD -- | grep -v 'searx/brand.py' | grep -v 'utils/brand.env')" ]; then
```
dalf, line 128: The change makes sense but I'm confused by the existing code:
* this line is using the git command line.
* line 149, 155, 160 uses `GIT_URL`

Can we have a scenario where `GIT_URL` is different from `git remote get-url origin` ?

setup.py
```diff
@@ -8,9 +8,9 @@
 import sys
 
 # required to load VERSION_STRING constant
-sys.path.insert(0, './searx')
-from version import VERSION_STRING
-import brand
+sys.path.insert(0, './')
```
dalf: I think this line can be removed since : 

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter.

( https://docs.python.org/3/library/sys.html#sys.path )

return42: Thanks for the tip, I removed the line.

searx/__init__.py
return42: > Or am I missing something ?

no, you are correct.

> But it is convenient way to remove searx/brand.py with few changes.

This was my intention, we can remove the usage of brand-namepace afterwards.

manage.sh
return42: I guess you missunderstand the lines of code; this is "grep -v" to suppress changes in the brand-files when calling a git diff.

Did I answered your question?

return42:
@dalf: with the last review this series merges all settings into settings.yml and the brand.py file has been removed. There are many tasks to do, but as the first that should be enough.  

If you think this PR is ready to merge, we should merge.  For the other tasks I have planned I will send PRs afterwards.

manage.sh
dalf: Nope :-)
* Line 128 uses `git` so it assumes that the current directory is a git repository most probably clone from somewhere: `git remote get-url origin` gives the origin.
* Line 159 uses `git rev-parse HEAD` to get the git commit
* Line 160 uses `GIT_URL` to get the git URL.

According to line 159 and 160, either `GIT_URL` equals `git remote get-url origin`, either we can have an incoherence.

So my question is: when `GIT_URL` is different from `git remote get-url origin`. 

Anyway this change was made before this PR, so I close the conversation.

searx/settings.yml
```diff
@@ -1,6 +1,14 @@
 general:
     debug : False # Debug mode, only for development
     instance_name : "searx" # displayed name
+    git_url: https://github.com/searx/searx
+    git_branch: master
+    issue_url: https://github.com/searx/searx/issues
+    docs_url: https://searx.github.io/searx
+    public_instances: https://searx.space
+    contact_url: False # mailto:contact@example.com
```
dalf: Is is possible to:
* keep `contact_url` is the `general` section : this is a value most admin will probably modify.
* move the other new keys to a `brand` section just after the `general` section. These values won't be modified expect for a new searx brand.

utils/brand.env
```diff
@@ -1,7 +1,9 @@
+export SEARX_URL=''
```
dalf, line 1: Since this file is built each time it is required using `make`, is it required to commit this file ? or can we remove this file and add it to .gitignore ?

searx/settings.yml
return42: > Is is possible to:

Yes, I can move such options to new brand section.

utils/brand.env
return42: My aim is to remove utils/brand.env entirely .. but this will take some days, since I am back at my customers these days.

searx/__init__.py
```diff
@@ -60,3 +60,49 @@
     settings['server']['secret_key'] = environ['SEARX_SECRET']
 if 'SEARX_BIND_ADDRESS' in environ:
     settings['server']['bind_address'] = environ['SEARX_BIND_ADDRESS']
+
+
+class _brand_namespace:
+
+    @classmethod
+    def get_val(cls, group, name, default=''):
+        return settings[group].get(name, False) or ''
```
dalf: should be `return settings.get(group, {}).get(name) or default`:
* to use `default`.
* to be fail safe in `group` doesn't exist in the settings.

see commit https://github.com/dalf/searx/commit/1074ad778071ca5208932d16e909fcd8dc811777

utils/build_env.py
```diff
@@ -0,0 +1,30 @@
...
+"""build environment used by shell scripts
+"""
+
+# set path
+import sys
+from os.path import realpath, dirname, join, sep
+repo_root = realpath(dirname(realpath(__file__)) + sep + '..')
+sys.path.insert(0, repo_root)
+
+from searx import brand
```
dalf, line 19: Here `searx.brand` is based on `searx.settings`.

If `SEARX_SETTINGS_PATH` is set, then `utils/brand.env` won't have the expected values from `searx/settings.yml`.

see commit https://github.com/dalf/searx/commit/6faaf5e976c23bd03abf406990b1501b5463bdaa

utils/brand.env
dalf: Implemented in the branch https://github.com/dalf/searx/commits/fix-rm-brand-make (commit https://github.com/dalf/searx/commit/a669c8a4cc1b3e3056f4dd925d320154c208d3a9 )

feel free to cherry-pick

utils/build_env.py
return42: Hm, I had to think twice about .. but under the assumption that a brand is always a fork, you are right.  I will cherry pick / thanks!

dalf: Actually I had the issue because `make test.robot` was not working, [without log](https://github.com/searx/searx/blob/dbfd5567aaa06a5f9792bb0839af74877ed5258c/searx/testing.py#L49-L56), I started searx with `SEARX_SETTINGS_PATH=$(pwd)/searx/testing.yml`, then `make test` started to fail.

So in the end, without the issue in `_brand_namespace.get_val`, I wouldn't have seen it.

searx/settings.yml
return42: I will pick your 1074ad77 / thanks!

utils/brand.env
return42: Thanks for a669c8a4 but I think this will not be enough.  The reason brand.env exists is because it is sourced by the shell scripts:

    ./manage.sh:15:. "${BASE_DIR}/utils/brand.env"
    ./utils/filtron.sh:9:source "${REPO_ROOT}/utils/brand.env"
    ./utils/searx.sh:9:source "${REPO_ROOT}/utils/brand.env"
    ./utils/morty.sh:8:source "${REPO_ROOT}/utils/brand.env"

One solution might be to generate the file on demand like shown here:

```diff
diff --git a/utils/searx.sh b/utils/searx.sh
index b7d3b8e1..ff5f2dd6 100755
--- a/utils/searx.sh
+++ b/utils/searx.sh
@@ -6,6 +6,11 @@
 # shellcheck source=utils/lib.sh
 source "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
 # shellcheck source=utils/brand.env
+if [ ! -f "${REPO_ROOT}/utils/brand.env" ]; then
+    pushd "${REPO_ROOT}" > /dev/null
+    make buildenv
+    popd > /dev/null
+fi
 source "${REPO_ROOT}/utils/brand.env"
 source_dot_config
 source "${REPO_ROOT}/utils/lxc-searx.env"
```
The drawback of this solution is, that the `/utils/searx.sh` is called with sudo rights.
Calling `make buildenv` will generate a lot of files owned by root .. nothing we should do.

searx/__init__.py
return42: I cherry picked 1074ad77 / thanks!

dalf (approved):

Issue #2252, opened 2020-10-09
Add searx/settings.py ?


dalf:
Currently [`searx/__init__.py`](https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/__init__.py#L30-L51) reads `searx/settings.yml`  into `searx.settings`.

Then `searx.settings` is read directly in numerous places (not exhaustive) :
* most of https://github.com/searx/searx/blob/master/searx/__init__.py
* https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/webapp.py#L93
* https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/webapp.py#L115
* https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/preferences.py#L318-L375
* https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/utils.py#L484-L496
* https://github.com/searx/searx/blob/da8b227044f45127f705f6ea94a72d368eea73bb/searx/poolrequests.py#L70-L71
* https://github.com/searx/searx/blob/474d56c77f4c8bbd3d03f6feec0d2a13dcd41d40/searx/search.py#L38-L48

Wouldn't it be better to:
* add searx/settings.py
* clearly defined the default values here (for example a default settings written with dict and list in Python)
* write some helper to read the different values. Related to
      * https://github.com/searx/searx/pull/2251
      * https://github.com/searx/searx/pull/1278
* write a loader that check the validity of settings.yml
?

return42:
I vote for! 

> [nearly all application setup is spread around the project and is done while importing modules](https://github.com/searx/searx/pull/2096#issuecomment-663826089)

kvch:
I agree, it would be nice to have some mechanism to check settings.yml. I think it could be something similar to the current `preferences.py`. We could even reuse the `*Setting` classes implemented there.

If we add the ability to check the settings.yml, we could also add a subcommand or whatever to let instance admins check the configuration without starting searx.

asciimoo:
I don't fully agree. It's true, it would make settings handling better, but I think, there are much more important tasks to solve like engine fixes or other open bugs/enhancements. It generated quite a few issues over the years, so this is a low priority task for me.

> If we add the ability to check the settings.yml, we could also add a subcommand or whatever to let instance admins check the configuration without starting searx.

[...]

PR #2411, opened 2020-12-21
Update secret key check

dalf:
## What does this PR do?

See the commit messages for the details.

Second version of https://github.com/searx/searx/pull/2386

searx don't check for the secret_key with `utils/standalone.py` and `make docs`.
   
https://github.com/searx/searx/blob/59217bb5bed86926b2bbd84f406ed022b0762c6e/tests/__init__.py#L2
is still there

## Author's checklist

* `make test`
* `make run`
* `make docs` : check the engines and plugins pages.
* start the application using uwsgi.

## Related issues

* https://github.com/searx/searx/pull/2386
* https://github.com/searx/searx/issues/2252

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,9 +27,15 @@ numfig = True
 
 exclude_patterns = ['build-templates/*.rst']
 
-from searx import webapp
+import searx.search
+import searx.engines
+import searx.plugins
+searx.search.initialize()
 jinja_contexts = {
-    'webapp': dict(**webapp.__dict__),
+    'searx': {
+        'engines': searx.engines.engines,
+        'plugins': searx.plugins.plugins
+    },
 }
 
 # usage::   lorem :patch:`f373169` ipsum
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -86,6 +86,11 @@ from searx.metrology.error_recorder import errors_per_engines
 from werkzeug.serving import WSGIRequestHandler
 WSGIRequestHandler.protocol_version = "HTTP/{}".format(settings['server'].get('http_protocol_version', '1.0'))
 
+# check secret_key
+if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
+    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
+    exit(1)
+
 # about static
 static_path = get_resources_directory(searx_dir, 'static', settings['ui']['static_path'])
 logger.debug('static directory is %s', static_path)
--- a/docs/admin/engines.rst
+++ b/docs/admin/engines.rst
@@ -46,7 +46,7 @@ Show errors   **DE**
 
 .. _configured engines:
 
-.. jinja:: webapp
+.. jinja:: searx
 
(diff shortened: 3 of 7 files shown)


docs/conf.py
```diff
@@ -27,9 +27,15 @@
...
+import searx.search
+import searx.engines
+import searx.plugins
+searx.search.initialize()
 jinja_contexts = {
-    'webapp': dict(**webapp.__dict__),
+    'searx': {
+        'engines': searx.engines.engines,
+        'plugins': searx.plugins.plugins
+    },
```
return42, line 38: Much better than `webapp.__dict__` from mine / thanks for this improvement.

searx/webapp.py
```diff
@@ -1135,9 +1130,37 @@ def __call__(self, environ, start_response):
...
+
+
+# uwsgi application
 application = app
+
 # patch app to handle non root url-s behind proxy & wsgi
 app.wsgi_app = ReverseProxyPathFix(ProxyFix(application.wsgi_app))
 
+# initiliaze the webapp when it is started not loaded
+app.wsgi_app = InitializeApplication(app.wsgi_app)
```
return42: Just for my interest; is it typical design in  _middleware_ to chain? What might be the drawback (except that the class name is inappropriate) placing the implementation into the existing `ReverseProxyPathFix` class?

return42 (dismissed):
I really like this solution / thanks!!

searx/webapp.py
dalf: It should be the right way
https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.wsgi_app 

But anyway the line 1162 creates an instance of InitializeApplication, so the code in __init__ runs, so it doesn't provide anything. So I'm going to remove this class, and check at import time.

Thank you for the question.

---

BTW starlette offers an `on_startup` option: https://www.starlette.io/applications/

return42:
> So I'm going to remove this class, and check at import time.

No, please not .. Sorry, ATM I do not have much time to investigate, but might this article help?

https://networklore.com/start-task-with-flask/

Otherwise I vote to remove the check entirely, as long, as we do not have a solution which does not break every toolchain, we have (e.g. 5min ago I tried to run `make useragents.update`).  Please take into account that checking the key is a _"nice to have"_ and nothing what is mandatory if admins read the installation instruction first (which should the common case).

@kvch: what do you think?

dalf:
I'm afraid a solution like https://networklore.com/start-task-with-flask/ while breaks the initialization in a way or another (with / without uwsgi, debug / not debug, etc).

The target on the long run is starlette : https://github.com/searx/searx/pull/2313

So, despite the fact the check on searx.webapp import creates issues, I think it is still an improvement.

return42:
> I think it is still an improvement.

_d'accord_ ... I tested on top of #2408 and it will fix the [build issue](https://github.com/searx/searx/pull/2408/checks?check_run_id=1613945057#step:6:91) / details see my https://github.com/searx/searx/pull/2408#issuecomment-751470102

approved / thanks a lot!

PR #2418, opened 2020-12-24
[fix] makefile.python: remove duplicate pyenv-(un)install targets

return42:
## What does this PR do?

Makefile targets 'pyenv-install' and 'pyenv-uninstall' are unused duplicates of 'pyenvinstall' and 'pyenvuninstall'.

no test needed / PR drops unused code duplications

### DIFF
--- a/utils/makefile.python
+++ b/utils/makefile.python
@@ -220,16 +220,8 @@ PHONY += pydebug
 pydebug: $(PY_ENV)
 	DEBUG=$(DEBUG) $(PY_ENV_BIN)/pytest $(DEBUG) -v $(TEST_FOLDER)/$(TEST)
 
-# install / uninstall python objects into virtualenv (PYENV)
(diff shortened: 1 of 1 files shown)


dalf (approved):
Thank you !

Out of topic : 

`make` could be faster if `pyenvinstall` calls `cmd_pyenvuninstall` only when `requirements*.txt` are differents.

One way to do that:
* `./local/py3/requirements-sha.txt` contains the sha512 of the content of `requirements*.txt`
* in `pyenvinstall`:
    * calls `cmd_pyenvuninstall`
	* only if `./local/py3/requirements-sha.txt` doesn't match the `requirements*.txt`

return42:
> make could be faster if pyenvinstall calls cmd_pyenvuninstall only when requirements*.txt are differents.

That is one point on my ToDo list :)

Your suggestion is a good starting point / thanks for the tip.

PR #2420, opened 2020-12-24
fix makefile `books/{name}.*` targets and add documentation about these targets

return42:
## What does this PR do?

see commit messages

```
$ make books/user.html
...
books/user.html --> file:///share/searx/dist/books/user

$ make books/user.pdf
...
SPHINX    books/user.pdf --> file:///share/searx/dist/books/user/pdf
```


dalf (approved):
Thank you !

PR #2421, opened 2020-12-24
[perf] optimize creation of the virtualenv & pyenvinstal targets

return42:
## What does this PR do?

Nearly all makefile targets have virtualenv and pipenvinstall in their chain of prerequisite.  This patch series optimize creation of the virtualenv & pyenvinstal targets.  By this all targets based on these targets will speed up significantly.

The third patch fix the message when you call `make PY=3.7`

Make a first build of pyenv like usual ..

```
$ make clean pyenvinstall
Collecting pip
...
PYENV     install .
Obtaining file:///share/searx
...
Successfully installed  ...
```

If you build the same target again  ...

```
make pyenv
$ make pyenvinstall
PYENV     install .
PYENV     . already installed

```

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -1381,7 +1381,7 @@ in_container() {
 LXC_ENV_FOLDER=
 if in_container; then
     # shellcheck disable=SC2034
-    LXC_ENV_FOLDER="lxc/$(hostname)/"
+    LXC_ENV_FOLDER="lxc-env/$(hostname)/"
 fi
 
 lxc_init_container_env() {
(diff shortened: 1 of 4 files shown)


return42:
Added two patches more:

- bc7bd1d
- d175a0f 

both are a bit off topic in this PR. Since all patches in this series modify the _boilerplate_ of make and bash scripts, I hope it is OK to pass them within this PR.

dalf (approved):
Thank you !

dalf:
Another possible optimization: 
`cmd_pylint      = $(PY_ENV_BIN)/python -m pylint -j 4 --rcfile $(PYLINT_RC) $2`

`-j 4` makes sure that pylint uses at 4 CPU cores.

On my laptop, `make test` is 30 second faster.

`$(grep 'cpu cores' /proc/cpuinfo | uniq)` (or similar) is probably better than the hard coded value `4`

PR #2454, opened 2021-01-10
Fix empty colon in query from selecting Chinese

### LINKED ISSUE #2453: Input turns language to Chinese
0.17.0-17b48ff6e858b0c74116068cf6444bd578bbb747

Manual install

**What happened?**
If I search the phrase `parser error : invalid character in attribute value`, the search language changes to `zh`.

**How To Reproduce**
This works on every searx instance I can find. Just search the phrase `parser error : invalid character in attribute value`.

**Expected behavior**
Results in the language chosen.

MarcAbonce:
Fixes  #2453.

An empty `:` bang matches every entry in [languages.py](https://github.com/searx/searx/blob/master/searx/languages.py) with an empty string (every row without a country), so any query with a space separated colon will mistakenly match with Chinese.

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -77,7 +77,7 @@ class RawTextQuery:
                     pass
 
             # this force a language
-            if query_part[0] == ':':
+            if query_part[0] == ':' and len(query_part) > 1:
                 lang = query_part[1:].lower().replace('_', '-')
 
                 # check if any language-code is equal with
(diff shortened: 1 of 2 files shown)


return42 (approved):
Works! Thanks for your fast fix of this issue!

dalf:
Thank you !

PR #2458, opened 2021-01-12
Hide links panel in mobile screens

### LINKED ISSUE #2086: On mobile, the Links box obscures results

MarcAbonce:
Fix #2086


dalf (approved):
Thank you !

Issue #2442, opened 2021-01-06
[UX] Additional info on search engines in preferences/engines


ghost:
**Is your feature request related to a problem?**
I discovered a new engine on the search engine settings page (for example: naver) ­– so I had to do an extra search on it to determine _what it was and if it might be relevant to my use case_.

**Describe the solution you'd like**
A short descripton (introtext of its Wikipedia page or just the descripton from Wikidata) of the search engine.
Maybe even a specific template for all engines on additional info, like: open|closed source, own index|meta search of Bing+Yandex, locality-specific|worldwide, server country, etc.
…when the engine list grows it might also be helpful to categorize (open|closed source or so) and rank (on popularity or so) on the settings page for easier overlook.

**Describe alternatives you've considered**
I've had to do an extra search on the search engines name.

I'd love to read **other opinions** on the issue and **possible implementations** (especially pertaining to user experience)!

unixfox:
Good feature request but may I ask you how would you implement that in the Searx interface?

Where should the description appear? When moving the mouse cursor over the engine name in the preferences?

Ideally we would like to find a way so that it wouldn't require adding JavaScript code in order to make it compatible with and without JavaScript enabled.

dalf:
I suggest to modify the searx source of the engines. For example with `searx/engines/github.py` :
https://github.com/searx/searx/blob/14a395a6718949015e53a0b593eaa2f28e43fd27/searx/engines/github.py#L1-L17

Becomes : 
```python
"""
Github (It)

"""

from json import loads
from urllib.parse import urlencode

# about
about = {
    "website": 'https://github.com/',
    "wikidata_id": 'Q364',
    "official_api_documentation": 'https://developer.github.com/v3/',
    "results": 'JSON',
    "use_official_api": True,
    "require_api_key": False,
}
...
```

From [Q364](https://www.wikidata.org/wiki/Q364), a [script](https://gist.github.com/dalf/db85f103de9c28b5fcb7896525a6d41b#file-fetch_engine_descriptions-py) can fetch the engine descriptions in different languages.

See the [result here](https://gist.github.com/dalf/db85f103de9c28b5fcb7896525a6d41b#file-output-json) (173KB gzip compressed).

All these information can be displayed on a tooltip in the preferences.

But, some engine don't have a wikidata page.

The straightforward solution doesn't match the `wikidata_id` idea: 
```
about = {
    "website": 'https://github.com/',
[...]

PR #2460, opened 2021-01-13
[enh] engines: add about variable

dalf:
## What does this PR do?

move meta information from comment to the about variable
so the preferences, the documentation can show these information

the different URL have been checked and updated (especially API documentation URL).

Also:
* the UTF8 encoding header is removed. According to https://docs.python.org/3/howto/unicode.html Python supports writing source code in UTF-8 by default.
* add the SPDX License Identifier
 
See https://github.com/searx/searx/issues/2442

Currently the engine list in `/prefereneces` or in the [documentation](https://searx.github.io/searx/admin/engines.html) show a list of name without a description or a link.

This PR adds the required information to implement this feature, but I suggest to implement the UI in another PR.

Without this PR, the top head comment in an engine provides some information:
* `@website` :  the user URL, exactly what would be nice to display to the user.
* `@provide-api` : does the actual engine provide an official API ? usually the source code include an URL.
* `@using-api`: does searx use this API ?
* `@results`: what kind of data searx deal with ? HTML, JSON, XML
* `@stable`: is the searx engine stable ? 
* `@parse`: what kind of data the searx engine return ?

IMO:
* `@stable` without update will be wrong:
   * an official API can change.
   * the HTML layout of an engine can remain stable over time.
   * I suggest to remove this attribute : `@results` = `HTML` is enough to understand that searx engine is "unstable".
* `@parse`: is it useful ? I suggest to remove this attribute

In this PR, these attributes are replaced by a variable:
```python
about = {
    "website": 'https://arxiv.org',   # official user URL
    "wikidata_id": 'Q118398',   # wikidata identifier : see https://www.wikidata.org/wiki/Q28134166
    "official_api_documentation": 'https://arxiv.org/help/api',  # API documentation URL 
    "use_official_api": True,  # does the engine use the official API ?
    "require_api_key": False,  # does the engine requires a key ?
    "results": 'XML-RSS',  # same as before : what kind of data the searx engine deal with ?
}
```

The `wikidata_id` is interesting for different purposes:
* check the user URL
* automatically get a description of the engine in different languages (and logo?), see https://gist.github.com/dalf/db85f103de9c28b5fcb7896525a6d41b

* For now, the `about` is not used, there is nothing to test.

## Related issues

* Related to https://github.com/searx/searx/issues/2442 
* Related to https://github.com/searx/searx/issues/302#issuecomment-325743059
* Related to https://github.com/searx/searx/issues/1801

## WIP

* [x] settings.yml : add `about` section for engines using the `json` or `xpath` engines.
* [ ] check the about validity in `searx.engines.__init__`

### DIFF
--- a/searx/engines/1337x.py
+++ b/searx/engines/1337x.py
@@ -1,7 +1,21 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+ 1337x
+"""
+
 from urllib.parse import quote, urljoin
 from lxml import html
 from searx.utils import extract_text, get_torrent_size, eval_xpath, eval_xpath_list, eval_xpath_getindex
 
+# about
+about = {
+    "website": 'https://1337x.to/',
+    "wikidata_id": 'Q28134166',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
 
 url = 'https://1337x.to/'
 search_url = url + 'search/{search_term}/{pageno}/'
--- a/searx/engines/acgsou.py
+++ b/searx/engines/acgsou.py
@@ -1,17 +1,21 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
 """
  Acgsou (Japanese Animation/Music/Comics Bittorrent tracker)
-
- @website      https://www.acgsou.com/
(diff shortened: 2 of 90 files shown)


unixfox (approved):
LGTM and great changes!

PR #2467, opened 2021-01-14
[fix] github actions: use ubuntu-20.04 instead of ubuntu-latest

dalf:
## What does this PR do?

Github actions: use `ubuntu-20.04` instead of `ubuntu-latest`.

Side effect:
* `make local/py3`  does nothing because the github action cache already contains some data
* it seems the cache was corrupted
* so the `yaml` dependency was found, and CI was crashing

This PR doesn't fix the root cause: it just change the cache entry name.

Green is better than red, at least for the CI.

## Related issues

The CI broke with this PR: https://github.com/searx/searx/pull/2460 
A commit to disable the cache fixed the build (but without cache).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -5,10 +5,10 @@ on: [push, pull_request]
 jobs:
   python:
     name: Python ${{ matrix.python-version }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


unixfox (approved):
If that fixes the CI then I'm fine with that.

dalf:
And now it is docker : `docker push` ends with `denied: requested access to the resource is denied`. 

https://github.com/searx/searx/runs/1703192126?check_suite_focus=true

PR #2481, opened 2021-01-19
Mod check

dalf:
## What does this PR do?

* replace [pycld3](https://pypi.org/project/pycld3/) dependency by [langdetect](https://pypi.org/project/langdetect/)
* disable the checker by default
* include a minor bug fix about 

* this PR replace pycld3 by langdetect, so searx doesn't require additional native library.
* this PR ask the admin to opt-in for the checker 
* note : the checker requires some manual configuration in uwgi.ini, otherwise it is disabled.

* run the checker `kill -SIGUSR1 <searx pid>` or `searx-checker -v`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,4 @@ pygments==2.1.3
 python-dateutil==2.8.1
 pyyaml==5.3.1
 requests[socks]==2.25.1
-pycld3==0.20
+langdetect==1.0.8
--- a/searx/search/checker/__main__.py
+++ b/searx/search/checker/__main__.py
@@ -74,6 +74,7 @@ def run(engine_name_list, verbose):
                 stdout.write(f'    {"found languages":15}: {" ".join(sorted(list(checker.test_results.languages)))}\n')
                 for test_name, logs in checker.test_results.logs.items():
                     for log in logs:
+                        log = map(lambda l: l if isinstance(l, str) else repr(l), log)
                         stdout.write(f'    {test_name:15}: {RED}{" ".join(log)}{RESET_SEQ}\n')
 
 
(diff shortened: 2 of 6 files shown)


HLFH:
> replace pycld3 dependency by langdetect 

OK, I'll change this required dependency to `python-langdetect`

> include a minor bug fix about

OK

> disable the checker by default

Why? I love the checker. It is a good thing to know what is working and what does not work. For [`Sonarr`](https://github.com/Sonarr/Sonarr), we have an automated task that reminds us which indexer has stopped working, and it is enabled by default.

> note : the checker requires some manual configuration in uwgi.ini, otherwise it is disabled.

It is not in `uwsgi.ini` for ArchLinux if you have a correct setup. For searx with uwsgi, I have a separate `searx.ini` file in `/etc/uwsgi/vassals/searx.ini`. I only need to add the recommended line [at the end of this file](https://aur.archlinux.org/cgit/aur.git/tree/searx.ini?h=searx-git) and update the package, and this won't be manual for people that install `searx-git`.

I am against disabling by default the checker, so I am against commit https://github.com/searx/searx/pull/2481/commits/73c86f9bf233aa4f265d1c01ea94d01563e299f8.

unixfox:
> > disable the checker by default
> 
> Why? I love the checker. It is a good thing to know what is working and what does not work. For [`Sonarr`](https://github.com/Sonarr/Sonarr), we have an automated task that reminds us which indexer has stopped working, and it is enabled by default.

Probably because the checker introduce additional requests to the engine. And these additional requests could be the ones that make the Searx instance banned from the engine.

By disabling it by default this make it so the user know that it will introduce more requests to the engines.

asciimoo (approved):
Great! Thanks!

PR #2478, opened 2021-01-17
[fix] google videos - update xpaths and regex

0xhtml:
## What does this PR do?

This PR fixes the google videos engine by updating the xpaths (update/remove class names) and using new regexes to find the thumbnail sources. I also added parsing for author, length and date.

## Why is this change important?

Previously google videos always returned an exception.

## How to test this PR locally?

!gov test
!gov searx

## Author's checklist

~Some thumbnails can't be loaded and require additional parsing, that I couldn't figure out. E.g. thumbnails with an url like https://encrypted-tbn1.gstatic.com/images?q\u003dtbn:ANd9GcTEeuQr4iHLbDGbHSRKcWNLWZfdJCPjjbdyYXIuMgaoqPytjU1R9oZkeuvg~
Fixed by replacing `\u003d` with `=`.

## Related issues

N/A

PR #2482, opened 2021-01-20
[fix] revise of the google-Video engine

return42:
This revise is based on the methods developed in the revise of the google engine
(see commit 410c2f9).

## What does this PR do?

fix google-video engine using methods defined in the google engine / alternative approach #2478

Search videos by google-video:

     !gov hitchhiker's guide to the galaxy


unixfox:
Is this PR a replacement of this one: https://github.com/searx/searx/pull/2478?

return42:
Yes, it is more elaborated: It does not [hardcode language to en](https://github.com/searx/searx/pull/2478#pullrequestreview-572476779) .. it reuse the definitions we already have in the google engine.

Additional info: in PR #1985 I made a revise of all the google engines .. to much for one commit.  The idea behind #1985 is to normalize all the google engines since they use a similar concept for the query parameters:

https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/google.py#L2-L9

https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/google.py#L40-L44

But there are also parsing aspects which are equal. Here are the imports (reuse) from the google engine:

https://github.com/searx/searx/blob/e36d70d3190dbe3415d0304723ee2766e1562ced/searx/engines/google_videos.py#L32-L44

0xhtml:
I tested this PR. It works and the implementation seems better than mine #2478. The only problem I had with this PR is that `supported_languages` was empty and therefore the language ended up something like `en-US-DE` (language set to `de`).

return42:
> The only problem I had with this PR is that supported_languages was empty 

The engines data files needed to be updated (pushed 6a8c15e ) .. can you give it another try / thanks!

0xhtml:
I tested again and the language selection works now. But I noticed that the duration isn't parsed. The duration xpaths `./div[1]//a/div[3]` or `./div[1]//a/div[@class="ij69rd UHe5G"]` work for me. 
BTW is there a reason why you're not using the duration (as well as author and published) field of the template?

return42:
> BTW is there a reason why you're not using the duration (as well as author and published) field of the template?

Just inexperience. FYI: https://github.com/searx/searx/pull/2483#discussion_r562068101

Thanks for the hint .. I fixed it ..

return42:
@dalf I normalize the languages aspects of all google engines in this PR, see commit messages for details

Here is how I filtered for redirects:

    make SEARX_DEBUG=1 \
        search.checker.google_images \
        search.checker.google_news \
        search.checker.google_videos \
        search.checker.google \
        | grep -A1 "HTTP/1.1\" 3[0-9][0-9]"

searx/engines/google_videos.py
```diff
@@ -17,83 +61,152 @@
...
 categories = ['videos']
-paging = True
-safesearch = True
+paging = False
+language_support = True
+use_locale_domain = True
 time_range_support = True
-number_of_results = 10
+safesearch = True
+language_aliases = {}
```
dalf: Why is it necessary ?

According to 
https://github.com/searx/searx/blob/f4a17acb7a4fa82c5cb629f4eaad11ef528f89e4/searx/engines/__init__.py#L117-L120

`language_aliases` is initialized to `{}` even without the line 78, but I most probably miss something here.

searx/engines/google_videos.py
```diff
@@ -17,83 +61,152 @@
...
+        if extract_text(eval_xpath(result, g_section_with_header)):
+            logger.debug("ingoring <g-section-with-header>")
+            continue
+
+        title = extract_text(eval_xpath(result, title_xpath)[0])
+        url = eval_xpath(result, href_xpath)[0]
+        c_node = eval_xpath(result, content_xpath)[0]
+
+        # <img id="vidthumb1" ...>
+        img_id = eval_xpath(c_node, './div[1]//a/g-img/img/@id')
```
dalf: lines from 181 to 184 can be:
```python
img_id = eval_xpath_getindex(result, './div[1]//a/g-img/img/@id', 0, default=None)
if img_id is None:
  continue
```

If the index 0 is not found ( = the result list is empty), `img_id` will be `None`.

Note: I've tried to find the right function signatures for the `eval_xpath_*` functions, I'm not sure if it's help : the purpose is use the same function call(s) for the same pattern.

See:
https://github.com/searx/searx/blob/f4a17acb7a4fa82c5cb629f4eaad11ef528f89e4/searx/utils.py#L583-L609

searx/search/processors/online.py
```diff
@@ -248,6 +248,10 @@ def get_default_tests(self):
                 'matrix': {'query': 'paris', 'lang': 'en'},
                 'result_container': ['not_empty', ('has_language', 'en')],
             }
+            tests['lang_de'] = {
```
dalf: This describe a new test for all the engines that support the `language` parameter (I haven't run the code, just read the code for now).

For information, it is possible to add some custom tests to one engine. Just declare the global `additional_tests` variable inside the engine code:

```python
additional_tests = {
  'lang_de': {
      'matrix': {'query': 'paris', 'lang': 'de'},
      'result_container': ['not_empty', ('has_language', 'de')],
  }
}
```

dalf (changes requested):
I have just read the code for now.

searx/engines/google_videos.py
```diff
@@ -17,83 +62,162 @@
     "use_official_api": False,
     "require_api_key": False,
     "results": 'HTML',
+    "template": 'video.html',
```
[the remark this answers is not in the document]
return42: ups  .. sorry .. old issue .. will fix it.

searx/engines/google.py
```diff
@@ -87,7 +88,7 @@
     'TH': 'google.co.th',   # Thailand
     'TR': 'google.com.tr',  # Turkey
     'UA': 'google.com.ua',  # Ukraine
-    # 'CN': 'google.cn',    # China, only from China ?
+    'CN': 'google.cn',      # China
```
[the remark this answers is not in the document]
return42: > There is no google.cn 

Yes I now, that was the reason it was first commented out (in June last year) ... but yesterday I tried the URL https://www.google.cn/  and it exists .. and is not redirected to google.com.hk 

But now I see it is only a link to google.com.hk ..

searx/engines/google_videos.py
```diff
@@ -17,83 +61,152 @@
...
     return params
 
 
-# get response from search-request
 def response(resp):
+    """Get response from google's search request"""
     results = []
 
+    # detect google sorry
+    resp_url = urlparse(resp.url)
```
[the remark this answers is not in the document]
return42: Yes, sorry .. I forgot

searx/engines/google_videos.py
```diff
@@ -17,83 +61,152 @@
...
+            'title':       title,
+            'content':     content,
+            'length':      length,
+            'author':      pub_info,
+            'thumbnail':   img_src,
+            'template':    'videos.html',
+            })
+
+    # parse suggestion
+    for suggestion in eval_xpath(dom, suggestion_xpath):
```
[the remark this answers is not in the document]
return42: Before answering other remarks, I have to say, that I was totally focused on the language aspects and forgot what we have already discussed on google-news .. sorry, big fail of mine .. I will fix all this issues.

searx/engines/google_news.py
```diff
@@ -70,50 +70,51 @@
...
-        'lr': "lang_" + language,
+        'hl': lang_info['hl'],
+        'lr': lang_info['lr'],
         'ie': "utf8",
         'oe': "utf8",
-        'ceid' : "%s:%s" % (country, language),
-        'gl' : country,
-    })
+        'gl': lang_info['country'],
+    }) + ('&ceid=%s' % ceid)
```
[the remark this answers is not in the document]
return42: No, the value of contains a `:` and this will be url-encoded ..  to not encode, I added this as normal string to the end of the URL

searx/engines/google_videos.py
return42: The languages implementations are hard to understand for me .. but with your hint I looked again and see, it is not needed.

For me it is not clear for what language_aliases is good and when/why it is used or set ..

https://github.com/searx/searx/blob/f4a17acb7a4fa82c5cb629f4eaad11ef528f89e4/searx/engines/__init__.py#L122-L128

searx/search/processors/online.py
return42: > For information, it is possible to add some custom tests to one engine. 

Ah, thanks for the hint .. I have to learn more about checker .. 

> This describe a new test for all the engines that support the language parameter (I haven't run the code, just read the code for now).

Is it a problem to have a test with lang 'de' on all engines? .. I know from articles, that there are many searx user from de.

dalf: >  I have to learn more about checker ..

For now the code is the documentation, far from ideal for this feature.

>Is it a problem to have a test with lang 'de' on all engines?

The `lang_de` test fails with some other engines.
For example `!qwant :de time` returns mostly English results, `!qwant :de Deutsch` seems to be better.

I think it requires some back and forth between the engine code and the checker tests. 

For now, I feel it is better (?) to move the `lang_de` test to google.py ? (and import the `additional_tests` variable in the other google engines). Later we can generalize / adjust the test for the other engines.

return42: > For now the code is the documentation, 

It is OK for now .. I'm currently dealing with it .. when I have more experience, I can possibly document something 

> I think it requires some back and forth between the engine code and the checker tests.

That's right .. tests needs to be balanced between development and maintenance. E.g the google engine is bot sensitive and releases very fast CAPTCHA issues.  I added some additional test but commented them out for now, see 7121d995. They have their value when it comes to deep tests in the development cycle.

searx/engines/google.py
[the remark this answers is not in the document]
return42: Fixed to: https://github.com/searx/searx/pull/2482/files#diff-b79b3d5cfc19e1c5844863e984e6a97df8d1dee4fcdef1737f9055888b442c54R89

searx/engines/google_images.py
```diff
@@ -156,55 +165,47 @@ def response(resp):
...
-
-            img_base64_id = eval_xpath(img_node, '@data-iid')
-            if img_base64_id:
-                img_base64_id = img_base64_id[0]
-                thumbnail_src = img_bas64_map[img_base64_id]
+    for img_node in eval_xpath_list(root, './/img[contains(@class, "rg_i")]'):
+
+        img_alt = eval_xpath_getindex(img_node, '@alt', 0)
+
+        img_base64_id = eval_xpath(img_node, '@data-iid')
```
dalf, line 172: Equivalent (?) behavior using `eval_xpath_getindex`:
```python
img_base64_id = eval_xpath_getindex(img_node, '@data-iid', 0, default=None)
if img_base64_id:
    thumbnail_src = img_bas64_map[img_base64_id]
else:
    thumbnail_src = eval_xpath_getindex(img_node, '@src', 0, default='')
    if thumbnail_src == '':
        thumbnail_src = eval_xpath_getindex(img_node, '@data-src', 0, default='')
```

searx/engines/google_images.py
```diff
@@ -156,55 +165,47 @@ def response(resp):
...
-        except Exception as e:  # pylint: disable=broad-except
-            logger.error(e, exc_info=True)
-            # from lxml import etree
-            # logger.debug(etree.tostring(img_node, pretty_print=True))
-            # import pdb
-            # pdb.set_trace()
-            continue
+                thumbnail_src = ''
+
+        link_node = eval_xpath_getindex(img_node, '../../../a[2]',0 )
```
dalf: There is a missing space `,0 )` -> `, 0 )`

dalf (commented):
About google_images.py line 172 : I try to see if eval_xpath_getindex helps or not. Feel free to comment about the suggested code.

Ortherwise, LGTM

searx/engines/google_images.py
return42: Hm, it is equivalent, but what do we win with?

`eval_xpath_getindex` call at first `eval_xpath` and this function will raise `SearxXPathSyntaxException` and `SearxXPathSyntaxException` if something is wrong.

On a empty match, `bool(img_base64_id)` is the same as setting a default when the index 0 does not exists.

If I am not wrong the `eval_xpath_getindex` is equivalent but with additional conditions, checking that the range is OK and the default is not `NOTSET`

https://github.com/searx/searx/blob/0f18e885bf1996603fd6b3021c742b1126abda90/searx/utils.py#L602-L606

Or did I miss something?

dalf: It is easier for me to read the `eval_xpath_getindex` version, but I wrote that function so, I most probably not objective.

I try to see how far we can enforce checking using different dedicated functions (`eval_xpath_*`, etc...).

The current version is perfectly fine.

dalf:
Thank you!

searx/engines/google_images.py
return42: I think we need a review for each engine, to make them more search.checker conform. This can be done successively when we see a engine-scrapper is broken and needs a fix.

PR #2483, opened 2021-01-20
[fix] revise of the google-News engine

return42:
## What does this PR do?

This revise is based on the methods developed in the revise of the google engine
(see commit 410c2f9).

fix google-news engine and normalize with other google engines

Test google-news engine and time range

    !gon foo


return42:
About concept & motivation see https://github.com/searx/searx/pull/2482#issuecomment-764561637

searx/engines/google_news.py
```diff
@@ -16,74 +45,155 @@
...
+            # The first <a> tag in the <article> contains the link to the
+            # article The href attribute of the <a> is a google internal link,
+            # we can't use.  The real link is hidden in the jslog attribute:
+            #
+            #   <a ...
+            #      jslog="95014; 4:https://www.cnn.com/.../index.html; track:click"
+            #      href="./articles/CAIiENu3nGS...?hl=en-US&amp;gl=US&amp;ceid=US%3Aen"
+            #      ... />
+
+            jslog = eval_xpath(result, './article/a/@jslog')[0]
```
dalf: This line can rewritten this way:
```python
title = eval_xpath_getindex(result, './article/a/@jslog', 0)
```

eval_xpath_getindex [raises](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/utils.py#L559-L609) an [SearxEngineXPathException](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/exceptions.py#L96-L103) when the XPath is not found or when the index 0 is not found. This exception include the XPath.

Engine exceptions are [recorded](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/search/processors/online.py#L155).

The [`/stats/errors` URL](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/webapp.py#L952-L977) returns some statistics about the exceptions. It should help to debug the engines more easily (at least I hope).

But this doesn't work if there is global try ...  except.
Is it possible to remove it ?

searx/engines/google_news.py
```diff
@@ -16,74 +45,155 @@
...
+
+    # which subdomain ?
+    # subdomain = resp.search_params.get('google_subdomain')
+
+    # convert the text to dom
     dom = html.fromstring(resp.text)
 
-    # parse results
-    for result in dom.xpath('//div[@class="g"]|//div[@class="g _cy"]'):
+    for result in eval_xpath(dom, '//div[@class="xrnccd"]'):
```
dalf: It can rewritten this way:
`eval_xpath_list(dom, '//div[@class="xrnccd"]')`

[eval_xpath_list](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/utils.py#L559-L580) make sure that the result is a list.

searx/engines/google_news.py
```diff
@@ -16,74 +45,155 @@
     "use_official_api": False,
     "require_api_key": False,
     "results": 'HTML',
+    "parse": ('url', 'title', 'img_src', 'content'),
```
dalf: my take is `parse` is useless : 
* the code  between 187 and 192 is straightforward.
* from the user point of view, I don't know how it is relevant (if it is shown in the /preferences for example or the documentation).
* worst case scenario, this variable is not updated when the code is.
* writing this comment, I release that the same comment is valid for `results`.

See PR #2460

searx/engines/google_news.py
```diff
@@ -16,74 +45,155 @@
...
 
-# get response from search-request
 def response(resp):
+    """Get response from google's search request"""
     results = []
 
+    # detect google sorry
+    resp_url = urlparse(resp.url)
+    if resp_url.netloc == 'sorry.google.com' or resp_url.path == '/sorry/IndexRedirect':
+        raise RuntimeWarning('sorry.google.com')
```
dalf: See
https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/google.py#L205-L211

It should be [`SearxEngineCaptchaException`](https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/exceptions.py#L76-L83) instead of `RuntimeWarning` : Searx will then stop sending request for long time.

The default suspend time is 1 day, but the `suspended_time` parameter can provide another value (in second).

searx/engines/google_news.py
return42: Thanks for your remarks .. I just wanted to have a discussion, if it make sense to define a _interface_ .. but with your remarks in mind I also think _the code between 187 and 192 is straightforward._.

What I sometimes miss is a documentation of what parsed values are required by searx core (e.g. title) or supported by the templates ...  for me it is also not clear if it is OK to embed HTML in the _content_  or any other returned value:

https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/youtube_api.py#L31-L33

searx/engines/google_news.py
```diff
@@ -16,74 +45,155 @@
...
+    # subdomain = resp.search_params.get('google_subdomain')
+
+    # convert the text to dom
     dom = html.fromstring(resp.text)
 
-    # parse results
-    for result in dom.xpath('//div[@class="g"]|//div[@class="g _cy"]'):
+    for result in eval_xpath(dom, '//div[@class="xrnccd"]'):
+
         try:
```
return42: @dalf what do you think, should I remove this try / except catching all exceptions?  I thinks so, otherwise SearxXPathSyntaxException and others will be catched.

return42: sorry I should have read https://github.com/searx/searx/pull/2483/files#r561800269 first :-)

searx/engines/google_news.py
return42: Thanks a lot for the explanation .. I didn't aware of. I will remove try/except and make use of the new eval_xpath_[getindex|list]

searx/engines/google_news.py
```diff
@@ -18,72 +52,155 @@
...
 
-# get response from search-request
 def response(resp):
+    """Get response from google's search request"""
     results = []
 
+    # detect google sorry
+    resp_url = urlparse(resp.url)
+    if resp_url.netloc == 'sorry.google.com' or resp_url.path == '/sorry/IndexRedirect':
+        raise SearxEngineCaptchaException('sorry.google.com')
```
dalf: Lines from 122 to 127 could be a function defined in google.py

Then the different google*.py engines could import this function.

----

there is no need for a message:
`raise SearxEngineCaptchaException()` 

The OnlineProcessor catchs the exception:
https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/search/processors/online.py#L180-L183

Then webapp.py get the translation:
https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/webapp.py#L664

https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/webapp.py#L743-L750

searx/engines/google_news.py
```diff
@@ -18,72 +52,155 @@
...
+            # jslog="95014; 5:W251bGwsbnVsbCxudW...giXQ==; track:click"
+            jslog = jslog.split(";")[1].split(':')[1].strip()
+            try:
+                jslog = b64decode(jslog)
+            except binascii.Error:
+                # Sometimes (very rare) the base64 encoded string is not a
+                # multiple of four.
+                try:
+                    # fix incorrect padding (if possible)
+                    padding = (4 -(len(jslog) % 4)) * "="
```
dalf: even if the padding is most of time empty, it could be the general case:
```python
try:
  padding = (4 -(len(jslog) % 4)) * "="
  jslog = b64decode(jslog + padding)
except binascii.Error:
  continue
```
?

So the code is simpler

searx/engines/google_news.py
return42: > Lines from 122 to 127 could be a function defined in google.py

Good point, I had already thought about that

> there is no need for a message:

aargh .. sorry I am a C&P idiot .. will fix it / thanks for detailed explanation.

searx/engines/google_news.py
return42: Worth to test .. I will try ...

searx/engines/google_news.py
dalf: There is dedicated field for embedded media, see :
https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/youtube_noapi.py#L80-L87

dalf:
When I run 
```sh
. ./local/py3/bin/activate
SEARX_DEBUG=1 searx-checker -v "google news"
```

The checker says everything is okay, but actually there is a HTTP redirect for each request:
```
https://news.google.com:443 "GET /search?q=computer&hl=en&lr=lang_en&ie=utf8&oe=utf8&ceid=US%3Aen&gl=US HTTP/1.1" 302 0
https://news.google.com:443 "GET /search?q=computer&hl=en-US&lr=lang_en&ie=utf8&oe=utf8&ceid=US:en&gl=US HTTP/1.1" 200 None
```
(the checker should report that).

Is it possible to remove that redirect ? If it is not possible can you add this line
```
   params['soft_max_redirects'] = 1
```
to `def request(query, params)`
?

So `/stats/errors` doesn't report an error for each redirect.

For reference: https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/search/processors/online.py#L93-L102

searx/engines/google_news.py
return42: > There is dedicated field for embedded media

Yes but only in the videos.html template ..  so I stay with:

> What I sometimes miss is a documentation of what parsed values are required by searx core (e.g. title) or supported by the templates.

Does it make sense to add an issue? Title:  _missing template documentation_

Or do you think it will be useless for some reason I don't see right now.

**update**: not all replacements are supported by all designs (by example 6cbca14e4). I think at least we need to document some  _rules_

return42:
> but actually there is a HTTP redirect for each request:

Wow, checker is really cool stuff .. does find my typos :-)

Pushed 5f92dfc fixing the typo.

dalf:
It seems `en` is a special case.

Query `!google_news time` (or for the last one `!google_news 分`)

| params['languages'] | hl= (in params['url'])  | hl= (after redirect)                                                              |
|---------------------|-------|-----------------------------------------------------------------------------------|
| en                  | en    | en-US                                                                             |
| en-US               | en-US | (no redirect)                                                                     |
| en-CA            | en-CA | (no redirect) |
| de                  | de-DE | de                                                                                |
| de-DE               | de-DE | de                                                                                |
| de-AT               | de-AT | de                                                                                |
| fr-FR               | fr-FR | fr                                                                                |
| fr                  | fr-FR | fr                                                                                |
| es                  | es-ES | redirect to https://support.google.com/news/publisher-center/answer/9609687?hl=es |
| es-MX           | es-MX | es-419 |
| es-CL           | es-CL | es-419 |
| zh-TW         | zh-TW-TW | zh-TW |

So it seems : 
* `'hl': language,` is better than `'hl': lang_country,`
* `params['soft_max_redirects'] = 1` is required

Note about zh-TW testn even with `'hl': language,` there is a redirect:
1. `https://news.google.com/search?q=%E5%88%86&hl=zh-TW&lr=lang_zh-TW&ie=utf8&oe=utf8&ceid=TW%3Azh-TW&gl=TW`
1. `https://news.google.com/search?q=%E5%88%86&hl=zh-TW&lr=lang_zh-TW&ie=utf8&oe=utf8&ceid=TW:zh-Hant&gl=TW`

return42:
Thanks for your analyses .. can we optimize afterwards?

Without going into the details of your analysis I will give the hint that `make engines.languages` is also a part  of the context.

If this PR is in master, I can normalize the languages aspects of all google engines in PR #2482. See https://github.com/searx/searx/pull/2482#issuecomment-765495162 and 6a8c15e9.

dalf (approved):
What can adjust afterward.

The thing I would like to include is `params['soft_max_redirects'] = 1` not to record false error.

PR #2490, opened 2021-01-22
Add support for multiple architectures

winux1:
## What does this PR do?
Adds support for ARM and 32-bit devices. It was tested on Raspberry Pi 4 running Ubuntu Server 20.04 LTS 64-bit, on Raspberry Pi 3 running 20.04 LTS 32-bit and on a 32-bit computer running Linux Mint 19.3

Because it adds the ability to be able to install searx on a Raspberry Pi or on a 32-bit device using the Installation Scripts.

Run `sudo -H ./utils/filtron.sh install all` and `sudo -H ./utils/morty.sh install all` on a ARM64, ARMv* and also on 32-bit device.


return42:
Just want to leave a remark that we need to consider test & debugging aspects .. ATM I do not have a clue how we can support architectures on which we do not have access ...

I will have a look on this PR later / ATM live is busy ...

unixfox:
> Just want to leave a remark that we need to consider test & debugging aspects .. ATM I do not have a clue how we can support architectures on which we do not have access ...

GitHub actions allow virtualization, and you can for example use KVM for virtualizing other architectures.

return42:
@unixfox thanks for your hints ...

> GitHub actions allow virtualization

OK this covers the aspect of _automated tests._

> and you can for example use KVM for virtualizing other architectures.

Yes, this is what I am looking for: _bug tracking on local PC (i86)._

Quemu might be one alternative? .. I have to read some articles ..

winux1:
> @unixfox thanks for your hints ...
> 
> > GitHub actions allow virtualization
> 
> OK this covers the aspect of _automated tests._
> 
> > and you can for example use KVM for virtualizing other architectures.
> 
> Yes, this is what I am looking for: _bug tracking on local PC (i86)._
> 
> Quemu might be one alternative? .. I have to read some articles ..

 I think quemu or virt-manager would be a good option

unixfox:
We are getting a bit off topic, I opened a GitHub discussions for that: https://github.com/searx/searx/discussions/2491

.config.sh
```diff
@@ -22,6 +22,18 @@ if [[ ${PUBLIC_URL} == "https://searx.me" ]]; then
...
+
+if [[ $(uname -m) == x86_64* ]]; then
+    export ARCH=amd64
+elif [[ $(uname -m) == i*86 ]]; then
+    export ARCH=386
+elif  [[ $(uname -m) == aarch64* ]]; then
+    export ARCH=arm64
+elif  [[ $(uname -m) == arm* ]]; then
+    export ARCH=armv6l
+fi
```
return42: Since it is about go we should better name this environment GO_ARCH (GOLANG_ARCH) I think.

There are already some go stuff defined in 

https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/utils/lib.sh#L499-L501

I recommend to place the architecture detection there.

winux1: Good idea I will move the architecture detection and rename the variable

winux1: GO_ARCH or GOLANG_ARCH, what do you think?

dalf:
The changes in the PR seems simple

If we can't test the change, I suggest:
* to add warning for architectures != amd64 ( Like `the support of this architecture is experimental` )
* to merge this PR

---
Out of topic: the `detect_goarch` function in https://nextdns.io/install provides a more complete version, not tested.

return42:
FYI: I am looking for a LXC solution https://wh-plmapp-p.wh.root.local/static/images/fontawesome/font/fontawesome-webfont.woff

Isn't trivial to script .. but I want to give it a try.

Give me some time before merge ...

return42:
sorry .. C&P wrong link in my last post .. 

FYI: I am looking for a LXC solution https://github.com/searx/searx/discussions/2491#discussioncomment-302128

Isn't trivial to script .. but I want to give it a try.

Give me some time before merge ...

utils/lib.sh
```diff
@@ -499,6 +499,18 @@ service_is_available() {
 # golang
 # ------
 
+# Architecture detection
+
+if [[ $(uname -m) == x86_64* ]]; then
```
kvch, line 504: Nit: Is the command `arch` not availalbe? To me, the command name is more descriptive.

unixfox: It's not available on all Linux distributions. For instance, on ArchLinux the command `arch` doesn't exist.

winux1: True @unixfox

PR #2497, opened 2021-01-24
[fix] lxc.sh - SC2034: ubu2010_boilerplate appears unused.

return42:
$ make test.sh
    In utils/lxc.sh line 42:
    ubu2010_boilerplate="$ubu1904_boilerplate"
    ^-----------------^ SC2034: ubu2010_boilerplate appears unused. Verify use (or export if used externally).

    make test.sh

## Author's checklist

Should we add Makefile target 'test.sh' to 'test'? ..  needs [Shellcheck to be installed](https://searx.github.io/searx/admin/installation-searx.html?highlight=shellchecker#install-packages).

### DIFF
--- a/utils/lxc.sh
+++ b/utils/lxc.sh
@@ -39,6 +39,7 @@ $ubu1904_boilerplate
 echo 'Set disable_coredump false' >> /etc/sudo.conf
 "
 
+# shellcheck disable=SC2034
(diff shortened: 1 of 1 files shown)


dalf (approved):
Fix the issue.

Later, I think `make test.sh` should be included in https://github.com/searx/searx/blob/master/.github/workflows/integration.yml (not in the python section since it doesn't depend on the Python version).

PR #2498, opened 2021-01-24
[fix] google_news: avoid one HTTP redirect except for the English results

dalf:
## What does this PR do?

See https://github.com/searx/searx/pull/2483#issuecomment-766087712 and comments above.

* avoid one HTTP redirect except for the English results
* add `params['soft_max_redirects'] = 1` to avoid false error reporting in `/stats/errors`

* HTTP redirects slow down response times (debatable choice since this PR add one HTTP direct for the English language results).
* keep clean up error report in `/stats/errors` 

* search for `!google_news :en time` : check there is one HTTP redirect, but no ErrorContext report
* search for `!google_news :de time`: check there is no HTTP redirect, and no ErrorContext report.

## Related issues

* see https://github.com/searx/searx/pull/2483#issuecomment-766087712

### DIFF
--- a/searx/engines/google_news.py
+++ b/searx/engines/google_news.py
@@ -85,7 +85,7 @@ def request(query, params):
 
     query_url = 'https://'+ subdomain + '/search' + "?" + urlencode({
         'q': query,
-        'hl': lang_country,
+        'hl': language,
         'lr': "lang_" + language,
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks! .. today I will see what I can do for the other google engines ....

PR #2499, opened 2021-01-24
[mod] dynamically set language_support variable

### LINKED ISSUE #2485: useless language_support variable ?
The documentation says 

> `language_support` support language choosing

which make sense according to the variable name. Actually, this variable definition exists but is never used :

```
$ git grep "language_support = False"
searx/engines/doku.py:language_support = False
searx/engines/duden.py:language_support = False
searx/engines/etools.py:language_support = False
searx/engines/genius.py:language_support = False
searx/engines/youtube_noapi.py:language_support = False

$ git grep "language_support" | grep -v "language_support = True" | grep -v "language_support = False"
# documentation
docs/admin/engines.rst:        - {{(mod.language_support and "y") or ""}}
docs/dev/engine_overview.rst:language_support        boolean     support language choosing

# default value is True
searx/engines/__init__.py:                       'language_support': True,

# /config URL
searx/webapp.py:            'language_support': engine.language_support,

# False positives
searx/webapp.py:            stats[e.name]['supports_selected_language'] = _is_selected_language_supported(e, request.preferences)
searx/webapp.py:def _is_selected_language_supported(engine, preferences):
```

The use of this variable has been removed before the version 0.12.0 : https://github.com/searx/searx/commit/fd65c1292179fb082e965a1ee6a88b9298a54fc1#diff-f5962794ee51e4631700f577aa174f19623c87377130020dba94bd55410f5420L376

According to what I understand, the value should be `language_support = len(supported_languages) > 0` :
* `supported_languages` gets it's value from `searx/data/engines_languages.json` https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/__init__.py#L116-L120
* `supported_languages` default value is an empty list : https://github.com/searx/searx/blob/f310305c54c3cd1d9fc74f09453294edbd2b5486/searx/engines/__init__.py#L48

So I suggest:
* to set `language_support = len(supported_languages) > 0`, so the documentation, `/config` can return the right values.
* to remove all `language_support = True` / `language_support = False` in the engines.

ping @MarcAbonce

dalf:
## What does this PR do?

The language_support variable is set to True by default,
and set to False in only 5 engines.

Except the documentation and the /config URL, this variable is not used.

This commit remove the variable definition in the engines, and
set value according to supported_languages length: False when the length is 0,
True otherwise.

This variable is misleading: 
* from the user point of view, it does nothing, since https://github.com/searx/searx/commit/fd65c1292179fb082e965a1ee6a88b9298a54fc1#diff-f5962794ee51e4631700f577aa174f19623c87377130020dba94bd55410f5420L376 (committed 4 years ago)
* in https://searx.github.io/searx/admin/engines.html  the "Language support" column is wrong.

* `make docs` : check the documentation.

## Related issues

Close #2485

### DIFF
--- a/docs/dev/engine_overview.rst
+++ b/docs/dev/engine_overview.rst
@@ -42,7 +42,6 @@ argument                type        information
 ======================= =========== ========================================================
 categories              list        pages, in which the engine is working
 paging                  boolean     support multible pages
-language_support        boolean     support language choosing
 time_range_support      boolean     support search time range
 engine_type             str         ``online`` by default, other possibles values are 
                                     ``offline``, ``online_dictionnary``, ``online_currency``
@@ -97,7 +96,6 @@ example code
    # engine dependent config
    categories = ['general']
    paging = True
-   language_support = True
 
 
 making a request
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -44,7 +44,6 @@ babel_langs = [lang_parts[0] + '-' + lang_parts[-1] if len(lang_parts) > 1 else 
 engine_shortcuts = {}
 engine_default_args = {'paging': False,
                        'categories': ['general'],
-                       'language_support': True,
                        'supported_languages': [],
                        'safesearch': False,
                        'timeout': settings['outgoing']['request_timeout'],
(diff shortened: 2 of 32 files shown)


return42 (dismissed):
> This variable is misleading:

made some spot checks: confirmed! 

Thanks! ... do you like to merge right now?

PR #2500, opened 2021-01-24
[enh] every Sunday, call utils/fetch_*.py scripts and create a PR automatically

dalf:
## What does this PR do?

Use https://github.com/peter-evans/create-pull-request to create a PR when there is data update.

data in https://github.com/searx/searx/tree/master/searx/data

Example PR : 
* https://github.com/dalf/searx/pull/8
* include .github/workflows/data-update.yml as long the file is not into the master branch
* this PR was made before `utils/fetch_languages.py` update (second commit of this PR) :  engines_languages.json  and languages.json are committed in the wrong place.
* if the PR is not merged, new updates are push forced in the same branch / same PR.

This PR also makes a minor update of `utils/fetch_languages.py`.

See https://github.com/searx/searx/issues/2052#issuecomment-655607459

N/A

## Related issues

See #2052

### DIFF
--- a/.github/workflows/data-update.yml
+++ b/.github/workflows/data-update.yml
@@ -1,0 +1,67 @@
+name: "Update searx.data"
+on:
+  schedule:
+    - cron: "37 13 * * 0"
+
+jobs:
+  updateData:
+    name: Update data
+    runs-on: ubuntu-20.04
+    if: env.DATA_PR_TOKEN != null
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Ubuntu packages
+        run: |
+          sudo ./utils/searx.sh install packages
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+          architecture: 'x64'
+
+      - name: Cache Python dependencies
+        id: cache-python
+        uses: actions/cache@v2
+        with:
+          path: ./local
+          key: python-${{ matrix.os }}-3.9-${{ hashFiles('requirements*.txt', 'setup.py') }}
+    
+      - name: Install Python dependencies
+        if: steps.cache-python.outputs.cache-hit != 'true'
+        run: |
+          make V=1 install
+
+      - name: Fetch data
+        run: |
+          source local/py3/bin/activate
+          python utils/fetch_firefox_version.py
+          python utils/fetch_languages.py
+          python utils/fetch_ahmia_blacklist.py
+          python utils/fetch_wikidata_units.py
+          # python utils/fetch_currencies.py
+
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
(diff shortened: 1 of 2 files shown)


return42:
I have no experience with the git-workflows: If I merge this into the master of my fork (return42/searx) on github .. will it also create PRs on my my fork? .. I guess .. if so .. can we set a condition in the git-workflow, that the workflow is active only on repository searx/searx?

dalf:
I have updated the PR: the workflow will return only if the `DATA_PR_TOKEN` secret is defined (without this change, the workflow runs and fails to create the PR).

`DATA_PR_TOKEN` is a personal access tokens.

return42 (approved):
LGTM .. 

OT:  just a question of mine

> DATA_PR_TOKEN is a personal access tokens.

Tried to find, where I can set it using https://docs.github.com/en/github/developing-online-with-codespaces/managing-encrypted-secrets-for-codespaces .. but cant find 'Codespaces' what is documented there ... do I need a special contract to set such tokens?

dalf:
To get the token: your profile, Settings, Developer settings tab, Personal access tokens tab.
To set the secret: in your fork, go into Settings, then the "Secrets" tab, then the "New repository secret".

return42:
Dump question, is this git-workflow inactive?

When I `make useragents.update` I got ...

```diff
modified   searx/data/useragents.json
@@ -1,8 +1,10 @@
 {
     "versions": [
+        "85.0.1",
+        "85.0",
+        "84.0.2",
         "84.0.1",
-        "84.0",
-        "83.0"
+        "84.0"
     ],
     "os": [
         "Windows NT 10.0; WOW64",
```

What I mean; I can't see that this PR does anything .. I'm wrong? .. Is there a place, where can we monitor the tasks from https://github.com/searx/searx/blob/ab8739809c8c5f4c1f12e4b15fa8c61afe30ef9f/.github/workflows/data-update.yml#L3-L4

I guess the notation is what we know from unix [cron entries](https://en.wikipedia.org/wiki/Cron).. right?

dalf:
It fails yesterday: https://github.com/searx/searx/actions/runs/545400665

This line
https://github.com/searx/searx/blob/ab8739809c8c5f4c1f12e4b15fa8c61afe30ef9f/.github/workflows/data-update.yml#L10
should be
`if: secrets.DATA_PR_TOKEN != null`

return42:
> It fails yesterday
 
I see!  ... thanks for the link .. I should have more time to learn more about git-workflows. 

> this line  ...should be ..

If you have time do you like to fix it ..

now I see that is already used here

https://github.com/searx/searx/blob/ab8739809c8c5f4c1f12e4b15fa8c61afe30ef9f/.github/workflows/data-update.yml#L50

return42:
Seems that ` .github/workflows/data-update.yml` had another issue: https://github.com/searx/searx/actions/runs/561157462/workflow and it seems you did already fixed it https://github.com/searx/searx/commit/50bde93dbbd2acf36d916c97cf6fe6de6bbe18a1 ?

PR #2501, opened 2021-01-24
RFC: add searx/settings_default.py

dalf:
## What does this PR do?

* searx loads `searx/settings.yml` into the `settings` variable.
* then, the default values are defined usually using this pattern similar to `settings[....].get('aVariable', default)` 
* it fails for `settings['ui']['static_path']`, see https://github.com/searx/searx/issues/2252#issuecomment-764755567

This PR adds a new function `searx.settings_set_defaults.settings_set_defaults` to set all the default values in one place.

Then instead of 
```
static_path = get_resources_directory(searx_dir, 'static', settings['ui']['static_path'])
```

It is possible to write ```settings['ui']['static_path']```

See https://github.com/searx/searx/issues/2252#issuecomment-764755567 (more precisely the `static_path` definition)

This PR is request for comment / for archive : most probably a better solution is to create some classes in similar way to the `*Settings` classes.

The various defaults work as before.
https://github.com/dalf/searx/blob/d82185ad52e042db9d506bc952f5fbf8b8138b99/searx/settings_defaults.py#L93-L126

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -16,37 +16,17 @@ along with searx. If not, see < http://www.gnu.org/licenses/ >.
 '''
 
 import logging
+
 import searx.settings_loader
-from os import environ
-from os.path import realpath, dirname, join, abspath, isfile
+from searx.settings_defaults import searx_dir, settings_set_defaults
 
 
-searx_dir = abspath(dirname(__file__))
-engine_dir = dirname(realpath(__file__))
-static_path = abspath(join(dirname(__file__), 'static'))
 settings, settings_load_message = searx.settings_loader.load_settings()
+if settings is not None:
+    settings = settings_set_defaults(settings)
 
-if settings['ui']['static_path']:
-    static_path = settings['ui']['static_path']
 
-'''
-enable debug if
-the environnement variable SEARX_DEBUG is 1 or true
-(whatever the value in settings.yml)
-or general.debug=True in settings.yml
-disable debug if
-the environnement variable SEARX_DEBUG is 0 or false
-(whatever the value in settings.yml)
-or general.debug=False in settings.yml
-'''
-searx_debug_env = environ.get('SEARX_DEBUG', '').lower()
-if searx_debug_env == 'true' or searx_debug_env == '1':
-    searx_debug = True
-elif searx_debug_env == 'false' or searx_debug_env == '0':
-    searx_debug = False
-else:
-    searx_debug = settings.get('general', {}).get('debug')
-
+searx_debug = settings['general']['debug']
 if searx_debug:
     logging.basicConfig(level=logging.DEBUG)
 else:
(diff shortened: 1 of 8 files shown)


return42:
> This PR is request for comment / for archive : most probably a better solution is to create some classes in similar way to the *Settings classes.

I think this PR has its own value  ... cleaning up the stat_path mess .. better solutions can be implemented afterwards on top of this PR.

kvch:
@dalf Is this PR ready for review? Or why is a draft?

kvch (commented):
The general sentiment behind the PR is good. I like how you moved initialization of several things into one single file. 

I am interested why did you choose to create a schema like this. We already have a method/approach to parsing settings-like data from dictionaries in `preferences.py`. Is that way not suitable for parsing `settings.yml`? Or alternatively why not create a Settings object with typed attributes?

dalf:
> We already have a method/approach to parsing settings-like data from dictionaries in preferences.py.

preferences.py:
* describes a set of (key, value)
* entangled the values with the read / write operations (from and to cookies & HTML form).

The `settings` variable is a tree without error checking. In the master branch, the default values are set when the code needs them.

I agree it would better to merge two, but that's not a simple task.

-----

This is out of topic: if Settings & Preferences are classes:
* at application level : 
     * a `engine_list` method returns the engines described in `settings.yml` (so the tests are easy to write).
* from a request:
     * the constructor takes the settings at application level (it is already [more or less the case](https://github.com/searx/searx/blob/640d6ce77877551361b0acd116a9cd7faf47c611/searx/preferences.py#L325) )
     * the `engine_list` method returns only the visible engines. The `Preferences` class already has the [`validate_token`](https://github.com/searx/searx/blob/640d6ce77877551361b0acd116a9cd7faf47c611/searx/preferences.py#L515) but engine lst is throw away in the constructor of Preferences.

The `engine_list` is just an example, the general idea is to have. 
Tests would be much more easier to write.

Load / save the user preferences are managed in a different class (or at least remove the [`resp.set_cookie`](https://github.com/searx/searx/blob/640d6ce77877551361b0acd116a9cd7faf47c611/searx/preferences.py#L130) call in each `Setting` class).

That's only my opinion.

kvch:
> > We already have a method/approach to parsing settings-like data from dictionaries in preferences.py.
> 
> preferences.py:
> 
>   * describes a set of (key, value)
>   * entangled the values with the read / write operations (from and to cookies & HTML form).
> 
> 
> The `settings` variable is a tree without error checking. In the master branch, the default values are set when the code needs them.
> 
> I agree it would better to merge two, but that's not a simple task.

I am sorry if I was not clear. I did not say we should merge the two. `Preferences` has more responsibilities than simply parsing preferences. It also sets those in cookies, shows them in the UI. Maybe we could generalize a few parts of it, but I do not think it is worth the effort.

My question about the other approach is not answered. And I am still curious about why not introduce a `Settings` class with all the available attributes (please ignore the naming in the example above, it can be done better probably) like this:

```python
class General:
    def __init__(self):
        self.debug_mode = False

# and so on for all setting groups...

# finally the "big" instance settings class which includes all setting groups
class Settings:
    def __init__(self):
        self.general = General()
        self.brand = Brand()
        self.search = Search()
        self.server = Server()
```

It is true that you had added a constructor which is awesome. But why not go all the way and add the specific class(es)?

return42:
> But why not go all the way and add the specific class(es)?

FWIW that is what I thought also :-)

PR #2514, opened 2021-01-28
[fix] Makefile target gh-pages & flatten history of branch gh.pages

return42:
[fix] Makefile target gh-pages & flatten history of branch gh.pages

1. This patch fixes error:

        rm -rf gh-pages/
        make V=1 gh-pages
            make[1]: Leaving directory '/800GBPCIex4/share/searx'
        [ -d "gh-pages/.git" ] || git clone  gh-pages
        fatal: repository 'gh-pages' does not exist

2. The gh-page build has been moved to ./build/gh-pages this also affects
   'travis-gh-pages'

3. The gh-pages commit messages now includes a ref to the repository and commit

4. Since a gh-pages history has only the drawback that the reposetory grows
   fast, this patch also flattens the history:

        cd build/gh-pages/; git log --oneline
        bash: cd: build/gh-pages/: Datei oder Verzeichnis nicht gefunden
        026126b (HEAD -> gh-pages, origin/gh-pages) make gh-pages: from https://github.com/return42/searx.git@71d66979c2935312e0aed7fc7c3cf6199fbe88a2

## Related issues

Remove cryptic fly dirt from the commit message: https://github.com/JamesIves/github-pages-deploy-action/pull/576

### DIFF
--- a/utils/makefile.sphinx
+++ b/utils/makefile.sphinx
@@ -10,7 +10,7 @@ SPHINX_CONF ?= conf.py
 DOCS_FOLDER = ./docs
 DOCS_BUILD  = ./$(LXC_ENV_FOLDER)build/docs
 DOCS_DIST   = ./$(LXC_ENV_FOLDER)dist/docs
-GH_PAGES    ?= gh-pages
+GH_PAGES    ?= build/gh-pages
 
 BOOKS_FOLDER = ./docs
 BOOKS_DIST   = ./$(LXC_ENV_FOLDER)dist/books
@@ -172,24 +172,22 @@ PHONY += prepare-gh-pages
 prepare-gh-pages:
 	cp -r $(DOCS_DIST)/* $(GH_PAGES)/
 	touch $(GH_PAGES)/.nojekyll
-	echo "<html><head><META http-equiv='refresh' content='0;URL=index.html'></head></html>" > $(GH_PAGES)/404.html	
+	echo "<html><head><META http-equiv='refresh' content='0;URL=index.html'></head></html>" > $(GH_PAGES)/404.html
 
-PHONY += $(GH_PAGES)
-$(GH_PAGES)::
-	$(MAKE) docs
-	[ -d "gh-pages/.git" ] || git clone $(GIT_URL) gh-pages
-	-cd $(GH_PAGES); git checkout gh-pages >/dev/null
-	-cd $(GH_PAGES); git pull
-	-cd $(GH_PAGES); ls -A | grep -v '.git$$' | xargs rm -rf
+PHONY += gh-pages
+gh-pages: docs-clean docs
+	- git worktree remove -f $(GH_PAGES) || exit 0
+	- git branch -D gh-pages || exit 0
+	git worktree add --no-checkout $(GH_PAGES) master
+	cd $(GH_PAGES); git checkout --orphan gh-pages && git rm -rfq .
 	$(MAKE) prepare-gh-pages
 	cd $(GH_PAGES);\
 		git add --all . ;\
-		git commit -m "gh-pages: updated" ;\
-		git push origin gh-pages
+		git commit -q -m "make gh-pages: from $(shell git config --get remote.origin.url)@$(shell git rev-parse HEAD)" ;\
+		git push -f origin gh-pages
 
 PHONY += travis-gh-pages
-travis-gh-pages:
-	$(MAKE) docs
+travis-gh-pages: docs-clean docs
 	rm -Rf $(GH_PAGES)
 	mkdir -p $(GH_PAGES)
 	$(MAKE) prepare-gh-pages
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -80,7 +80,7 @@ jobs:
       with:
         GITHUB_TOKEN: ${{ github.token }}
         BRANCH: gh-pages
-        FOLDER: gh-pages
+        FOLDER: build/gh-pages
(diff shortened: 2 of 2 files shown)


utils/makefile.sphinx
```diff
@@ -172,20 +172,19 @@ PHONY += prepare-gh-pages
...
 
 PHONY += $(GH_PAGES)
 $(GH_PAGES)::
 	$(MAKE) docs
-	[ -d "gh-pages/.git" ] || git clone $(GIT_URL) gh-pages
-	-cd $(GH_PAGES); git checkout gh-pages >/dev/null
-	-cd $(GH_PAGES); git pull
-	-cd $(GH_PAGES); ls -A | grep -v '.git$$' | xargs rm -rf
+	. ./utils/brand.env; [ -d "gh-pages/.git" ] || git clone $$GIT_URL gh-pages
+	-cd $(GH_PAGES); git checkout --orphan gh-pages || exit 0
```
dalf: Does it make sense to replace `git clone $$GIT_URL gh-pages` by `git clone $$(git config --get remote.origin.url)` ?

With a soft fail (`|| exit 0`), we won't know if the documentation is deployed without looking at the logs.
With a loud and clear fail, we just have to check for the green icon.

return42: > replace git clone $$GIT_URL gh-pages by git clone 

makes sense, i will give it a try

> A loud and clear fail rather than || exit 0 : with a soft 

This catches only the checkout if branch is already checked out. Remember, this is a Makefile, this exit 0 does only affect this one line .. the following lines will proceed

dalf: > Remember, this is a Makefile, this exit 0 does only affect this one line .

I forgot that point. Thank you.
The thing that is not clear for me is when this line fails ?

return42: It fails when you run this target second time.  Here is a example what is suppressed by the exit 0:

     make gh-pages
     ....
     make gh-pages
     ...
     cd gh-pages; git checkout --orphan gh-pages
     fatal: A branch named 'gh-pages' already exists.
     make: [utils/makefile.sphinx:181: gh-pages] Error 128 (ignored)
     rm -rf gh-pages/* gh-pages/.gitignore
     make prepare-gh-pages
     ....

And this shows how it looks like with the exit 0

     ...
     cd gh-pages; git checkout --orphan gh-pages || exit 0
     fatal: A branch named 'gh-pages' already exists.
     rm -rf gh-pages/* gh-pages/.gitignore
     make prepare-gh-pages
     ....

dalf: Reading this https://stackoverflow.com/questions/21151178/shell-script-to-check-if-specified-git-branch-exists

Would
`-cd $(GH_PAGES); [ $$(git branch --list gh-pages) ] || git checkout --orphan gh-pages`
do the trick ?

Note: if for some reason the clone already exists, and the current branch is not gh-pages, then the script fails.

return42: Before you investigate more time .. I will try to implement a different solution using git worktree .. which will make cloning from a remote obsolete .. anyway thanks for your comments which helps me to take a closer look at this implementation  and to question the concepts I chose.

return42: > Would .. do the trick ?

Yes, but what do we win by this except additional git calls?  I will try to give a short explanation.. 

The `exit 0` exists only to suppress error messages from 'make' itself.  In the front there is a minus:

    -cd $(GH_PAGES); git checkout --orphan gh-pages || exit 0

This minus instructs [make to go further even if exit status of this line is not zero](https://www.gnu.org/software/make/manual/make.html#Errors) but also to report if this shell command fails. Doing both (exit 0 and the minus) might be superfluous (at least the minus), but the minus here is for better readability saying: _whatever the exit code of this line, ignore it and go further_

dalf: I'm not very familiar with Makefile.

So my suggestion is `cd $(GH_PAGES); [ $$(git branch --list gh-pages) ] || git checkout --orphan gh-pages`  ( I feel `|| exit 0` is like a `try: ... except: pass` which may hide an unexpected exception).

But that's okay with `-cd $(GH_PAGES); git checkout --orphan gh-pages || exit 0`.

return42:
I implemented a solution ... using git worktree .. also see updated https://github.com/searx/searx/pull/2514#issue-563565199

Do you know why the py3.8 build fails? I guess it has nothing to do with this PR .. or did I missed something?

dalf:
The worktree solution is nice.

> Do you know why the py3.8 build fails? I guess it has nothing to do with this PR .. or did I missed something?

I know, it is not related to this PR.
It github action cache problem and perhaps https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/utils/makefile.python#L101

> The gh-page build has been moved to ./build/gh-pages this also affects 'travis-gh-pages'

https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/.github/workflows/integration.yml#L83

Should be `FOLDER: build/gh-pages`

return42:
> Should be FOLDER: build/gh-pages

Ah, OK I will fixed it / thanks!

return42:
> It github action cache problem and perhaps 

I had a look but ...
I'm not very familiar with github actions :)

Do you know why it crashes now and how we can fix it?

return42:
Just want to leave a link here: https://github.com/JamesIves/github-pages-deploy-action/pull/576#issuecomment-770456094

dalf (approved):
Thank you !

Issue #492, opened 2016-01-11
Relax new requirements a bit


Glandos:
I know that supporting a software is easier when you control the environment, but is it possible to relax the requirements.txt versioning a bit?

Currently, I'm trying to use as less as possible the pip dependencies, and rely as much as possible on my distro's package manager (Debian/testing). I do this because I don't have the time to run `pip install -U` on every Python application I have every time a security issue appears. And the distro handles everything for me (which is why I'm using a distribution ;) ).

Maybe some requirements like `>=2.11.0` should be enough?

asciimoo:
@Glandos I understand your point, but I don't want to allow `>=` in requirements.txt. The main reason behind it is that if a dependency has backward incompatible changes, then it breaks searx.
You can still experiment with other dependency versions, probably searx will work fine with it.

unixfox:
@dalf Is it still a case that we want to support in the future, reducing the amount of included python libraries?

return42:
We recommend virtualenv and we have pined our [requirements](https://github.com/searx/searx/blob/master/requirements.txt).

Glandos:
In another project I contribute to, we already had [this discussion](https://github.com/spiral-project/ihatemoney/issues/658#issue-660754645). The conclusion is to use lower bounds and exclusion with known incompatible version and semantic matching (like `babel~=2.9`). Otherwise it renders packaging much harder, and security is also a concern here. If a new `certifi` package is published with removed certificate because of trust issues, then users must wait for `requirements.txt` to be updated first.
Of course, things like dependabot renders this easier.

To avoid infinite debate, I'll avoid to answer more here, so if you still don't agree, it's not needed to add more comment here :wink:

return42:
>  then users must wait for requirements.txt to be updated first.

That is right .. see last paragraph tagged with TLDR in my https://github.com/searx/searx/pull/2473#issuecomment-761582153

All solutions have their PROs and CONs and we rated the robustness and code maintenance higher than upward compatibility.

Searx is a meta search engine and its scrappers are needing regularly updates. That is why we recommend  to install from clone and maintain/update instance in short periods of time.

This scenario is complete different from developing and deploying a versioned application.

To pin versions is just a decision which fits best to our development and deployment cycles.

PR #2517, opened 2021-01-29
Update pyenv pyenvinstall Make targets

dalf:
## What does this PR do?

Change about utils/makefile.python: 
* `make pyenv` makes sure that there is a python executable in the virtualenv, if not the virtualenv is installed again.
* `make pyenvinstall` makes to install `pyyaml` before running setup.py (setup.py requires pyyml to read searx/settings.yml)

* github action workflow caches the ./local directory : 
https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/.github/workflows/integration.yml#L25-L32
* sometimes this cache can be corrupted (it is not clear how the cache gets corrupted).
* this PR ensure that `make test` works as expected even if
    * the python interpreter is missing
    * the virtualenv exists but pyyaml is missing

```
make pyenvinstall
rm ./local/py3/bin/python
make test
source ./local/py3/bin/activate
pip uninstall pyyaml
make pyenvinstall
```

### DIFF
--- a/utils/makefile.python
+++ b/utils/makefile.python
@@ -99,7 +99,8 @@ quiet_cmd_pyinstall = INSTALL   $2
 quiet_cmd_pyenvinstall = PYENV     install $2
       cmd_pyenvinstall = \
 	if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
-		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) ;\
+		rm -f $(PY_ENV)/requirements.sha256; \
+		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) &&\
 		sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ;\
 	else \
 		echo "PYENV     $2 already installed"; \
@@ -119,13 +120,13 @@ quiet_cmd_pyenvuninstall = PYENV     uninstall   $2
 # $2 path to folder where virtualenv take place
 quiet_cmd_virtualenv  = PYENV     usage: $ source ./$@/bin/activate
       cmd_virtualenv  = \
-	if [ ! -d "./$(PY_ENV)" ];then                                  \
-		$(PYTHON) -m venv $(VTENV_OPTS) $2;                     \
+	if [ -d "./$(PY_ENV)" -a -x "./$(PY_ENV_BIN)/python" ]; then \
+		echo "PYENV     using virtualenv from $2"; \
+	else \
+		$(PYTHON) -m venv $(VTENV_OPTS) $2; \
 		$(PY_ENV_BIN)/python -m pip install $(PIP_VERBOSE) -U pip wheel setuptools; \
-		$(PY_ENV_BIN)/python -m pip install $(PIP_VERBOSE) -r requirements.txt;     \
-	else                                                            \
-		echo "PYENV     using virtualenv from $2";              \
-        fi
+		$(PY_ENV_BIN)/python -m pip install $(PIP_VERBOSE) -r requirements.txt; \
+	fi
 
 # $2 path to lint
 quiet_cmd_pylint      = LINT      $@
--- a/utils/makefile.python
+++ b/utils/makefile.python
@@ -87,6 +87,22 @@ python-exe:
 	@:
 endif
 
+msg-pip-exe:
+	@echo "\n  $(PIP) is required\n\n\
+  Make sure you have updated pip installed, grab it from\n\
+  https://pip.pypa.io or install it from your package\n\
+  manager. On debian based OS these requirements are\n\
+  installed by::\n\n\
+    sudo -H apt-get install python$(PY)-pip\n" | $(FMT)
+
+ifeq ($(shell which $(PIP) >/dev/null 2>&1; echo $$?), 1)
+pip-exe: msg-pip-exe
+	$(error The '$(PIP)' command was not found)
+else
+pip-exe:
+	@:
+endif
+
 # ------------------------------------------------------------------------------
 # commands
 # ------------------------------------------------------------------------------
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -34,7 +34,7 @@ jobs:
         make V=1 install
         make V=1 gecko.driver
     - name: Run tests
-      run: make V=1 test
+      run: make V=1 ci.test
     - name: Test coverage
       run: make V=1 test.coverage
     - name: Store coverage result
--- a/Makefile
+++ b/Makefile
@@ -251,6 +251,11 @@ test.clean:
 # travis
 # ------
 
+PHONY += ci.test
+ci.test:
+	$(PY_ENV_BIN)/python -c "import yaml"  || make clean
+	$(MAKE) test
+
 travis.codecov:
 	$(Q)$(PY_ENV_BIN)/python -m pip install codecov
 


dalf:
From https://github.com/searx/searx/pull/2514#issuecomment-769731867 :

> Do you know why it crashes now and how we can fix it?

The cache contains a copy './local'. 
It is not possible to clear the cache: https://github.com/actions/cache/issues/2#issuecomment-673493515 (see the top of the page)

It seems the cache is corrupted.
* the directory contains `./local` without proper initialization.
* so `make pyenv` does nothing since the `./local` exists.
* so `make pyenvinstall` fails because `pyyaml` is not installed (since `./setup.py` requires pyyaml)

dalf:
It should fix the CI.

@return42  can you give a look ?

utils/makefile.python
```diff
@@ -99,7 +99,8 @@ quiet_cmd_pyinstall = INSTALL   $2
 quiet_cmd_pyenvinstall = PYENV     install $2
       cmd_pyenvinstall = \
 	if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
-		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) ;\
+		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install $$(cat ./requirements.txt | grep -i pyyaml); \
```
return42: This line can also be written:

     $$(grep -i pyyaml ./requirements.txt)

utils/makefile.python
```diff
@@ -99,7 +99,8 @@ quiet_cmd_pyinstall = INSTALL   $2
...
+		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) &&\
 		sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ;\
```
return42, line 120: OK, you only want to create a `$(PY_ENV)/requirements.sha256 ` when pip succeeds .. make sense to me.

Does it make sense to have a:

    rm -f $(PY_ENV)/requirements.sha256 ;

before pip installs?

utils/makefile.python
```diff
@@ -119,13 +120,13 @@ quiet_cmd_pyenvuninstall = PYENV     uninstall   $2
 # $2 path to folder where virtualenv take place
 quiet_cmd_virtualenv  = PYENV     usage: $ source ./$@/bin/activate
       cmd_virtualenv  = \
-	if [ ! -d "./$(PY_ENV)" ];then                                  \
-		$(PYTHON) -m venv $(VTENV_OPTS) $2;                     \
+	if [ -d "./$(PY_ENV)" -a -x "./$(PY_ENV_BIN)/python" ]; then \
```
return42, line 139: Checking the existence of `./$(PY_ENV_BIN)/python"` makes also sense to me.  This might help in cases where a previous build of the virtualenv has been failed for any reason.  

I thought twice about a more robust criteria but I don't have anything better to suggest.

return42 (dismissed):
See my comments.  All in all it LGTM .. there are two things that bother me a bit.

1. I made some test and killed the `make clean pyenv` to get a corrupt `/local` .. the pyyaml problem still exists under some circumstance.
2. `makefile.python` is _abstract_, only and handles _common python, pip and setuptools issues_. This patch adds some `pyyaml` aspects and is not an _abstract_ solution.

At the moment I don't have more time and I also don't have a better solution right now.  Is it OK for you to wait with merge.  Otherwise I can also think about a more _abstract_ solution afterwards.

utils/makefile.python
dalf: Done in fe3085d68e965732ca7291a14225b34a7e046e92

dalf:
> makefile.python is _abstract_

Abstract version:
* replace `$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install $$(grep pyyaml ./requirements.txt);`
* by `$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -r ./requirements-presetup.txt;`
* `requirements-presetup.txt` contains only one line `pyyaml==5.4.1` (to content has to be synchronized with `requirements.txt`).

But:
* If we could avoid to another requirements*.txt, it would be nice
* I prefer to find a way to remove the pyyaml dependency (related to https://github.com/searx/searx/pull/2408 )
* Is it a problem if utils/makefile.python is not abstract ?

utils/makefile.python
return42: Another thought:  now, with removing ` $(PY_ENV)/requirements.sha256` before running `pip install` and the `&& sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256` from line 104-105, checking existence of  ` $(PY_ENV)/requirements.sha256`  could be a good criteria here?

utils/makefile.python
```diff
@@ -99,7 +99,9 @@ quiet_cmd_pyinstall = INSTALL   $2
 quiet_cmd_pyenvinstall = PYENV     install $2
       cmd_pyenvinstall = \
 	if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
-		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) ;\
+		rm -f $(PY_ENV)/requirements.sha256; \
+		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install $$(grep pyyaml ./requirements.txt); \
+		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) &&\
```
return42, line 119: Thought: if I ignore the git workflow caching issue and have a complete new look at this two lines, I cant understand why line 103 is needed .. or did I lost the overview?

dalf: https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/setup.py#L11

brand is a variable initialized in `searx/__init__.py` :

https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/searx/__init__.py#L27

https://github.com/searx/searx/blob/71d66979c2935312e0aed7fc7c3cf6199fbe88a2/searx/settings_loader.py#L8

utils/makefile.python
dalf: Is there a case where the `pyenv` target is invoked but not the `pyenvinstall` target ?
* if no, why there are two targets ?
* if yes, then we _can't_ `$(PY_ENV)/requirements.sha256` : each invocation of `pyenv` will reinstall the virtualenv.

utils/makefile.python
return42: Sorry I am an idiot .. I totally forgot the main problem

utils/makefile.python
return42: These makefiles has been implemented for general use cases.  I use them in several projects as [boilerplate](https://return42.github.io/boilerplate/slides/index.html#/).  The motivation was to have  typical development workflows localy at hand, build on simple tools available nearly everywhere.

Back to your question _why there are two targets ?_ 

pyenv: builds the environment for development / typically `requirements.txt`
penvinstall: installs the python package in this environment (`pip install -e`) / typically `setup.py` 

In the searx project `requirements.txt` and `setup.py` include identical _requirements_ but this is [not typical for python projects](https://github.com/searx/searx/issues/492#issuecomment-768923721).  Beside this you can also think about projects containing more than one package.

> Is there a case where the pyenv target is invoked but not the pyenvinstall target ?

With the words above, my answer is: _it depends .._ But from `makefile.python`'s POV the answer is : yes

Your questions helped me to sort my thoughts, but for today I am to tired .. tomorrow I have more time, then I'll think about it again ..

- pyyaml in setup.py
- _each invocation of pyenv will reinstall the virtualenv_

utils/makefile.python
return42: Hm .. after thinking again .. pyyaml  should already be installed when the virtualenv is build

https://github.com/dalf/searx/blob/fe3085d68e965732ca7291a14225b34a7e046e92/utils/makefile.python#L122-L129

and this will be installed before pyenvinstall is called:

https://github.com/dalf/searx/blob/fe3085d68e965732ca7291a14225b34a7e046e92/utils/makefile.python#L193

In https://github.com/searx/searx/pull/2517#discussion_r567049019 I described it:

> pyenv: builds the environment for development / typically `requirements.txt`
> penvinstall: installs the python package in this environment (`pip install -e`) / typically `setup.py` 

The other point you asked:

> each invocation of pyenv will reinstall the virtualenv

Our mistake was to build `sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ;\` in ` cmd_pyenvinstall` .. this optimizing should take place in `cmd_virtualenv`.

Or I am wrong? .. I will give it a try ...

dalf: > Hm .. after thinking again .. pyyaml should already be installed when the virtualenv is build

cmd_virtualenv seems to fail sometimes on github action:
* in this case, the ./local contains an initialized virtualenv, but without pyyaml
* cmd_virtualenv won't detect that pyyaml is missing (or other dependencies): the virtualenv exists, `...python` is an executable.
* then cmd_pyenvinstall fails because pyyaml is missing.
* the cache is updated with a broken virtualenv.
* all new builds on github action fail.

An alternative implementation (without `...grep pyyaml ./requirements.txt)...` ) :
* after `$(PY_ENV_BIN)/python -m pip install $(PIP_VERBOSE) -r requirements.txt;`, a file is created (or just a copy of requirements.txt)
* `if [ -d "./$(PY_ENV)" -a -x "./$(PY_ENV_BIN)/python" ]; then` also check if this file exists.
* basically it duplicates `if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then` in cmd_virtualenv.

return42: I work on a solution .. give me some time this day and I will show you my solution:

> An alternative implementation 
> ...
> ..

All this goes in the right direction to hardening, but instead hardening the installation, we have to hardening the venv build .. I will consider it in my solution / thanks!

return42: @dalf I implemented a alternative PR #2524 ... its all described there ..

Issue #2519, opened 2021-01-29
Prometheus metrics searx


matthiasantierens:
**Describe the solution you'd like**

It would be nice to have Prometheus metrics for Searx.

for example:
 - amount of searches
 - response time
 - gets and posts
 - status codes
 - health endpoints
 - amount registered ip's

PR #2523, opened 2021-01-30
Add more metrics

dalf:
## What does this PR do?

Updated version of https://github.com/searx/searx/pull/477 : 
* [jqplot](https://github.com/jqPlot/jqPlot/releases) to display the plots. (FIXME: no new version since 2016 and use javascript).
* this is a WIP

With this PR, searx records more statistics, see the `/stats` page: 

At the bottom of the page, there is the search response time: average and quartiles
image

Above the response time per engine:
image

If you click on engine time, you can get the quartiles. Here the details for the wikidata engine (two requests have been recorded):
image

Above there are the errors per engine. FIXME : Here sometimes the percentage can go above 100% (example, in the same response, two errors are recorded : a redirect AND a CAPTCHA). 
image

At the top of the page:
image

TODO: detail description of the changes in `searx.search.processors`

More detail statistics can help:
* to find why sometimes searx is slow
* to understand the errors (more readable than `/stats/errors`)

* `make run `
* use different engines

## Related issues

* Superseed https://github.com/searx/searx/pull/477
* More or less related to https://github.com/searx/searx/issues/2519


dalf:
Request For Comment: jqplot can be replaced by pygal :

> pygal is a dynamic SVG charting library written in python. 

* http://www.pygal.org/en/stable/
* https://github.com/Kozea/pygal
* no need dependency except pygal.
* SVG as an image is safe: https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image#restrictions

The output format can be png instead of svg : `pip install pygal[png]` to install [`cairosvg`](https://pypi.org/project/CairoSVG/) which requires [Pillow](https://github.com/python-pillow/Pillow)

return42:
Wow ... very cool .. some unsorted remarks from mine:

JS files from jqplot include [CR+LF](https://en.wikipedia.org/wiki/Newline#Representation) is this needed? .. I guess the origin is utf-8 unix style (LF only).

> Request For Comment: jqplot can be replaced by pygal

At some point we might want a template without JavaScript. Having a server side solution makes sense to me.

> no need dependency except pygal.

cool .. 

> The output format can be png instead of svg 

SVG seems perfect to me ..  Having Cairo and Pillow on board seems to bring some mess to me.

What do you think, will it be hard to implement a server side solution using pygal? I haven't understand all details right now, but the architecture seems to make it possible  .. [stats.js](https://github.com/searx/searx/pull/2523/files#diff-0b5147a348fa76c00361854829fc3b918f768ae4f62682f401bc90fbb970ad37) LGTM!!

dalf:
I close this PR in favor of https://github.com/searx/searx/pull/2561

PR #2524, opened 2021-01-30
Hardening travis 'Run tests' workflow when git-workflow caching fails

return42:
## What does this PR do?

Hardening travis 'Run tests' workflow. This PR is a alternative to #2517.  I recommend to read along the git comments.  First the makefile.python is polished and the requirement files are organzied.  The last commit [ensures that `make test` works as expected](https://github.com/searx/searx/pull/2517#discussion_r567240235) even if

- the python interpreter is missing
- the virtualenv exists but pyyaml is missing

github action workflow caches the ./local directory.  Sometimes this cache can be corrupted (it is not clear how the cache gets corrupted).

- https://github.com/searx/searx/pull/2514#issuecomment-769731867
- #2517

1. initial run `make travis.test`
2. manipulate the virtualenv, e.g. delete the files from the yaml module

now you have _a coroupted cache_ ....

3. run `make travis.test` again

The tests should run without fail.

### DIFF
--- a/docs/dev/makefile.rst
+++ b/docs/dev/makefile.rst
@@ -36,27 +36,33 @@ Python environment
 
    ``source ./local/py3/bin/activate``
 
-With Makefile we do no longer need to build up the virtualenv manually (as
-described in the :ref:`devquickstart` guide).  Jump into your git working tree
-and release a ``make pyenv``:
+With Makefile we do no longer need to build up the virtualenv manually. Jump
+into your git working tree and release a ``make pyenv``:
 
 .. code:: sh
 
    $ cd ~/searx-clone
    $ make pyenv
-   PYENV     usage: source ./local/py3/bin/activate
+   PYENV     [virtualenv] usage: source ./local/py3/bin/activate
+   PYENV     [virtualenv] installing requirements*.txt into ./local/py3
    ...
+   Successfully installed ...
 
 With target ``pyenv`` a development environment (aka virtualenv) was build up in
 ``./local/py3/``.  To make a *developer install* of searx (:origin:`setup.py`)
 into this environment, use make target ``install``:
 
-.. code:: sh
+.. code:: text
 
    $ make install
-   PYENV     usage: source ./local/py3/bin/activate
-   PYENV     using virtualenv from ./local/py3
-   PYENV     install .
+   PYENV     [virtualenv] usage: source ./local/py3/bin/activate
+   PYENV     [virtualenv] using ./local/py3 // glob pattern requirements*.txt --> sha256 OK
+   ...
+   ModuleNotFoundError: No module named 'searx'
+   PYENV     [pyenvinstall]  pip install -e .\[test\]
+   ...
+     Running setup.py develop for searx
+   Successfully installed ... searx ...
 
 You have never to think about intermediate targets like ``pyenv`` or
 ``install``, the ``Makefile`` chains them as requisites.  Just run your main
@@ -64,13 +70,26 @@ target.
 
 .. sidebar:: drop environment
 
-   To get rid of the existing environment before re-build use :ref:`clean target
-   <make clean>` first.
+   To get rid of the existing environment before re-build use: :ref:`make clean`
 
-If you think, something goes wrong with your ./local environment or you change
-the :origin:`setup.py` file (or the requirements listed in
+If you think, something goes wrong with your ``./local`` environment or you
+change the :origin:`setup.py` file (or the requirements listed in
 :origin:`requirements-dev.txt` and :origin:`requirements.txt`), you have to call
 :ref:`make clean`.
+
+``PYENV_REQ``: requirement files (glob pattern: ``requirements*.txt``)
+  For ever reasons, if you want to install only ``requirements.txt``
+
+  .. code:: text
+
+     $ make PYENV_REQ=requirements.txt clean pyenv
+     make PYENV_REQ=requirements.txt clean pyenv
+     CLEAN     ...
+     ...
+     PYENV     [virtualenv] usage: source ./local/py3/bin/activate
+     PYENV     [virtualenv] installing requirements.txt into ./local/py3
+     ...
+     Successfully installed ...
 
 
 .. _make run:
@@ -86,9 +105,15 @@ browser (:man:`xdg-open`):
 .. code:: sh
 
   $ make run
-  PYENV     usage: source ./local/py3/bin/activate
-  PYENV     install .
-  ./local/py3/bin/python ./searx/webapp.py
+  PYENV     [virtualenv] usage: source ./local/py3/bin/activate
+  PYENV     [virtualenv] installing requirements*.txt into ./local/py3
+  ...
+  PYENV     [pyenvinstall] pip install -e .\[test\]
+  ...
+    Running setup.py develop for searx
+  Successfully installed searx
+  ...
+  SEARX_DEBUG=1 ./local/py3/bin/python ./searx/webapp.py
   ...
   INFO:werkzeug: * Running on http://127.0.0.1:8888/ (Press CTRL+C to quit)
   ...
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
+-r requirements.txt
```
dalf: why ? 
It makes the setup more complicated.

Usually the way to go is:
* `pip install -r requirements.txt`
* `pip install -r requirements-dev.txt`
* `pip install -e .[test]`

utils/makefile.python
```diff
@@ -19,6 +19,7 @@ PY       ?=3
 PYTHON   ?= python$(PY)
 PIP      ?= pip$(PY)
 PIP_INST ?= --user
+PY_PRJ_REQ ?= requirements.txt
```
dalf: when to use `make PY_PRJ_REQ=requirements-dev.txt pyenv` and not `make pyenvinstall` ?
(related to `-r requirements.txt` in `requirements-dev.txt` )

Makefile
```diff
@@ -251,6 +251,11 @@ test.clean:
 # travis
 # ------
 
+PHONY += travis-pyenv
+travis.test:
```
dalf: `travis.codecov` is misleading: we don't use travis anymore, but [github actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).

A generic name can be `ci.test` (so why don't have to rename the target if we move away from github actions).

Anyway it is simple effective way to make sure the CI will work !

utils/makefile.python
```diff
@@ -97,13 +115,7 @@ quiet_cmd_pyinstall = INSTALL   $2
...
 # $2 path to folder with setup.py, this uses pip from pyenv (not OS!)
 quiet_cmd_pyenvinstall = PYENV     install $2
-      cmd_pyenvinstall = \
-	if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
-		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) ;\
-		sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ;\
-	else \
-		echo "PYENV     $2 already installed"; \
-	fi
+      cmd_pyenvinstall = $(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS)
```
dalf: Why remove the check about `requirements-dev.txt` ?

Without it, `make test` check each time the dependencies of `requirements-dev.txt`: everything since there is `-r requirements.txt`.

requirements-dev.txt
return42: Is it? then we should implement it. ATM in branch master the workflow of `make pyenvisntall` is slightly different:

1. `pyenv` --> `pip install  -U pip wheel setuptools; pip install -r requirements.txt`
2. `pyenvinstall` --> `pip  install -e .\[test\] ; sha256sum requirements*.txt > ./local/py3/requirements.sha256`

ATM in branch master the `requirements-dev.txt` is never used by a pip command.  But `pip  install -e .\[test\]` will read from this file

But I will try to implement what you suggested.

utils/makefile.python
return42: The question should be: _In current master branch, where is `requirements-dev.txt` used?_ .. see https://github.com/searx/searx/pull/2524#discussion_r567641395

requirements-dev.txt
return42: 'pyenv' now installs `requirements*.txt` ... see commit message https://github.com/searx/searx/pull/2524/commits/b66df00

utils/makefile.python
return42: I renamed this variable to `PYENV_REQ ?= requirements*.txt` .. a glob pattern matching requirement files.

Makefile
return42: Good point .. renamed to `ci.test` (https://github.com/searx/searx/pull/2524/commits/b53b91e1c4c7034f7777426f1679f6b68d79a487).

utils/makefile.python
return42: > Why remove the check about requirements-dev.txt ?

The sha256sum of the `requirements*.txt` is done in the `cmd_virtualenv` function.

The function `cmd_pyenvinstall` does now try to `import searx,  pkg_resources` .. installation is only done if this import fails.

return42:
@dalf since you have merged #2517 there is only one commit 88d870b left from this PR. Do you like to have a look?  / Thanks!

utils/makefile.python
```diff
@@ -112,15 +114,19 @@ quiet_cmd_pyinstall = INSTALL   $2
...
+quiet_cmd_pyenvinstall = PYENV     [pyenvinstall] $(PY_ENV): install $2
       cmd_pyenvinstall = \
-	if ! cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
-		rm -f $(PY_ENV)/requirements.sha256; \
-		$(PY_ENV_BIN)/python -m pip $(PIP_VERBOSE) install -e $2$(PY_SETUP_EXTRAS) &&\
-		sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ;\
+	_IMP_STATEMENT="import $(foreach var,$(wildcard $(PYOBJECTS)),$(var), ) pkg_resources"; \
+	_PRINT="print('          [pyenvinstall] pkg_resources.working_set --> resources %s' % len([True for i in pkg_resources.working_set]))"; \
+	if ! $(PY_ENV_BIN)/python -c \
+		"import sys; sys.path.pop(0); $$_IMP_STATEMENT; $$_PRINT" 2>/dev/null;\
```
dalf: I end up to understand, but why display this information ?

return42: > I end up to understand, but why display this information ?

good question .. I will remove it.

return42: > I end up to understand, but why display this information ?

Line has been removed ..

docs/dev/makefile.rst
```diff
@@ -36,42 +36,61 @@ Python environment
...
+   PYENV     [virtualenv] installing requirements*.txt into ./local/py3
    ...
+   Successfully installed ...
 
 With target ``pyenv`` a development environment (aka virtualenv) was build up in
 ``./local/py3/``.  To make a *developer install* of searx (:origin:`setup.py`)
 into this environment, use make target ``install``:
 
-.. code:: sh
+.. code:: txt
```
kvch: Why did you change it to `txt`? Isn't this block showing shell output?

docs/dev/makefile.rst
```diff
@@ -36,42 +36,61 @@ Python environment
...
-the :origin:`setup.py` file (or the requirements listed in
+If you think, something goes wrong with your ``./local`` environment or you
+change the :origin:`setup.py` file (or the requirements listed in
 :origin:`requirements-dev.txt` and :origin:`requirements.txt`), you have to call
 :ref:`make clean`.
 
+``PYENV_REQ``: requirement files (glob pattern: ``requirements*.txt``)
+  For ever reasons, if you want to install only ``requirements.txt``
+
+  .. code:: txt
```
kvch: Same: Why is it `txt`?

docs/dev/makefile.rst
```diff
@@ -102,10 +127,13 @@ Drop all intermediate files, all builds, but keep sources untouched.  Includes
 target ``pyclean`` which drops ./local environment.  Before calling ``make
 clean`` stop all processes using :ref:`make pyenv`.
 
-.. code:: sh
+.. code:: txt
```
kvch: `txt`?

docs/dev/makefile.rst
return42: > Isn't this block showing shell output?

It shows the output! .. The (plain text) output of any command should not be highlighted by a bash syntax.  But thanks for the hint: it should be `text` not `txt` .. I fixed, rebased and force pushed ..

docs/dev/makefile.rst
return42: see https://github.com/searx/searx/pull/2524#discussion_r579314307

docs/dev/makefile.rst
return42: see https://github.com/searx/searx/pull/2524#discussion_r579314307

return42:
Closed in favor of #2594

return42:
Closed in favor of #2594

PR #2529, opened 2021-02-02
RFC : [mod] work on Makefile

dalf:
## What does this PR do?

The Makefile files allow to do a lot of things all from the `make` command.
As far I understand, it is possible to do a lot of thing all from `make`.

Unfortunately, the code is really not easy to understand.

With PR, Makefile focuses on few thing.
Other features requires to use activate the virtualenv.

Reading only Makefile, it is possible to understand what the different targets do.


Makefile
```diff
@@ -85,6 +111,121 @@ prebuild-includes:
...
+# ----------------
+
+quiet_cmd_pyenv  = PYENV     usage: $ $(PY_ENV_ACT)
+      cmd_pyenv  = \
+	if [ -d "./$(PY_ENV)" -a -x "./$(PY_ENV_BIN)/python" ]; then \
+		echo "PYENV     using virtualenv from $(PY_ENV)"; \
+	else \
+		rm -f $(PY_ENV)/requirements.sha256; \
+		$(PYTHON) -m venv $(VTENV_OPTS) $(PY_ENV); \
+	fi
```
dalf, line 125: Here pyenv does one thing: install the virtualenv, nothing more.

It is possible to check if `$(PYTHON)` exists, but anyway the command will fail.

Makefile
```diff
@@ -85,6 +111,121 @@ prebuild-includes:
...
+      cmd_pyenvinstall = \
+	if cat $(PY_ENV)/requirements.sha256 2>/dev/null | sha256sum --check --status 2>/dev/null; then \
+		echo "PYENV     already installed"; \
+	else \
+		rm -f $(PY_ENV)/requirements.sha256; \
+		$(PY_ENV_ACT); \
+		python -m pip install $(PIP_VERBOSE) -U pip wheel setuptools; \
+		python -m pip install --use-feature=2020-resolver -r requirements.txt -r requirements-dev.txt && \
+		sha256sum requirements*.txt > $(PY_ENV)/requirements.sha256 ; \
+	fi
```
dalf, line 137: pyenvinstall installs requirements.txt and requirements-dev.txt, so pyyaml is not required.

Makefile
```diff
@@ -104,7 +245,7 @@ useragents.update:  pyenvinstall
 	$(Q)echo "Update searx/data/useragents.json with the most recent versions of Firefox."
 	$(Q)$(PY_ENV_ACT); python utils/fetch_firefox_version.py
 
-buildenv: pyenv
+buildenv: pyenvinstall
```
dalf, line 248: pyenv won't contains the dependencies from `requirements.txt`.

Makefile
```diff
@@ -181,8 +324,8 @@ $(ENGINE_TARGETS): pyenvinstall
 # test
 # ----
 
-PHONY += test test.sh test.pylint test.pep8 test.unit test.coverage test.robot
-test: buildenv test.pylint test.pep8 test.unit gecko.driver test.robot
+PHONY += test test.sh test.pep8 test.pylint test.unit test.coverage test.robot
+test: buildenv test.pep8 test.pylint test.unit test.robot
```
dalf, line 328: * pylint is slow
* pep8 is fast, and check typo (more than that).

test.pep8 first allow to fail fast on a typo (missing whitespace for example)

docs/conf.py
```diff
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
 
 import  sys, os
+
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).parent))
+sys.path.append(str(Path(__file__).parent.parent))
```
dalf, line 8: So it is possible to import shpinx_build_tools without this 

`SITE_PYTHON ?=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))site-python`

(I understood why it was required after I moved sphinx_build_tools.py )

utils/makefile.sphinx
```diff
@@ -1,20 +1,16 @@
 # -*- coding: utf-8; mode: makefile-gmake -*-
 
-export DOCS_FOLDER DOCS_BUILD DOCS_DIST BOOKS_FOLDER BOOKS_DIST
+export DOCS_FOLDER DOCS_BUILD DOCS_DIST
```
dalf, line 3: We don't publish the .tex  / .pdf versions.
Tex requires a lot of dependencies (not remove from utils/searx.sh in this PR )

return42:
What is the intention of this PR .. would just reduce the number of make-files?

I haven't understand al changes of this PR but I will try to leaf my remarks _here and there_

> As far I understand, it is possible to do a lot of thing all from make.

Makefile are just set of [rules](https://www.gnu.org/software/make/manual/html_node/Rule-Example.html#Rule-Example).  A rule mainly consist of :

     target: prerequisite
          recipe to build target

Most often a `target` is a file name except targets marked as PHONY (nearly all targets we have are PHONY targets).
`prerequisite`s can have prerequisites to assemble complex build structures .The command `make` reads all rules and chain the prerequisites to build the top target from bottom up.  

> Unfortunately, the code is really not easy to understand.

It depends on what you are used to, for me it is difficult to read YAML files ..We use make as build tool, there are other build tools, I know some from NPM or e.g github-workflows.  Compared to other tools `mak` is available everywhere and we can run it local without installing any infrastructure.

utils/makefile.sphinx
return42: Tex is a intermediate product of PDF generation / see https://github.com/searx/searx/pull/2530 and https://searx.github.io/searx/admin/buildhosts.html

    $ sudo -H ./utils/searx.sh install buildhost  # install Tex and more to build PDF
    $ make books/user.pdf

docs/conf.py
return42: For me it is a question of architecture.  This conf.py is a part of the documentation.  The rules to build documentation are defined in `./utils` [makefile.sphinx](https://github.com/searx/searx/blob/master/utils/makefile.sphinx) and [site-python/sphinx_build_tools.py](https://github.com/searx/searx/blob/master/utils/site-python/sphinx_build_tools.py) is a part of [makefile.sphinx](https://github.com/searx/searx/blob/master/utils/makefile.sphinx).

The makefile.python file does all the python stuff, one task is to make python implementations from the build tools available in build processes.  This is done by extending the `PYTHONPATH`

https://github.com/searx/searx/blob/006f206dc9ff30893cfb6e431c2ee3a7f7768973/utils/makefile.python#L6-L7

Makefile
return42: the tool pep8 does no longer exists, we use the successor 

https://github.com/searx/searx/blob/006f206dc9ff30893cfb6e431c2ee3a7f7768973/requirements-dev.txt#L4

Pycodestyle : identifies pep8 style violations.
Pylint : throws warnings or errors for a combination of programming and style issues.

_pycodestyle is a subset of pylint_

> pep8 is fast, and check typo (more than that).

Only simple typo checks .. [pylint](https://www.pylint.org/) does a deep inspection.

dalf:
@kvch  @asciimoo may I have your opinion on this PR ?

Makefile
dalf: I should I have written:
* test.pylint is slow
* test.pep8 (the Makefile target) is fast, and check typo ~(more than that)~.

It is better to fail fast for a typo, then check more in details.

return42:
> What is the intention of this PR 

If the abstraction of the boilerplate of makefiles is to complex, I recommend to implement the build tasks in bash scripts.

asciimoo:
> If the abstraction of the boilerplate of makefiles is to complex, I recommend to implement the build tasks in bash scripts.

That  used to be our workflow before makefiles, I support this.

asciimoo (approved):

return42:
> That used to be our workflow before makefiles, I support this.

One feature I will miss is the tab-completion which I am used to from the make command .. maybe it's just me because I'm lazy and use tab-completion 1000 times a day :)

return42:
I want to suggest to merge https://github.com/searx/searx/pull/2524 first, it fixes a issue we had discussed.

Afterwards I can implement a bash solution as a Drop-in replacement. Contains:

- bash script to run build tasks
- minmal `./Makefile` which uses this bash script and is compatible to our CI 
- drop `./utils/makefile.python`, `./utils/makefile.sphinx` (may be more `./utils/makefile.*`, lets see)
- update `./doc/makefile.rst`

kvch (approved):

kvch:
I like the general direction of the PR. Let me know when this becomes ready for review/merge.

return42:
I am working on an alternative to this PR .. this will take some days, I hope this weekend I will be finished and create a PR for.

return42:
@kvch  @dalf @asciimoo I send my alternative  #2594 to this PR.

In #2594 all build tasks are implemented in a bash script `./manage` and the `./Makefile` is just a wrapper for IDE and CI.  With the new make-implementations are reduced to a minimal complexity.

asciimoo:
Closing this because #2594 is merged. Thanks again @return42, great work!

PR #2531, opened 2021-02-03
[fix] Get correct locale with country from browser

### LINKED ISSUE #2525: Display wrong language in the search interface.

Just use it in the web browser.
**What happened?**

The UI language is wrong.
**How To Reproduce**

Open searx search page.
**Expected behavior**

See the right language.
**Screenshots & Logs**

圖片

**Additional context**

I'm using Traditional Chinese, but it displays the Simplified Chinese.
https://www.transifex.com/asciimoo/searx/dashboard/
It is fully translated to my language, but it still displays Simplified Chinese.
It also searches lots of websites from China, although I have selected Taiwan.

MarcAbonce:
## What does this PR do?

Some of our interface locales include uppercase country codes, which are separated by `_` instead of the more common `-`.
Also, a browser's `Accept-Language` header could be in lowercase.

This commit attempts to normalize those cases so a browser's language+country codes can better match with our locales.

1. Change your browser's `Accept-Language` header to add a country specific locale that we support (such as `pt-BR`, `nl-BE` or `zh-TW`). Note: This locale has to be the first one in the `Accept-Language` list.
2. Go to preferences.

Expected result: Correct locale is selected (`pt_BR` rather than `pt` and so on).

I also added a unit test to automate this.

## Author's checklist
This solution assumes that our UI locales have nothing more than language and optionally a country.
If we ever add a script specific locale like `zh_Hant_TW`* this would have to change to accommodate that, but the idea would be pretty much the same as this fix.

*By the way, in [our Transifex page](https://www.transifex.com/asciimoo/searx/languages/) it looks like `zh_Hant_TW` was removed (I assume it was merged with `zh_TW` which is already in traditional Chinese). Should we remove [this directory](https://github.com/searx/searx/tree/master/searx/translations/zh_Hant_TW/LC_MESSAGES) or am I making a wrong assumption here?

## Related issues
Closes #2525

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -142,6 +142,7 @@ babel = Babel(app)
 
 rtl_locales = ['ar', 'arc', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'fa_IR', 'glk', 'he',
                'ku', 'mzn', 'pnb', 'ps', 'sd', 'ug', 'ur', 'yi']
+ui_locale_codes = [l.replace('_', '-') for l in settings['locales'].keys()]
 
 # used when translating category names
 _category_names = (gettext('files'),
@@ -175,6 +176,9 @@ def _get_browser_or_settings_language(request, lang_list):
     for lang in request.headers.get("Accept-Language", "en").split(","):
         if ';' in lang:
             lang = lang.split(';')[0]
+        if '-' in lang:
+            lang_parts = lang.split('-')
+            lang = "{}-{}".format(lang_parts[0], lang_parts[-1].upper())
         locale = match_language(lang, lang_list, fallback=None)
         if locale is not None:
             return locale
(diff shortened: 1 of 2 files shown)


return42 (dismissed):
I made a slightly different test:

1. In the settings I added `de_AT : Deutsch (Austria)` to the _locales_ section.
2. In my browser I set `Accept-Language: de-at`

> Expected result: Correct locale is selected (pt_BR rather than pt and so on).

I tested this on branch `master` and on your `MarcAbonce:fix-browser-locale`.  Both with the same result; in the preferences the search-language is 'de-AT'.

With your patch applied, the only difference I see is:

branch `master` :

    DEBUG:searx.webapp:http://127.0.0.1:8888/preferences uses locale `de` from browser
    DEBUG:searx.webapp:http://127.0.0.1:8888/preferences uses locale `de` from preferences

branch `MarcAbonce:fix-browser-locale`

    DEBUG:searx.webapp:http://127.0.0.1:8888/ uses locale `de_AT` from browser
    DEBUG:searx.webapp:http://127.0.0.1:8888/ uses locale `de_AT` from preferences

To summarize: from my POV this patch fixes the issue, but I needed to check the debug log.

laichiaheng (dismissed):

kvch:
@MarcAbonce Please update the PR, so I can merge it.

kvch (dismissed):

laichiaheng (approved):

dalf (approved):
Thank you!

PR #2541, opened 2021-02-07
[enh] add engine MediathekViewWeb (API)

return42:
## What does this PR do?

Add new engine MediathekViewWeb using API from https://mediathekviewweb.de

     !mvw heute-show

### DIFF
--- a/searx/engines/mediathekviewweb.py
+++ b/searx/engines/mediathekviewweb.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""MediathekViewWeb (API)
+
+"""
+
+# pylint: disable=missing-function-docstring
+
+import datetime
+from json import loads, dumps
+
+about = {
+    "website": 'https://mediathekviewweb.de/',
+    "wikidata_id": 'Q27877380',
+    "official_api_documentation": 'https://gist.github.com/bagbag/a2888478d27de0e989cf777f81fb33de',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['videos']
+paging = True
+time_range_support = False
+safesearch = False
+
+def request(query, params):
+
+    params['url'] = 'https://mediathekviewweb.de/api/query'
+    params['method'] = 'POST'
+    params['headers']['Content-type'] = 'text/plain'
+    params['data'] = dumps({
+        'queries' : [
+	    {
+	        'fields' : [
+		    'title',
+		    'topic',
+	        ],
+	    'query' : query
+	    },
+        ],
+        'sortBy' : 'timestamp',
+        'sortOrder' : 'desc',
+        'future' : True,
+        'offset' : (params['pageno'] - 1 )* 10,
+        'size' : 10
+    })
+    return params
+
+def response(resp):
+
+    resp = loads(resp.text)
+
+    mwv_result = resp['result']
+    mwv_result_list = mwv_result['results']
+
+    results = []
+
+    for item in mwv_result_list:
+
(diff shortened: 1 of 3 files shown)


searx/engines/mediathekviewweb.py
```diff
@@ -0,0 +1,80 @@
...
+	    #     'query' : 'ndr'
+	    # },
+        ],
+        'sortBy' : 'timestamp',
+        'sortOrder' : 'desc',
+        'future' : False,
+        'offset' : 0,
+        'size' : 10
+    })
+    #logger.debug("POST --> %s" % params['url'] )
```
kvch: Please remove all commented out code lines.

kvch (commented):
Small request. Otherwise I've tested the engine and everything works well.

searx/engines/mediathekviewweb.py
```diff
@@ -0,0 +1,80 @@
...
+about = {
+    "website": 'https://mediathekviewweb.de/',
+    "wikidata_id": 'Q27877380',
+    "official_api_documentation": 'https://gist.github.com/bagbag/a2888478d27de0e989cf777f81fb33de',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['general']
```
kvch: I think this engine belongs to the category "videos".

return42: > I think this engine belongs to the category "videos".

Not really, it is a list of content from the public service media.  It consists of TV and radio productions and can be news, videos, podcast, shows ...

This list has nothing what is typical for videos, like Thumbnails etc.

dalf: In the general category, the expected result are HTML pages, so `item['url_website']` would be better than `item['url_video_hd']`.

But the interesting link is `item['url_video_hd']` which might by .mp4, .m3u8, and most probably other media extensions.
IMO, it should belongs to the "audio/videos" category (= what produce TV and radios : audio and videos), but this category doesn't exist 

The closer match is the "Videos" category, even it might return expected results : audio result in the video category.

dalf:
paging seems supported using

 ```
paging = True
....
'offset' : (params['pageno'] - 1 )* 10,
 ```

But the number of result is not always 10.

A similar implemetation: https://github.com/itouch5000/plugin.video.mediathekviewweb/blob/master/resources/lib/mediathekviewweb.py#L16

searx/engines/mediathekviewweb.py
return42: > The closer match is the "Videos" category, even it might return expected results : audio result in the video category.

This was also what I thought first, but then I realized that I can't use the video template since I do not have a thumb, author, publisher or date.  Listing the results of this engine in category "videos" looks very poor ..

And that was the point where I had a doubt about categories in general .. and I asked myself what will my users be expect?

Me and may be other german speaker will type e..g 'tatort' at google's common search and expect to find the TV movie:

- here is what you get using google https://www.google.de/search?q=tatort
- here is what you get  using searx: https://darmarit.org/searx/search?q=tatort&categories=general&language=de

For me it is hard to decide what fits better .. "videos" .. "general", "general" .. "videos" .. nothing fits well, it is (mixed audio/video) content for mainly german speakers ..

If you think it is better placed in video lets place it in video / I can configure 'general' as additional category in my instance.

What do you think?

return42: To shorten up .. I placed the engine in the category "videos".

return42:
> paging seems supported using

Thanks for the hint / is now implemented ..

return42:
@dalf / sorry for being impatient .. if you have time, could you have a look on the new implementation resolving your requests / thanks a lot.

PR #2429, opened 2020-12-28
Add Bandcamp search engine.

SuperSonicHub1:
Bandcamp engine
## What does this PR do?
It adds [Bandcamp](https://bandcamp.com) as a search engine to Searx. I mostly modeled my engine around the existing SoundCloud engine and looked to the Bing engine for some XPath help. It supports the parsing of URLs, titles, content, published dates, embeds, and thumbnails from Bandcamp albums, tracks, and singles.

## Why is this change important?
Bandcamp is a large music platform and I was surprised to see that no one had created an engine for it yet,

## How to test this PR locally?
The engine is completely compatible with `make test`. If you want to see the engine in action:
* `make run`
* search for `!bc zelda & chill` under the music category to test album support
* search for `!bc prototype remix tall black guy` under the music category to test track/single support

## Author's checklist
I could make the content field meatier by adding tags and album length, but I wanted to keep my first commit simple.

P.S.: `chore: Added self to README.` should actually be `chore: Added self to AUTHORS.rst.` My mistake.

PR #2543, opened 2021-02-07
Add Docker Hub search engine.

### LINKED ISSUE #2433: Engine request: DockerHub
**Working URL to the engine**
https://hub.docker.com/search?q=alpine&type=image

**Why do you want to add this engine?**
The largest Docker container registry

**Features of this engine**
N/A

**How can Searx fetch the information from this engine?**
The HTML result page or with the API: https://index.docker.io/v1/search?q=alpine&page=1 (`star_count` and `is_official` fields are nice to have besides the `name`)

Result entry landing pages:
- The `name` doesn't have a `/` (equivalent to `is_official=true` when using the API): https://hub.docker.com/_/alpine
- The `name` has a `/` (equivalent to `is_official=false` when using the API): https://hub.docker.com/r/alpine/git

**Applicable category of this engine**
IT

**Additional context**
N/A

SuperSonicHub1:
MySQL Docker image
Searx Docker image
## What does this PR do?
This adds [Docker Hub](https://hub.docker.com) to Searx as a search engine and resolves issue #2433. I modeled this engine around the existing SoundCloud engine, the engine's favicon was downloaded from the Docker Hub website with `wget` and converted to a PNG with ImageMagick, and the shortcut was plucked from DDG's bangs database. It supports the parsing of URLs, titles, content, published dates, and thumbnails of Docker images.

Once again, I wanted to resolve issue #2433. As a semi-regular user of the Hub myself, I'm sure many developers would find the direct ability to search for Docker images very convenient. 

This PR is fully compatible with `make test`. To see it in action:
* `make run`
* query `!dockerhub searx`

## Related issues
Closes #2433


.gitignore
```diff
@@ -26,3 +26,5 @@ dist/
 local/
 gh-pages/
 searx.egg-info/
+
+.vscode/
```
kvch, line 30: Please do not add any editor specific line to this file.

searx/settings.yml
```diff
@@ -309,6 +309,11 @@ engines:
   - name : digg
     engine : digg
     shortcut : dg
+  
+  - name: docker hub
+    engine: docker_hub
+    shortcut: dockerhub
```
kvch, line 315: Do you mind coming up with a shorter shortcut? :D Maybe dh or drh? Or you think using the same "longcut" as DDG is valuable?

searx/engines/docker_hub.py
```diff
@@ -0,0 +1,69 @@
...
+Docker Hub (IT)
+
+@website     https://hub.docker.com
+@provide-api yes (https://docs.docker.com/registry/spec/api/)
+
+@using-api   yes
+@results     JSON
+@stable      yes
+@parse       url, title, content, thumbnail, publishedDate
+"""
```
return42, line 11: Please use 'about' notation / like any other engine does, here is an example from the google engine:

https://github.com/searx/searx/blob/ab8739809c8c5f4c1f12e4b15fa8c61afe30ef9f/searx/engines/google.py#L22-L29

searx/engines/docker_hub.py
```diff
@@ -0,0 +1,69 @@
...
+def request(query, params):
+    '''pre-request callback
+    params<dict>:
+      method  : POST/GET
+      headers : {}
+      data    : {} # if method == POST
+      url     : ''
+      category: 'search category'
+      pageno  : 1 # number of the requested page
+    '''
```
return42, line 33: I guess these doc-strings can be dropped .. if there is something special to this engine, it is good to write down a doc-string, but if it is only C&P better leaf it.

return42:
@SuperSonicHub1 thanks a lot for another engine!!  ..I really like your contributions, but there is one thing that disturbs me: it is the commit history.  In the master branch we want to avoid a cluttered history.  It is very hard to maintain code if the history is cluttered.  Can you please squash your commits to one and have one clear commit message / thanks!

In https://github.com/searx/searx/pull/2429#issuecomment-752412734 I gave you some links about git-workflows at hand.

I guess the cluttered commit series is also one reason why #2429 hangs at the moment.

searx/engines/docker_hub.py
dalf: The wikidata_id is Q100769064 : https://www.wikidata.org/wiki/Q100769064

.gitignore
SuperSonicHub1: OK. How do you suggest I stop my config from being pushed without manually staging each of my commits?

searx/settings.yml
SuperSonicHub1: I think maintaining parity with DDG is important, and `dh` and `drh` both seem pretty unintuitive to me.

.gitignore
return42: > without manually staging each of my commits?

what should that tell me?!? Please learn git and SCM basics .. to me it seems you won't learn for what all this is good.

dalf: > How do you suggest I stop my config from being pushed without manually staging each of my commits?

See https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

unixfox:
Hello,

Could you squash your commits to a single commit? 
Here is a tutorial about how to do it: https://blog.carbonfive.com/always-squash-and-rebase-your-git-commits/. It's not the only one, so feel free to learn from other tutorials if you want.

return42:
@SuperSonicHub1 see my comment at : https://github.com/searx/searx/pull/2429#issuecomment-808864844

PR #2544, opened 2021-02-07
[Engine] Add Library of Congress engine

mrwormo:
## What does this PR do?

Add https://www.loc.gov/pictures as a new source for images.

I dunno if it's important, but it's a great source of historical images.

Make run and initiate a search.

### DIFF
--- a/searx/engines/loc.py
+++ b/searx/engines/loc.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+
+ Library of Congress : images from Prints and Photographs Online Catalog
+
+"""
+
+from json import loads
+from urllib.parse import urlencode
+
+
+about = {
+    "website": 'https://www.loc.gov/pictures/',
+    "wikidata_id": 'Q131454',
+    "official_api_documentation": 'https://www.loc.gov/pictures/api',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['images']
+
+paging = True
+
+base_url = 'https://loc.gov/pictures/search/?'
+search_string = "&sp={page}&{query}&fo=json"
+
+IMG_SRC_FIXES = {
+    'https://tile.loc.gov/storage-services/': 'https://tile.loc.gov/storage-services/',
+    'https://loc.gov/pictures/static/images/': 'https://tile.loc.gov/storage-services/',
+    'https://www.loc.gov/pictures/cdn/': 'https://tile.loc.gov/storage-services/',
+}
+
+
+def request(query, params):
+
+    search_path = search_string.format(
+        query=urlencode({'q': query}),
+        page=params['pageno'])
+
+    params['url'] = base_url + search_path
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    json_data = loads(resp.text)
+
+    for result in json_data['results']:
+        img_src = result['image']['full']
+        for url_prefix, url_replace in IMG_SRC_FIXES.items():
+            if img_src.startswith(url_prefix):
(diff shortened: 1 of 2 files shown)


searx/settings.yml
```diff
@@ -698,6 +698,11 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name : library of congress
+    engine : loc
```
return42: Can you remove the trailing whitespace / thanks.

searx/engines/loc.py
```diff
@@ -0,0 +1,51 @@
...
+    return params
+
+
+def response(resp):
+    results = []
+
+    json_data = loads(resp.text)
+
+    for result in json_data['results']:
+        results.append({'url': result['links']['item'],
```
return42: Aligned with opening delimiter is PEP-8 conform, but I wanted to ask if we  can we use hanging indents which is most often better to read:

```python
    for result in json_data['results']:
        results.append({
            'url': result['links']['item'],
            'title': result['title'],
            'img_src': result['image']['thumb'],
            'template': 'images.html'
        })
```

return42 (changes requested):
LGTM and works .. only some nits I have ...

return42 (dismissed):
LGTM / thanks for the fixes!

searx/engines/loc.py
mrwormo: I agree, it's easier to read.

dalf:
The JSON API returns broken URL for `result['image']['full']` but it is possible to fix the URL looking at the HTML page.

Here a patch, include also `results['author'] = result['creator']` :

```diff
diff --git a/searx/engines/loc.py b/searx/engines/loc.py
index a1b4f310..5c09ceff 100644
--- a/searx/engines/loc.py
+++ b/searx/engines/loc.py
@@ -25,6 +25,12 @@ paging = True
 base_url = 'https://loc.gov/pictures/search/?'
 search_string = "&sp={page}&{query}&fo=json"
 
+IMG_SRC_FIXES = {
+    'https://tile.loc.gov/storage-services/': 'https://tile.loc.gov/storage-services/',
+    'https://loc.gov/pictures/static/images/': 'https://tile.loc.gov/storage-services/',
+    'https://www.loc.gov/pictures/cdn/': 'https://tile.loc.gov/storage-services/',
+}
+
 
 def request(query, params):
 
@@ -43,10 +49,19 @@ def response(resp):
     json_data = loads(resp.text)
 
     for result in json_data['results']:
+        img_src = result['image']['full']
+        for url_prefix, url_replace in IMG_SRC_FIXES.items():
+            if img_src.startswith(url_prefix):
+                img_src = img_src.replace(url_prefix, url_replace)
+                break
+        else:
+            img_src = result['image']['thumb']
         results.append({
             'url': result['links']['item'],
             'title': result['title'],
-            'img_src': result['image']['thumb'],
+            'img_src': img_src,
+            'thumbnail_src': result['image']['thumb'],
+            'author': result['creator'],
             'template': 'images.html'
         })
 
```

----

`'publishedDate' : result['created_published_date']` won't work because `publishedDate` must have this format `%Y-%m-%d %H:%M:%S%z`

https://github.com/searx/searx/blob/e4cc7f13a397732c519f28a749b780bc32755522/searx/webapp.py#L641-L644

mrwormo:
> The JSON API returns broken URL for `result['image']['full']` but it is possible to fix the URL looking at the HTML page.
> 
> Here a patch, include also `results['author'] = result['creator']` :

Clever ! I didn't find the solution for that... 

> `'publishedDate' : result['created_published_date']` won't work because `publishedDate` must have this format `%Y-%m-%d %H:%M:%S%z`

Yeah, i saw 'publishedDate' was not in the right format...

Many thanks @dalf. One more time, i've learned something :+1:

PR #901, opened 2017-04-19
ignore double-quotes when highlighting query parts

danielhones:
This is to fix a minor UI bug where using a search with an exact phrase in quotes wouldn't highlight the matching phrase in the search results.  For example:

search: `master of none`
result: **master of none**

search: `"master of none"`
result: master **of** none

PR #2553, opened 2021-02-09
Ignore double-quotes when highlighting query parts

danielhones:
## What does this PR do?

This is to fix a minor UI bug where using a search with an exact phrase in quotes wouldn't highlight the matching phrase in the search results.  This replaces #901 which I'll close.

For example:

search: `master of none`
result: **master of none**

search: `"master of none"`
result: master **of** none
 
The cause of this bug is that a query like `"master of none"` gets split into `['"master', 'of', 'none"']`, and the code tries to highlight strings like `none"` which aren't typically in the text.  This change works by ignoring double quotes in the query.  I added test coverage for other paths through the `webutils.highlight_content` function.  Currently there are no tests that cover the scenario where there needs to be `span` elements inserted in the result.

Without this change, only the interior parts of a phrase in quotes will be highlighted.  In many cases that might be the least important words of the phrase and it's surprising behavior to a user.

The unit tests for this function pass and to test manually, you can search for several queries, first without quotation marks, and then with them, and the highlighted words in the results should be the same.

## Author's checklist

I considered pulling the existing test cases in to the new `data` tuple in the unit test, but I decided it was useful to keep them separate because they're all testing edge cases or cases where no highlights would be inserted.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -154,6 +154,6 @@ generally made searx better:
 - @mrwormo
 - Xiaoyu WEI @xywei
 - @joshu9h
-
+- Daniel Hones
 
 
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -119,7 +119,10 @@ def highlight_content(content, query):
     else:
         regex_parts = []
         for chunk in query.split():
-            if len(chunk) == 1:
+            chunk = chunk.replace('"', '')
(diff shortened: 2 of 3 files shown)


return42 (approved):
Perfect / thanks!

dalf (approved):
Thank you !

Issue #2537, opened 2021-02-06
It displays Simplified Chinese results if I select "Default"


laichiaheng:
**Version of Searx, commit number if you are using on master branch and stipulate if you forked Searx**

**How did you install Searx?**

**What happened?**

If I don't choose zh-TW but choose Default, it shows me lots of Simplified Chinese and Traditional Chinese results.
However, the default language of my browser is zh-TW.
**How To Reproduce**

Search something without selecting searching language.
**Expected behavior**

It shows the results in default language which is zh-TW
**Screenshots & Logs**

All Traditional Chinese if I select zh-TW
![zh-tw](https://user-images.githubusercontent.com/9532105/107113321-46150600-6899-11eb-9378-a5a4056e4aa4.png)
There are Simplified Chinese if I select Default
![default](https://user-images.githubusercontent.com/9532105/107113322-47463300-6899-11eb-9b4e-2f0c3bdc85ea.png)

**Additional context**

dalf:
I'm not sure why the issue is closed.

I don't speak / write Chinese, so I'm may wrong: car is 汽車 in traditional chinese, and 汽车  in simplified chineese (?)

If I search for 汽車 using the zh-TW language preference:
* google results contain mostly 汽車 and some with 汽车. It seems google.com.tw behave in the same way (???). If it is the case, nothing can be done.
* wikipedia & wikidata return results about 汽车. If I understand correctly, to get the traditional version of the article, the link should be
	* https://zh.wikipedia.org/zh-hant/汽車 or https://zh.wikipedia.org/zh-hant/汽车
	* and not https://zh.wikipedia.org/wiki/汽车 as it is the case now.

Here most probably, something can be done.

ping @MarcAbonce

unixfox:
Could also be an issue from the engine, in this case Google.

laichiaheng:
Google sucks.....

MarcAbonce:
I just made #2554 for the Wikipedia issue.

About the "Default" language option, the problem is that it simply doesn't send any language preference to the engine, so the engine will return results in whatever languages it wants.

PR #2554, opened 2021-02-09
Add support for Chinese variants in Wikipedia

MarcAbonce:
## What does this PR do?

Show Wikipedia's infobox in the regional Chinese script specified in the query.

Search for `!wp 出租車` with each of the available Chinese options either from the language dropdown or with a `:` bang.

Expected results:
* `中文 - zh` should show 出租車.
* `中文 (中国) - zh-CN` should show 出租车.
* `中文 (台灣) - zh-TW` should show 計程車.
* `:zh-HK` should show 的士.
* `:zh-SG` should show 德士.

The Wikipedia link returned by the API is still the same in all cases (https://zh.wikipedia.org/wiki/%E5%87%BA%E7%A7%9F%E8%BB%8A) but if your browser's `Accept-Language` is set to any of the above, Wikipedia automatically returns the desired script in their webpage.

## Note

Actually, this problem affects other languages like Serbian as well (see: https://sr.wikipedia.org/sr-ec/Аутомобил vs. https://sr.wikipedia.org/sr-el/Automobil) but in that case the difference can't be determined by a country code but only by a script code and we currently don't support those anyway.

## Related issues
#2537

### DIFF
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -22,6 +22,7 @@ about = {
 # search-url
 search_url = 'https://{language}.wikipedia.org/api/rest_v1/page/summary/{title}'
 supported_languages_url = 'https://meta.wikimedia.org/wiki/List_of_Wikipedias'
+language_variants = {"zh": ("zh-cn", "zh-hk", "zh-mo", "zh-my", "zh-sg", "zh-tw")}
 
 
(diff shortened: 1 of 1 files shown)


dalf (approved):
Thank you !

Does the same patch applied to the wikidata engine ?

Issue #2259, opened 2020-10-10
duckduckgo: the language preference is sometime ignored


dalf:
The duckduckgo engine has been updated: #2254

But the language support still have a problem: it works if there is no more than one request per minute, otherwise it is ignored (it seems to pick the Russian language ?)

But it seems to work with curl:
```sh
curl --http1.1 'https://html.duckduckgo.com/html/' -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:81.0) Gecko/20100101 Firefox/81.0' --data-raw 'q=test&b=&kl=fr-fr&df=' | grep "leconjugueur.lefigaro.fr"
```

 so most probably I'm missing something ?

https://github.com/searx/searx/blob/e78bfd4d8b62a521e08029bb0e708b2bcec9958d/searx/engines/duckduckgo.py#L53-L82

CaffeinatedTech:
I'm seeing so many Russian results lately.  It's almost unusable.

MarcAbonce:
I was able to reproduce this with curl:
`curl --http1.1 'https://html.duckduckgo.com/html/' -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:81.0) Gecko/20100101 Firefox/81.0' --data-raw 'q=beatles&b=&kl=fr-fr&df=' | grep "и"`
I get Russian results the second time.

Strangely enough, I could even reproduce this bug directly with the UI in https://html.duckduckgo.com/html.

dalf:
> Strangely enough, I could even reproduce this bug directly with the UI in https://html.duckduckgo.com/html.

I have seen that `https://duckduckgo.com/t/sl_h` is blocked by ublock origin, so here a patch that call this URL with the same headers, it seems better, but I can't confirm.

```patch
diff --git a/searx/engines/duckduckgo.py b/searx/engines/duckduckgo.py
index c1c98462..58c361de 100644
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -15,7 +15,8 @@
 
 from lxml.html import fromstring
 from json import loads
-from searx.utils import extract_text, match_language, eval_xpath
+from searx.utils import extract_text, match_language, eval_xpath, dict_subset
+from searx.poolrequests import get
 
 # engine dependent config
 categories = ['general']
@@ -36,6 +37,7 @@ language_aliases = {
 
 # search-url
[...]

PR #2560, opened 2021-02-09
Fix duckduckgo

dalf:
## What does this PR do?

poolrequests: for one (user request, engine) always use the same HTTPAdapter

The duckduckgo engine requires an additional request after the results have been sent.
This commit makes sure that the second request uses the same HTTPAdapter
= the same IP address, and the same proxy.

----

duckduckgo engine: better support of the language preference
After the main request, send a second to https://duckduckgo.com/t/sl_h

It improve the user language support.
This is empirical : when there is this additional request, it seems better.
See https://github.com/searx/searx/issues/2259

Check that the duckduckgo engine respect the user language preference

Check with multiple requests and from different IP.

## Related issues

Close (?) : https://github.com/searx/searx/issues/2259

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -5,7 +5,8 @@
 
 from lxml.html import fromstring
 from json import loads
-from searx.utils import extract_text, match_language, eval_xpath
+from searx.utils import extract_text, match_language, eval_xpath, dict_subset
+from searx.poolrequests import get
 
 # about
 about = {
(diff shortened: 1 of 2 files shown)


return42 (approved):
I can''t test multiple IPs but the ddg ping helps to improve result lists from ddg.  Nevertheless, I have the impression that the ddg has more quirks than just the ping. But this needs more investigation for what I do not have time ATM.

For me this PR is a big improvement and I would vote to merge / approved! ... thanks!!

PR #2561, opened 2021-02-10
Metrics & processors refactoring

dalf:
## What does this PR do?

* Move metrics & statistics code into a dedicate module : `searx.metrics`
* No additional dependency / no additional exposed data.
* The interface is based on https://github.com/prometheus/client_python : a not written plugin can send the metrics or provide plot like in https://github.com/searx/searx/pull/2523
* `/stats` keeps the same layout.
* Use `timeit.default_timer` instead of `time.time` (more accurate)

* `searx.search.processors.abstract.EngineProcessor` contains additional methods:
    * `is_suspended` (property) with `suspend_reason`, later another PR can use these properties to show to the user that an engine is suspend and why.
    * `_suspend`
    * `_resume`
    * `_handle_exception(self, result_container, reason, exception, suspend=False)`
    * `_extend_container_basic(self, result_container, start_time, search_results)`
* the code for the actual processors becomes straightforward
* an offline engine can raise `SearxEngineAccessDeniedException` to suspend the engine for a time

In the offline engine, the ValueError exception is not recorded as an error: the engine can raise this exception when the query is invalid.

* Refactor the code in one place, so it is easy to add some additional stats / export the stats / easy to hack
* `searx.engines.engines`  contains only engine definition, not live variables.
* The processors becomes easier to write.

* starts searx
* search for some queries
* see `/stats`


dalf:
useless code sorry

Issue #2146, opened 2020-08-14
text encoding problem (cyrilic)


sashagra:
Ubuntu 64-bit, Firefox 79
![Screenshot_2020-08-14 МАПИд - searx](https://user-images.githubusercontent.com/45414622/90259670-14276780-de53-11ea-9c28-59dc58e1ec6d.png)

asciimoo:
@sashagra could you provide a test query in text form and any special settings required to reproduce this issue?

sashagra:
@asciimoo 
It is any Cyrillic query. There are no special settings
Test query: МАПИД

sashagra:
Now I see some difference with request from address bar and from web page. Also I get other results in the private window where is other style of results page
I get this issue with address bar request and 
![image](https://user-images.githubusercontent.com/45414622/95762188-09a01780-0cb6-11eb-9250-f4b28a514c67.png)

unixfox:
Is this still an issue?

unixfox:
@sashagra I tried your test query, but I couldn't replicate the bug.

Reopen if you are still experiencing this issue.

PR #2564, opened 2021-02-11
[fix] fix seznam engine

dalf:
## What does this PR do?

Fix the seznam engine

In the master branch, the engine returns no result

* `!seznam time`
* `!seznam seznam`
* `!seznam поисковая система`

Note: this PR adds an additional request when searx starts:
* the `request` function send a HTTP request to get an id (the id seems to change each time).
* [`searx.engines._set_https_support_for_engine`](https://github.com/searx/searx/blob/436d366448131088a9cddcedc9b789a44b38ee97/searx/engines/__init__.py#L305) calls the `request` function to detect HTTP usage.

## Related issues

Related to https://github.com/searx/searx/issues/2146 : make sure the HTML is decoded as UTF-8 using `resp.content.decode()` instead of [`resp.text` (in the xpath engine)](https://github.com/searx/searx/blob/436d366448131088a9cddcedc9b789a44b38ee97/searx/engines/xpath.py#L42)

### DIFF
--- a/searx/engines/seznam.py
+++ b/searx/engines/seznam.py
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+ Seznam
+"""
+
+from urllib.parse import urlencode, urlparse
+from lxml import html
+from searx.poolrequests import get
+from searx.exceptions import SearxEngineAccessDeniedException
+from searx.utils import extract_text, eval_xpath_list, eval_xpath_getindex
+
+# about
+about = {
+    "website": "https://www.seznam.cz/",
+    "wikidata_id": "Q3490485",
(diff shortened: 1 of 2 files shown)


fgossel:
_!seznam поисковая система_
CZ language doesn't use Cyrillic. Make no sense such a test.

return42:
> CZ language doesn't use Cyrillic. Make no sense such a test.

I guess seznam.cz has also russian results? (similar to google.com)

We are a very small team with very limited skills in internationalization and need  support in russian, czech and many other languages. If you have any hints or criticism for us you are very welcome. Really, we need help from international contributors!

fgossel:
Cool, seznam.cz with the latest fix seems to work correctly now. 

!szn Příliš žluťoučký kůň úpěl ďábelské ódy

PR #993, opened 2017-08-12
[mod] update currencies.json and fetch_currencies.py

dalf:
use a sparql request on wikidata to get the list of currencies.

currencies.json contains the translation for all supported searx languages.

PR #2585, opened 2021-02-19
[mod] update currencies.json and fetch_currencies.py

dalf:
## What does this PR do?

use a sparql request on wikidata to get the list of currencies.

currencies.json contains the translation for all supported searx languages.

In the master branch, `utils/fetch_currencies.py` doesn't work anymore, so it is not possible to update `searx/data/currencies.json`.

* `searx-checker currency`
* search for `1 USD in €`

## Related issues

* Supersede #993
* Related to #2052
* Related to https://github.com/searx/searx/wiki/Milestones#milestone-110---data-upgrade--build-process

### DIFF
--- a/utils/fetch_currencies.py
+++ b/utils/fetch_currencies.py
@@ -1,163 +1,151 @@
-# -*- coding: utf-8 -*-
+#!/usr/bin/env python
 
-
-import json
 import re
 import unicodedata
-import string
-from urllib.parse import urlencode
-from requests import get
+import json
 
-languages = {'de', 'en', 'es', 'fr', 'hu', 'it', 'nl', 'jp'}
+# set path
+from sys import path
+from os.path import realpath, dirname, join
+path.append(realpath(dirname(realpath(__file__)) + '/../'))
 
-url_template = 'https://www.wikidata.org/w/api.php?action=wbgetentities&format=json&{query}&props=labels%7Cdatatype%7Cclaims%7Caliases&languages=' + '|'.join(languages)
-url_wmflabs_template = 'http://wdq.wmflabs.org/api?q='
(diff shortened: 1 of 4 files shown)


utils/fetch_currencies.py
```diff
@@ -1,163 +1,139 @@
...
 
-# static
-add_currency_name("euro", 'EUR')
-add_currency_name("euros", 'EUR')
-add_currency_name("dollar", 'USD')
-add_currency_name("dollars", 'USD')
-add_currency_name("peso", 'MXN')
-add_currency_name("pesos", 'MXN')
+    with open(get_filename(), 'w', encoding='utf8') as f:
+        json.dump(db, f, ensure_ascii=False, separators=(',', ':'))
```
return42: As long as we commit data files in the SCM we should make them [diffable](https://en.wiktionary.org/wiki/diffable), I would suggest. Here is my patch:
```diff
@@ -133,7 +133,7 @@ def main():
     add_currency_name(db, "pesos", 'MXN')
 
     with open(get_filename(), 'w', encoding='utf8') as f:
-        json.dump(db, f, ensure_ascii=False, separators=(',', ':'))
+        json.dump(db, f, ensure_ascii=False, separators=(',', ':'), sort_keys=True, indent=2)
 ```

I choose a indent of 2 to save some space, but 4 is more common / delta is about ~70kB.  But I don't know if we should care about the size (compare `searx/data/bangs.json`)

return42 (dismissed):
> use a sparql request on wikidata to get the list of currencies.

good job done! .. just a nit about diffable json files in SCM .. approved!

return42 (approved):

utils/fetch_currencies.py
dalf: Updated.

PR #2592, opened 2021-02-22
[mod] add utils/fetch_external_bangs.py

dalf:
## What does this PR do?

Based on duckduckgo bangs
Store bangs on a trie to allow autocomplete (not in this commit)

Searx loads `searx/data/external_bangs` without data structure conversion.

In the master branch the external bangs:
* are never updated.
* are based on a 2 years list from jivesearch: https://github.com/jivesearch/jivesearch/tree/master/bangs
* the language support is implemented

* use some external bangs, for example: ``!!ddg time`

## Related issues

Related to https://github.com/searx/searx/wiki/Milestones#milestone-110---data-upgrade--build-process


utils/fetch_external_bangs.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
+
```
return42, line 15: Since there is a shebang, I would suggest to set file mode to 755.

OT: there are some trailing whitespace which can be dropped:

```diff
@@ -22,7 +22,7 @@ RE_BANG_VERSION = re.compile('\/bang\.v([0-9]+)\.js')
 def get_bang_url():
     response = requests.get(URL_BV1)
     response.raise_for_status()
-    
+
     r = RE_BANG_VERSION.findall(response.text)
     return f'https://duckduckgo.com/bang.v{r[0]}.js', r[0]
 
@@ -36,7 +36,7 @@ def fetch_bangs(url):
 def merge_when_no_leaf(node):
     """Minimize the number of nodes
 
-    A -> B -> C 
+    A -> B -> C
     B is child of A
```

utils/fetch_external_bangs.py
```diff
@@ -0,0 +1,139 @@
...
+# set path
+sys.path.append(realpath(dirname(realpath(__file__)) + '/../'))
+
+from searx import searx_dir
+
+
+# from https://duckduckgo.com/newbang
+URL_BV1 = 'https://duckduckgo.com/bv1.js'
+
+RE_BANG_VERSION = re.compile('\/bang\.v([0-9]+)\.js')
```
return42: The string of the regexp contains anomalous backslashes .. it is better to use raw strings for regexp (`r'\/bang\.v([0-9]+)\.js'`). BTW it easy to find issues like that with pylint.  Here is my patch to do a pylint:

```diff
diff --git a/Makefile b/Makefile
index 13f965b6..fbf5f94b 100644
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,8 @@ PYLINT_FILES=\
        searx/engines/google_news.py \
        searx/engines/google_videos.py \
        searx/engines/google_images.py \
-       searx/engines/mediathekviewweb.py
+       searx/engines/mediathekviewweb.py \
+       utils/fetch_external_bangs.py
 
 test.pylint: pyenvinstall
        $(call cmd,pylint,$(PYLINT_FILES))
diff --git a/utils/fetch_external_bangs.py b/utils/fetch_external_bangs.py
old mode 100644
new mode 100755
index 99f820a9..72155cba
--- a/utils/fetch_external_bangs.py
+++ b/utils/fetch_external_bangs.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# pylint: disable=C0116
+

```

With `make test.pylint` I got this messages::

```pylint
utils/fetch_external_bangs.py:21:0: W1401: Anomalous backslash in string: '\/'. String constant might be missing an r prefix. (anomalous-backslash-in-string)
utils/fetch_external_bangs.py:21:6: W1401: Anomalous backslash in string: '\.'. String constant might be missing an r prefix. (anomalous-backslash-in-string)
utils/fetch_external_bangs.py:21:17: W1401: Anomalous backslash in string: '\.'. String constant might be missing an r prefix. (anomalous-backslash-in-string)
utils/fetch_external_bangs.py:27:0: C0303: Trailing whitespace (trailing-whitespace)
utils/fetch_external_bangs.py:41:15: C0303: Trailing whitespace (trailing-whitespace)
utils/fetch_external_bangs.py:1:0: C0114: Missing module docstring (missing-module-docstring)
utils/fetch_external_bangs.py:15:0: C0413: Import "from searx import searx_dir" should be placed at the top of the module (wrong-import-position)
utils/fetch_external_bangs.py:89:16: W0621: Redefining name 'bangs' from outer scope (line 135) (redefined-outer-name)
```

return42 (changes requested):
LGTM but I would request small changes ... and give pylint a try ... especially when we create new files.

searx/external_bang.py
```diff
@@ -1,14 +1,73 @@
...
+
+def get_bang_definition_and_autocomplete(bang):
+    bang_definition, bang_ac_list = get_bang_definition_and_ac(bang)
+
+    new_autocomplete = []
+
+    current = [*bang_ac_list]
+    done = set()
+    while len(current) > 0:
+        bang_ac = current[0]
```
kvch: Nit:
```python
bang_ac = current.pop(0)
```
seems a bit simpler.

searx/external_bang.py
```diff
@@ -1,14 +1,73 @@
...
+        del current[0]
+        done.add(bang_ac)
+
+        bang_definition, bang_ac_list = get_bang_definition_and_ac(bang_ac)
```
kvch: Nit: `bang_definition` and `bang_ac_list` shadows the variable from the outer scope. I would prefer that you renamed these to `current_bang_definition` and `current_bang_ac_list`. This way the code becomes more readable and we do not accidentally make mistakes if/when these variables are renamed.

searx/external_bang.py
```diff
@@ -1,14 +1,73 @@
...
+        bang_ac_list = [before + k for k in node.keys() if k != '*']
+    elif isinstance(node, str):
+        bang_definition = node
+        bang_ac_list = []
+
+    return bang_definition, bang_ac_list
+
+
+def resolve_bang_definition(bang_definition, query):
+    result = bang_definition.split(chr(1))
```
kvch: Please either introduce meaningful index names when accessing the members of this list or give variables more descriptive names.

Example 1:
```python
URL = 0
result[URL] = result[URL].replace(chr(2), query)
```

Or alternatively
```python
(url, order) = bang_definition.split(chr(1))
# ...
return url, order
```

searx/external_bang.py
```diff
@@ -1,14 +1,73 @@
...
-# https://pseitz.github.io/toml-to-json-online-converter/
-# NOTE only use the get_bang_url
+from searx.data import EXTERNAL_BANGS
 
-bangs_data = {}
-for bang in bangs_loader()['bang']:
-    for trigger in bang["triggers"]:
-        bangs_data[trigger] = {x: y for x, y in bang.items() if x != "triggers"}
+
+def get_node(bang):
```
kvch: Please pass `EXTERNAL_BANGS` as a parameter so these functions can be tested with a smaller list of test bangs.

utils/fetch_external_bangs.py
kvch: :+1: for making it executable.

utils/fetch_external_bangs.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # bang_list
+        bang_url = bang_definition['u']
+        if '{{{s}}}' not in bang_url:
+            # ignore invalid bang
+            continue
+
+        bang_url = bang_url.replace('{{{s}}}', chr(2))
+        if bang_url.startswith('https://'):
+            bang_url = bang_url[6:]
```
kvch: Are you removing `https:` here? IMHO getting rid of the magic number makes it cleaner:
```python
if bang_url.startswith('https:'):
    bang_url = bang_url[len('https:'):]
```

utils/fetch_external_bangs.py
```diff
@@ -0,0 +1,139 @@
...
+    bang_trie = {}
+    bang_urls = {}
+
+    for bang_definition in bangs:
+        # bang_trie
+        bang = bang_definition['t']
+
+        # bang_list
+        bang_url = bang_definition['u']
+        if '{{{s}}}' not in bang_url:
```
kvch, line 110: Nit: you can get rid of the comment by refactoring the code

```python

def _is_invalid_bang_url(bang_url):
    return '{{{s}}}' not in bang_url
```
then call it here:
```python
if _is_invalid_bang_url(bang_url):
    continue
```

kvch (changes requested):
Great PR! Thank you for this. I have added a few minor comments. Addressing them would make the code a bit more readable, but it is not necessary. 

However, most of these functions could be tested if `EXTERNAL_BANGS` was passed as a parameter. Please refactor the code and add a few tests.

utils/fetch_external_bangs.py
return42: For me the existing implementation is more readable .. and function calls are (minimal) more time consuming .. but I would have another nit :-) ..  I would place cancellation criteria at the beginning  (checking `bang_url` before eval `bang` in line 93).

utils/fetch_external_bangs.py
return42: I agree that your suggestion is more readable but on the other hand, calculating a constant value each time in a loop is time consuming .. I would suggest to set a comment like:  _removing "https:" from url_ (same in line 109 with "http:")

utils/fetch_external_bangs.py
dalf: I moved the code at the top of the loop, so the pattern is obvious

utils/fetch_external_bangs.py
dalf: It is now executable, and pylint checks this file.

searx/external_bang.py
dalf: Done: `url, order = bang_definition.split(chr(1))`

utils/fetch_external_bangs.py
dalf: There are two global variables HTTPS_COLON and HTTP_COLON.

Since the script is executed only from time to time, I haven't optimize the run time.

There is also a comment.

dalf:
> However, most of these functions could be tested if EXTERNAL_BANGS was passed as a parameter. Please refactor the code and add a few tests.

Done, there is also a new file: `tests/unit/test_external_bangs.py`.

searx/external_bang.py
dalf: Done. 
This signature is `get_node(external_bangs_db, bang):`

utils/fetch_external_bangs.py
dalf: Done

kvch (approved):
Thanks for the quick turnaround!

return42:
I have a question about the sorting of bangs.  By example, in [data/external_bangs.json](https://github.com/searx/searx/blob/master/searx/data/external_bangs.json) we have an entry named _Blockchain'_:

```json
    {
      "name": "Blockchain",
      "triggers": [
        "blockchain",
        "bc",
        "bitcoin",
        "bchain",
        "btcaddr",
        "btc"
      ],
      "regions": {
        "default": "https://blockchain.info/search?search={{{term}}}/"
      }
    },
```

In the origin bang.v620.js from ddg there is only one entry with URL `https://blockchain.info/search?search=...`:

```json
    {
        "c": "Online Services",
        "u": "https://blockchain.info/search?search={{{s}}}",
        "sc": "Tools",
        "s": "Blockchain.info",
        "d": "blockchain.info",
        "r": 6,
        "t": "bitcoin"
    },
```

Other triggers like `btcaddr` do use a different URL scheme

```json
    {
        "d": "www.blockchain.com",
        "t": "btcaddr",
        "r": 4,
        "s": "Blockchain",
        "sc": "Cryptocurrency",
        "c": "Tech",
        "u": "https://www.blockchain.com/btc/address/{{{s}}}"
    },
```

Is it OK to assemble all triggers under the same search URL? .. or do I miss something?

Issue #2240, opened 2020-10-02
searx.utils.is_valid_lang & searx.query.RawTextQuery: redundant code ?


dalf:
This two following peace of code do the same thing in different way: check if a string describes a language defined [searx.language.language_codes](https://github.com/searx/searx/blob/master/searx/languages.py)

https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/utils.py#L167-L181

https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/query.py#L92-L112

Different questions:
* Could it be one function ?
* Could the code in [autocomplete.py](https://github.com/searx/searx/blob/ecb9f28869f081dc75182053c727523ed7e20d12/searx/autocomplete.py#L84-L100) be merged too ?
* wouldn't be more efficient and simplier to build on runtime a dict with:
	* key : lower case of `lang_id`, `lang_name`, `country`, `english_name`
	* value: (`lang_id`, `lang_name`, `country`, `english_name`)
	?
	
ping @MarcAbonce

MarcAbonce:
Apparently, `is_valid_lang` is only used in two translation engines (https://github.com/searx/searx/search?q=is_valid_lang). Ideally, both engines should match against their own language lists rather than our own `languages.py` which would allow us to remove `is_valid_lang`. I'll try to get that on the weekend. Probably something like e1d9ddc but hopefully less annoying this time.

The autocomplete code is definitely duplicated code. In fact, I think most of the `searx_bang` function in there is duplicated from the query parsing.

And I agree that passing `language_codes` to a dict would indeed work better.
For language names that would be easy, but now that I think about it, I don't even know how this is supposed to work with country names like Canada where two languages could match or with most countries that not named in `languages.py` even though they technically have a row.
Also, I don't know if it's worth trying to match language codes at the query level anymore. Engines are supposed to handle this because each one has their own quirks. So perhaps `_parse_query` should only convert the language (and country?) names, but leave anything matching the language code regex for the engines to handle.

PR #2593, opened 2021-02-22
Update autocomplete

dalf:
## What does this PR do?

Autocomplete and actual query parsing are nearly identical code, but these features are split into two different files.

This PR merges [`searx.autocomplete.searx_bang`](https://github.com/searx/searx/blob/9ace31c829daa545b7316db5ff9adf2c11ffa5cb/searx/autocomplete.py#L38) into [`searx.query.RawTextQuery`](https://github.com/searx/searx/blob/9ace31c829daa545b7316db5ff9adf2c11ffa5cb/searx/query.py#L28) : 

Change on `RawTextQuery`:
* A new field `autocomplete_list` with the possible auto-completion.
* A new method `get_autocomplete_full_query(text)` to apply the auto-completion 
* Auto-completion always works on the last part of the query (on the master branch it is only the first part).
* Auto-completion with the external bangs.
* Auto-completion with the time-out  ( The query `<4 time`  searches for `time`, and set the timeout to 4 seconds, when the user types the first character `<`, searx suggests `<3` and `<850` )
* Auto-completion return `searx.answers` results. (search for `sum 42 1337` for example)

Note done in the PR, but possible with this PR:
* Auto-completion can show only valid language knowing the selected engines
* Auto-completion can forbid mix of external bangs and bang(s).

* Auto-completion works only on the first part of the query, but now the bangs can be at the end of the query.
* External bangs are not documented, auto-completion allows to discover them.
* The PR removes some minor bugs (different parsing between autocomplete.py and query.py)

Activate  autocomplete then type:
* `sum 1 2 3 4 5 6`: checks you see the sum of 
* `the query <`: autocomplete shows `the query <3`
* `the query !!g`: autocomplete shows external bangs for `google`, `gi`, `gim`, `gimimages`, `googleimages`.
* `the query !g`
* `the query <` : autocomplete shows `the query <3` and `the query <850`.

Note: 
* `the query !g` will auto complete with 'the query !go` ....
* but `the query !go` will auto complete with `!go the query` with `!go` in front.
* other special special characters behaves in the same way: 
   * fully parsed parts ( like `!go` for example) are moved at the front of the query : there are not parts of the query sent to the engines.
   * other parts ( like `!g` ) are parts of the query sent to the engines.
* for reference, see https://github.com/dalf/searx/blob/40cadbaa87ceeeb3e331f9e6959cfc5ea84daa0e/searx/query.py#L280

## Related issues

* Based on #2592
* Related to https://github.com/searx/searx/issues/2240
* Related to https://github.com/searx/searx/wiki/Milestones#milestone-19---autocomplete

### DIFF
--- a/searx/answerers/__init__.py
+++ b/searx/answerers/__init__.py
@@ -32,7 +32,7 @@ def ask(query):
     results = []
     query_parts = list(filter(None, query.query.split()))
 
-    if query_parts[0] not in answerers_by_keywords:
+    if not query_parts or query_parts[0] not in answerers_by_keywords:
         return results
 
     for answerer in answerers_by_keywords[query_parts[0]]:
--- a/searx/query.py
+++ b/searx/query.py
@@ -1,162 +1,330 @@
-#!/usr/bin/env python
+# SPDX-License-Identifier: AGPL-3.0-or-later
 
-'''
-searx is free software: you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
-
-searx is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Affero General Public License for more details.
-
-You should have received a copy of the GNU Affero General Public License
-along with searx. If not, see < http://www.gnu.org/licenses/ >.
-
-(C) 2014 by Thomas Pointhuber, <thomas.pointhuber@gmx.at>
-'''
-
+from abc import abstractmethod, ABC
 import re
 
 from searx.languages import language_codes
 from searx.engines import categories, engines, engine_shortcuts
+from searx.external_bang import get_bang_definition_and_autocomplete
 from searx.search import EngineRef
 from searx.webutils import VALID_LANGUAGE_CODE
+
+
+class QueryPartParser(ABC):
+
+    __slots__ = "raw_text_query", "enable_autocomplete"
+
+    @staticmethod
+    @abstractmethod
+    def check(raw_value):
+        """Check if raw_value can be parsed"""
+
+    def __init__(self, raw_text_query, enable_autocomplete):
+        self.raw_text_query = raw_text_query
+        self.enable_autocomplete = enable_autocomplete
+
+    @abstractmethod
+    def __call__(self, raw_value):
+        """Try to parse raw_value: set the self.raw_text_query properties
+
+        return True if raw_value has been parsed
+
+        self.raw_text_query.autocomplete_list is also modified
+        if self.enable_autocomplete is True
+        """
+
+    def _add_autocomplete(self, value):
+        if value not in self.raw_text_query.autocomplete_list:
+            self.raw_text_query.autocomplete_list.append(value)
+
+
+class TimeoutParser(QueryPartParser):
+
+    @staticmethod
+    def check(raw_value):
+        return raw_value[0] == '<'
+
+    def __call__(self, raw_value):
+        value = raw_value[1:]
+        found = self._parse(value) if len(value) > 0 else False
+        if self.enable_autocomplete and not value:
+            self._autocomplete()
+        return found
+
+    def _parse(self, value):
+        if not value.isdigit():
+            return False
+        raw_timeout_limit = int(value)
+        if raw_timeout_limit < 100:
+            # below 100, the unit is the second ( <3 = 3 seconds timeout )
+            self.raw_text_query.timeout_limit = float(raw_timeout_limit)
+        else:
+            # 100 or above, the unit is the millisecond ( <850 = 850 milliseconds timeout )
+            self.raw_text_query.timeout_limit = raw_timeout_limit / 1000.0
+        return True
+
+    def _autocomplete(self):
+        for suggestion in ['<3', '<850']:
+            self._add_autocomplete(suggestion)
+
+
+class LanguageParser(QueryPartParser):
+
+    @staticmethod
+    def check(raw_value):
+        return raw_value[0] == ':'
+
+    def __call__(self, raw_value):
+        value = raw_value[1:].lower().replace('_', '-')
+        found = self._parse(value) if len(value) > 0 else False
+        if self.enable_autocomplete and not found:
+            self._autocomplete(value)
+        return found
+
+    def _parse(self, value):
+        found = False
+        # check if any language-code is equal with
+        # declared language-codes
+        for lc in language_codes:
+            lang_id, lang_name, country, english_name = map(str.lower, lc)
+
+            # if correct language-code is found
+            # set it as new search-language
+
+            if (value == lang_id
+                or value == lang_name
+                or value == english_name
+                or value.replace('-', ' ') == country)\
+               and value not in self.raw_text_query.languages:
+                found = True
+                lang_parts = lang_id.split('-')
+                if len(lang_parts) == 2:
+                    self.raw_text_query.languages.append(lang_parts[0] + '-' + lang_parts[1].upper())
+                else:
+                    self.raw_text_query.languages.append(lang_id)
+                # to ensure best match (first match is not necessarily the best one)
+                if value == lang_id:
+                    break
+
+        # user may set a valid, yet not selectable language
+        if VALID_LANGUAGE_CODE.match(value):
+            lang_parts = value.split('-')
+            if len(lang_parts) > 1:
+                value = lang_parts[0].lower() + '-' + lang_parts[1].upper()
+            if value not in self.raw_text_query.languages:
+                self.raw_text_query.languages.append(value)
+                found = True
+
+        return found
+
+    def _autocomplete(self, value):
+        if not value:
+            # show some example queries
+            for lang in [":en", ":en_us", ":english", ":united_kingdom"]:
+                self.raw_text_query.autocomplete_list.append(lang)
+            return
+
+        for lc in language_codes:
+            lang_id, lang_name, country, english_name = map(str.lower, lc)
+
+            # check if query starts with language-id
+            if lang_id.startswith(value):
+                if len(value) <= 2:
+                    self._add_autocomplete(':' + lang_id.split('-')[0])
+                else:
+                    self._add_autocomplete(':' + lang_id)
+
+            # check if query starts with language name
+            if lang_name.startswith(value) or english_name.startswith(value):
+                self._add_autocomplete(':' + lang_name)
+
+            # check if query starts with country
+            # here "new_zealand" is "new-zealand" (see __call__)
+            if country.startswith(value.replace('-', ' ')):
+                self._add_autocomplete(':' + country.replace(' ', '_'))
+
+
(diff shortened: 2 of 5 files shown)


tests/unit/test_query.py
```diff
@@ -14,6 +14,16 @@ def test_simple_query(self):
         self.assertEqual(len(query.languages), 0)
         self.assertFalse(query.specific)
 
+    def test_multiple_spaces_query(self):
```
kvch, line 30: :tada: :tada: :tada:

tests/unit/test_query.py
```diff
@@ -105,3 +115,23 @@ def test_timeout_invalid(self):
         self.assertEqual(query.getQuery(), query_text)
         self.assertEqual(query.timeout_limit, None)
         self.assertFalse(query.specific)
+
+    def test_external_bang(self):
```
kvch, line 192: :tada: :tada: :tada:

tests/unit/test_query.py
```diff
@@ -105,3 +115,23 @@ def test_timeout_invalid(self):
...
+        query_text = '!!ddg the query'
+        query = RawTextQuery(query_text, [])
+
+        self.assertEqual(query.getFullQuery(), query_text)
+        self.assertEqual(len(query.query_parts), 1)
+        self.assertFalse(query.specific)
+
+    def test_external_bang_autocomplete(self):
```
kvch, line 208: :tada: :tada: :tada:

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
 
             # this force the timeout
             if query_part[0] == '<':
```
kvch: Maybe create a helper function `_is_timeout_setting` so the ugly check can be hidden:

```python
if _is_timeout_setting(query_part):
    # set timeout whatever
```

Helper func:
```python
def _is_timeout_setting(query_part):
    return query_part[0] == '<'
```

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
-                if VALID_LANGUAGE_CODE.match(lang):
-                    lang_parts = lang.split('-')
-                    if len(lang_parts) > 1:
-                        lang = lang_parts[0].lower() + '-' + lang_parts[1].upper()
-                    if lang not in self.languages:
-                        self.languages.append(lang)
-                        searx_query_part = True
-
-            # external bang
+            if query_part[0] == ':':
```
kvch: Same as for timeout above.

kvch (commented):
I have skimmed through the code. I will look into it more deeply after your other PR is merged.

searx/answerers/__init__.py
```diff
@@ -32,7 +32,7 @@ def ask(query):
     results = []
     query_parts = list(filter(None, query.query.split()))
 
-    if query_parts[0] not in answerers_by_keywords:
+    if len(query_parts) == 0 or query_parts[0] not in answerers_by_keywords:
```
return42: I would use : 

    if not query_parts or ...

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+                self.timeout_limit = float(raw_timeout_limit)
             else:
-                self.user_query_parts.append(query_part)
+                # 100 or above, the unit is the millisecond ( <850 = 850 milliseconds timeout )
+                self.timeout_limit = raw_timeout_limit / 1000.0
+            return True
+
+    def _language_autocomplete(self, value):
+        if len(value) == 0:
+            # show some example queries
```
kvch: Please refactor it to a for loop:
```python
for lang in [":en", ":en_us", ":english", ":united_kingdom"]:
    self.autocomplete_list.append(lang)
```

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+            self.autocomplete_list.append(":en")
+            self.autocomplete_list.append(":en_us")
+            self.autocomplete_list.append(":english")
+            self.autocomplete_list.append(":united_kingdom")
+        else:
```
kvch: Instead of putting these instructions in `else`, you could just return at the end of adding example quearies and we could win back one level of indentation.
```python
if len(value) == 0:
    for lang in [":en", ":en_us", ":english", ":united_kingdom"]:
        self.autocomplete_list.append(lang)
     return

for lc in language_codes:
    lang_id, lang_name, country, english_name = map(str.lower, lc)
    # etc
```

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+        found = False
+        # check if any language-code is equal with
+        # declared language-codes
+        for lc in language_codes:
+            lang_id, lang_name, country, english_name = map(str.lower, lc)
+
+            # if correct language-code is found
+            # set it as new search-language
+
+            if (value == lang_id
```
kvch, line 94: Please extract this long condition list into a helper function with a descriptive name. E.g. `is_valid_language_code`.

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+                found = True
+
+        if found:
+            return True
+        elif enable_autocomplete:
+            self._language_autocomplete(value)
+            return False
+
+    def _external_bang_autocomplete(self):
+        self.autocomplete_list.append('!!g')
```
kvch: Please use `for` loop.

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+            for bang_autocomplete in bang_ac_list:
+                self.autocomplete_list.append('!!' + bang_autocomplete)
+        return found
+
+    def _bang_autocomplete(self, first_char, value):
+        if len(value) == 0:
+            # show some example queries
+            for suggestion in ['images', 'wikipedia', 'osm']:
+                if suggestion not in self.disabled_engines or suggestion in categories:
+                    self.autocomplete_list.append(first_char + suggestion)
```
kvch: Please return here and move the code out from `else` similarly to what I have suggested in https://github.com/searx/searx/pull/2593/files#r584158089

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+
+                # check if query starts with language name
+                if lang_name.startswith(value) or english_name.startswith(value):
+                    self.autocomplete_list.append(':' + lang_name)
+
+                # check if query starts with country
+                if country.startswith(value.replace('_', ' ')):
+                    self.autocomplete_list.append(':' + country.replace(' ', '_'))
+
+    def _language_parse(self, value, enable_autocomplete):
```
kvch: This function does too many things. It should not call `_language_autocomlete` because it has nothing to do with configuring autocomplete. Instead please refactor the caller site (from line 83) like this:

```python
if len(value) > 0:
    searx_query_part = self._language_parse(value)
    if not searx_query_part and enable_autocomplete:
        self._language_autocomplete(value)

```

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+
+    def _bang_autocomplete(self, first_char, value):
+        if len(value) == 0:
+            # show some example queries
+            for suggestion in ['images', 'wikipedia', 'osm']:
+                if suggestion not in self.disabled_engines or suggestion in categories:
+                    self.autocomplete_list.append(first_char + suggestion)
+        else:
+            # check if query starts with categorie name
+            for categorie in categories:
```
kvch: Tyop: category

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+        elif enable_autocomplete:
+            self._language_autocomplete(value)
+            return False
+
+    def _external_bang_autocomplete(self):
+        self.autocomplete_list.append('!!g')
+        self.autocomplete_list.append('!!ddg')
+        self.autocomplete_list.append('!!bing')
+
+    def _external_bang_parse(self, value, enable_autocomplete):
```
kvch, line 203: Please extract `_bang_autocomplete` in a similar fashion as in `_language_parse`.

searx/query.py
```diff
@@ -63,100 +70,225 @@ def _parse_query(self):
...
+            for engine in engines:
+                if engine.startswith(value):
+                    self.autocomplete_list.append(first_char + engine.replace(' ', '_'))
+
+            # check if query starts with engine shortcut
+            for engine_shortcut in engine_shortcuts:
+                if engine_shortcut.startswith(value):
+                    self.autocomplete_list.append(first_char + engine_shortcut)
+
+    def _bang_parse(self, first_char, value, enable_autocomplete):
```
kvch: Please extract `_bang_autocomplete` in a similar fashion as in `_language_parse`.

kvch (changes requested):
All in all good changes. I left minor refactoring requests to separate responsibilities more cleanly.

searx/query.py
dalf: There is a now a `QueryPartParser` class, and subclasses : `TimeoutParser`, `LanguageParser`, `ExternalBangParser`, `BangParser`.

Each `QueryPartParser`:
* has a class method `check(raw_value)` to check if the raw_value can be parsed.
* has a method `__call__(self, raw_value)` to actual parse the raw_value. The method may return False (for example `!this is not a bang`)

`RawTextQuery._parse_query` becomes very simple.

`RawTextQuery.PARSER_CLASSES` allows a plugin to add some other parsers (the information won't be sent to the engine, but that's a start).

searx/answerers/__init__.py
dalf: Done

searx/query.py
dalf: Done

searx/query.py
dalf: Done

searx/query.py
dalf: Done

searx/query.py
dalf: Done

searx/query.py
dalf: Fixed

searx/query.py
dalf: [`get_bang_definition_and_autocomplete`](https://github.com/searx/searx/blob/e2fb50089219207f7e11bcfc049b48f2f4b9a6c8/searx/external_bang.py#L47) returns the bang_definition and the autocompletion information (since the external bangs are stored in a trie, it is "free").

Now `ExternalBangParser._parse` returns `found, bang_ac_list` but doesn't add anything to `autocomplete_list`.

searx/query.py
dalf: Done

searx/query.py
dalf: It can be 
```python
    def match_language(lang_id, lang_name, english_name, country):
       return value == lang_id
                  or value == lang_name
                  or value == english_name
                  or value.replace('-', ' ') == country

    def _parse(self, value):
        found = False
        # check if any language-code is equal with
        # declared language-codes
        for lc in language_codes:
            lang_id, lang_name, country, english_name = map(str.lower, lc)

            # if correct language-code is found
            # set it as new search-language

            if match_language(lang_id, lang_name, english_name, country) \
               and value not in self.raw_text_query.languages:
                found = True
                ...
```

Related to #2240

Can it be the topic of another PR ?

kvch: Sure.

kvch (approved):

PR #2596, opened 2021-02-24
Add Freesound engine

GazoilKerozen:
## What does this PR do?

Add search engine from [freesound.org](https://freesound.org) with embedded player. 

It's an important source of sounds for sound workers (musicians, radio artists, designers...)
I'm dreaming about an universal search engine for sounds which are not only music.

In settings.yml set `disabled` to `False` and add your API key, documentation link : https://freesound.org/docs/api

## Author's checklist

- Check `!fnd` search prefix
- Check player in preview.
- Check pagination
- Enjoy !

## Related issues

none


searx/engines/freesound.py
```diff
@@ -0,0 +1,67 @@
...
+from json import loads
+from urllib.parse import urlencode
+from datetime import datetime
+
+disabled = True
+api_key = ""
+
+# about
+about = {
+    "website": "https://freesound.org",
```
dalf, line 15: can add the field `"wikidata_id": "Q835703",`

(see https://www.wikidata.org/wiki/Q835703 )

.gitignore
```diff
@@ -26,3 +26,4 @@ dist/
 local/
 gh-pages/
 searx.egg-info/
+searx/settings_perso.yml
```
dalf: Can you remove this change in .gitignore ?

searx/settings.yml
```diff
@@ -431,6 +431,14 @@ engines:
       website: https://directory.fsf.org/
       wikidata_id: Q2470288
 
+  - name : freesound
+    engine : freesound
+    shortcut : fnd
+    disabled : True
+    timeout : 15.0
+    api_key : MyAPIkey
+    # API key required, see : https://freesound.org/docs/api/overview.html
```
dalf, line 440: Since the engine requires an API key, can you comment this section ?

searx/engines/freesound.py
GazoilKerozen: done

.gitignore
GazoilKerozen: yup

searx/settings.yml
GazoilKerozen: ok, ready  to go :-) thanx !

searx/engines/freesound.py
```diff
@@ -0,0 +1,68 @@
+"""
```
return42, line 2: Can you add a .

    # SPDX-License-Identifier: AGPL-3.0-or-later

searx/settings.yml
dalf: Sorry my comment was not clear.

It should be:
```
    # - name : freesound
    # engine : freesound
    # shortcut : fnd
    # disabled : True
    # timeout : 15.0
    # api_key : MyAPIkey
    # API key required, see : https://freesound.org/docs/api/overview.html
```

Otherwise the engine is available but not usable (since there is no API key by default).

---

`disable: True` means disable by default, but it can be enabled using `/preferences` URL.

searx/engines/freesound.py
GazoilKerozen: sure

searx/settings.yml
GazoilKerozen: oh, ok ! thanks for the clarifications

dalf (approved):
Thank you !
Sorry for the delay.

PR #2600, opened 2021-02-25
Add searx_extra package

dalf:
## What does this PR do?

Split the utils directory into:
* `searx_extra` contains the scripts to update the data, standalone_searx.py
* utils contains the files to build and setup searx.

RFC: the underscore is discouraged in `searx_extra` but it doesn't seems to be a problem (the searx pypi package doesn't contains this directory, see setup.py).

There is a lot Python files in the `utils`  directory.
Each of them, includes something similar to `path.append(realpath(dirname(realpath(__file__)) + '/../'))`.
The test of `utils/standalone_searx.py` has to dynamically import the module.

It is easier to declare the directory has a Python module.

However the `utils` directory contains also the scripts to setup & install searx / filtron / morty.

* `make test`
* `python searx_extra/standalone_searx.py`
* check each script in `./searx_extra/update/*.py` runs without errors as before

Note `./searx_extra/update/update_translations.sh` has been moved but not tested / updated.


kvch (dismissed):
After a rebase, this is good to go.

kvch (approved):

PR #2608, opened 2021-02-28
[py2to3] use unittest from py3, remove unittest2 from py2

return42:
- unittest2 is a backport of the new features added to the unittest testing framework in Python 2.7
- unittest2 was only needed in py2 and can be dropped now


dalf (approved):
Thank you!

Issue #2495, opened 2021-01-23
Use virtual hosts instead of location blocks


winux1:
**Is your feature request related to a problem? Please describe.**
Location blocks in the Apache site configuration file allow less costumizability than virtual hosts and they don't 
support subdomains. So you can not install Searx to a subdomain using the installation scripts. See #2432 and #2480
**Describe the solution you'd like**
I would like to change the Apache and Nginx configuration files from location block to virtual hosts when using the installation scripts.

unixfox:
What are you referring to? The wiki?

Please also use our issue template. It's really important! I just added back the fields so please fill them.

winux1:
@unixfox please remove the doc label

unixfox:
Why? You didn't even answer to my second question. Is it related to the wiki of Searx?

unixfox:
@return42 Can you confirm that we can't install Searx in a subdomain using the official installation scripts?

winux1:
@unixfox Sorry I'm a bit tired.  It is related to the installations scripts not to the docs. There isn't any problem with the docs.

return42:
>  It is related to the installations scripts not to the docs.

It is more or less related to both, the installation scripts do what is documented ;-)

> Can you confirm that we can't install Searx in a subdomain using the official installation scripts?

No, since we do not set Virtualhost for some reasons (see below)

Haven't tested my [prosposal](https://github.com/searx/searx/discussions/2480#discussioncomment-292640) .. @winux1 reported that it does [not work](https://github.com/searx/searx/discussions/2480#discussioncomment-292765) .  The solution @winux1 has been [posted](https://github.com/searx/searx/discussions/2480#discussioncomment-296806) does not really differ from my proposal (if you know how apache configuration works and except that the Virtualhost in this solution listen on port 80 / we won't support unencrypted communication).

Why I think setting Virtualhost is wrong by concept:  The scripts are made to integrate smooth into the different distributions.  

Each distribution does have its own concept of how a service should be set up. If you compare the different distributions you will see that Apache and uwsgi is installed and confgured very different.  By example: [The Debian Layout](https://searx.github.io/searx/admin/installation-apache.html#the-debian-layout) differs totally from what apache recommends.

All the distributions support one default Virtualhost from their packages, to this default Virtualhost you can add your modules or sites (this is what the searx scripts are doing). On all distributions you are free to configure additional (sub-) domains, because it is Apache, but none package from the distributors will maintain it.

[...]

PR #2618, opened 2021-03-03
Fix security vulnerabilities in provided nginx configuration

thezeroalpha:
## What does this PR do?

This PR fixes misconfigurations in the suggested nginx configuration files, both in the templates and in the documentation, as reported by [Gixy](https://github.com/yandex/gixy).

The misconfigurations lead to security vulnerabilities and possible exploits such as [host spoofing](https://github.com/yandex/gixy/blob/master/docs/en/plugins/hostspoofing.md) and [path traversal](https://github.com/yandex/gixy/blob/master/docs/en/plugins/aliastraversal.md). If a user copies the configuration directly from the documentation, they will introduce these vulnerabilities into their web server.

### DIFF
--- a/docs/admin/filtron.rst
+++ b/docs/admin/filtron.rst
@@ -173,7 +173,7 @@ Use it along with ``nginx`` with the following example configuration.
    location /searx {
        proxy_pass         http://127.0.0.1:4004/;
 
-       proxy_set_header   Host             $http_host;
+       proxy_set_header   Host             $host;
        proxy_set_header   Connection       $http_connection;
        proxy_set_header   X-Real-IP        $remote_addr;
        proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;
--- a/docs/admin/installation-nginx.rst
+++ b/docs/admin/installation-nginx.rst
@@ -182,7 +182,7 @@ Started wiki`_ is always a good resource *to keep in the pocket*.
 	 location /searx {
 	     proxy_pass         http://127.0.0.1:4004/;
 
-	     proxy_set_header   Host             $http_host;
+	     proxy_set_header   Host             $host;
 	     proxy_set_header   Connection       $http_connection;
 	     proxy_set_header   X-Real-IP        $remote_addr;
 	     proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;
@@ -190,8 +190,8 @@ Started wiki`_ is always a good resource *to keep in the pocket*.
 	     proxy_set_header   X-Script-Name    /searx;
 	 }
 
-	 location /searx/static {
-	     alias /usr/local/searx/searx-src/searx/static;
+	 location /searx/static/ {
+	     alias /usr/local/searx/searx-src/searx/static/;
 	 }
 
 
@@ -205,7 +205,7 @@ Started wiki`_ is always a good resource *to keep in the pocket*.
 	 location /morty {
              proxy_pass         http://127.0.0.1:3000/;
 
-             proxy_set_header   Host             $http_host;
+             proxy_set_header   Host             $host;
              proxy_set_header   Connection       $http_connection;
              proxy_set_header   X-Real-IP        $remote_addr;
              proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;
@@ -309,8 +309,8 @@ Started wiki`_ is always a good resource *to keep in the pocket*.
              proxy_buffering off;
          }
 
-         location /searx/static {
-             alias /usr/local/searx/searx-src/searx/static;
+         location /searx/static/ {
+             alias /usr/local/searx/searx-src/searx/static/;
          }
 
       The ``X-Script-Name /searx`` is needed by the searx implementation to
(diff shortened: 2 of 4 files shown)


unixfox (dismissed):
Thank you for contribution.

return42:
Thanks for having an eye on security! .. Before I review (test) this PR I have general questions:

- how trustworthy is Gixy? .. last changes to the tool are from 2019  .. last release is v0.1.20 from Sep 28, 2018
- should we really change from `$http_host` to `$http` they are different in some circumstances and this will release new issues & question, I think .. on the other hand what is the security we win when using `$http`  .. correct me if I am wrong, but I guess the security improvement is near by zero? (.. or did I miss something?)

Adding add a subsequent `/` is something we should do (.. even if the security increment is also near by zero).

thezeroalpha:
All good questions. I can't speak to the trustworthiness of Gixy, though as it does not make any changes to the code but only notifies in the case of a detected misconfiguration, I don't think it matters much. Though it is true that its suggestions may no longer be up-to-date.

My rationale for using `$host` instead of `$http_host` is based off [a discussion in the nginx mailing list](https://forum.nginx.org/read.php?2,257134,257134#msg-257134) and [this discussion on StackExchange](https://serverfault.com/a/706439), as well as other discussions around the internet. Though there are arguments both for and against `$http_host`. Here's an [example](https://github.com/yandex/gixy/issues/67) of how spoofing can be done, and potentially mitigated by not relying on `$http_host`. However, I am not very familiar with Searx's codebase, so I'm not sure whether this change could raise any other issues; perhaps someone else will be able to weigh in.

It's true that for this specific setup, the security benefit is low.  My main concern is that it's better if the configuration provided by Searx is free of such vulnerabilities, because someone may take this configuration and change it to the point where this _does_ make a significant difference, but they might not be aware of it.

return42:
@thezeroalpha thanks for sharing your thoughts

> I can't speak to the trustworthiness of Gixy, though as it does not make any changes to the code but only notifies in the case of a detected misconfiguration, I don't think it matters much. 

Does not matter here in this changes but might answer the question if I should use Gixy in the future ;)

> My rationale for using $host instead of $http_host is based ... Here's an example of how spoofing can be done, and potentially mitigated by not relying on $http_host

After thinking twice I came to the conclusion we should give `$host` a try (i will make some test).  This includes `/templates/etc/nginx/default.apps-available/morty.conf`

My Problem is,  that this requires the correct `server_name` in the *server block* of the ngingx configuration .. all setups we support do not touch the server  *server block* .. see my comment [Why I think setting Virtualhost is wrong by concept: ...](https://github.com/searx/searx/issues/2495#issuecomment-766147787)

@dalf what do you think, do we need `$host` at all? .. I mean we already have the `base_url` in settings.yml .. or is this setting needed by the underlying http request implementations?

return42:
Strange, I can't approve this (a already approved) PR on github .. anyway:

I tested with the following setup (and a cup of coffee):

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index e45afb59b..54186270a 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -4,7 +4,7 @@ general:
     contact_url: False # mailto:contact@example.com
 
 brand:
-    git_url: https://github.com/searx/searx
+    git_url: https://github.com/thezeroalpha/searx
     git_branch: master
     issue_url: https://github.com/searx/searx/issues
     docs_url: https://searx.github.io/searx
```

```
make buildenv
sudo -H ./utils/lxc.sh remove
sudo -H ./utils/lxc.sh build
sudo -H ./utils/lxc.sh install suite
sudo -H ./utils/lxc.sh cmd -- FORCE_TIMEOUT=0 ./utils/filtron.sh nginx install
# Check nginx config ...
sudo -H ./utils/lxc.sh cmd -- cat /etc/nginx/default.apps-available/searx.conf | grep 'Host'
# List URLs to open in WEB browser ...
sudo -H ./utils/lxc.sh show suite | grep  "http://[^ ]*/searx"
```

side note: CentOS and Fedora are failing for some other reasons / not related to this PR.

Over all it seems that changing to `$host` does not have any negative side effect, so I would vote to merge this PR.

There is only one request I have: can you fix `$http_host` in the morty template also ..

https://github.com/searx/searx/blob/c748fc66cf7c4a4ebfecde61dd683422dd6b3901/utils/templates/etc/nginx/default.apps-available/morty.conf#L3-L6

Thanks for helping us to improve searx!

dalf (approved):
Thank you!

thezeroalpha:
@return42 Thank you for testing it!

> Does not matter here in this changes but might answer the question if I should use Gixy in the future ;)

I think it's a useful tool, of course its recommendations should be taken with a grain of salt but it might help detect misconfigurations like missing trailing slashes, etc. I use it on my own web servers, so far pretty successfully.

> My Problem is, that this requires the correct server_name in the server block of the ngingx configuration .. all setups we support do not touch the server server block .. see my comment Why I think setting Virtualhost is wrong by concept: ...

Yes, you're right. This is a good point. 

> There is only one request I have: can you fix $http_host in the morty template also ..

I must've missed it while looking for places where $http_host was used, thanks for catching it! Done.

PR #2638, opened 2021-03-06
plugin: generic_url_rewrite.py

### LINKED ISSUE #2609: Capability to rewrite hits url
**Is your feature request related to a problem? Please describe.**
Not related to a problem with `searx`, per se. I am looking to have `searx` rewrite the hits urls to another domain. 

**Describe the solution you'd like**

For example, if my results have a few urls from `www.reddit.com`, I'd like to have those rewitten to `old.reddit.com`.  If you could point me to where to start looking at, I'd be glad to send a PR if I am successful. My apologies if this already exists - if so, please guide how to accomplish this. 

In `settings.yml`, a section to include plain domain rewrites
<pre>
www.reddit.com: old.reddit.com
www1.domain.com: www3.domain.com
old.foo.org: new.foo.org
</pre>
**Describe alternatives you've considered**

Currently, I go to the address bar and perform manual rewrites. It would be great if `searx` can do the rewrites for me instead.  
**Additional context**

Sometimes, you'd want to access a specific endpoint for a results URI.

anmathew:
## What does this PR do?
Adds a new plugin <b>generic_url_rewrite</b> that allows to rewrite hits url. 

A new plugin, as we discussed in https://github.com/searx/searx/issues/2609 

Plugin to rewrite generic hits, may be useful to others 

Edit file `plugins/genericRewrites/config.yml` and add a test entry as 
<pre>
www.reddit.com: old.reddit.com
</pre>
And enable the plugin from `Preferences`.
Then perform a search for `reddit searx` - the output urls should be rewritten to `old.reddit.com` 

## Author's checklist
Tested locally on my lxc machine, 

Please feel free to tweak, if you'd like. 
## Related issues
Closes #2609

### DIFF
--- a/searx/plugins/generic_url_rewrite.py
+++ b/searx/plugins/generic_url_rewrite.py
@@ -1,0 +1,73 @@
+# See https://github.com/searx/searx/issues/2609
+# Author: <a.mathew@outlook.com>
+
+from searx import searx_dir
+from searx.plugins import logger
+from os.path import join as pathJoin
+from flask_babel import gettext
+
+
+name = gettext("Hits rewrite")
+description = gettext("Rewrite hits to custom endpoints")
+default_on = False
+preference_section = "general"
+
+configPath = pathJoin(searx_dir, "plugins/genericRewrites/config.yml")
+config = None
+parsed = "parsed_url"
+
+logger = logger.getChild("generic_url_rewrite")
+
+
+def get_config():
+    """get config"""
+    global config
+    if not config:
+        config = getConfigDict(configPath)
+    if config:
+        return config
+    return False
+
+
+def getConfigDict(configFile):
+    try:
+        ourDict = {}
+        with open(configFile) as fobj:
+            for entry in fobj:
+                splat = entry.split(":")
+                if len(splat) != 2:
+                    continue
+                wonky = [1 for i in splat if not i.strip()]
+                if wonky:
+                    continue
+                key = splat[0].strip()
+                val = splat[1].strip()
+                ourDict[key] = val
+        return ourDict
+    except:
+        return {}
+
+
+def rewriteUrl(result, newHostname):
+    splat = result["url"].split("/")
+    splat[2] = newHostname
+    newUrl = "/".join(splat)
+    result["url"] = newUrl
+
+    parsedUrl = result[parsed]
+    snu = "{url}:{port}".format(url=newUrl, port=parsedUrl.port)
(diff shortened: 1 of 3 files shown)


unixfox:
Works great. I tested with:
```
en.wikipedia.org: en.m.wikipedia.org
```

I have been thinking that the plugin could list the URLs that it will rewrite in the plugins section in the preferences? This plugin is great for a local Searx installation but for a public/shared instance the users don't have access to the config.yml and don't know which URLs are going to be rewritten.

3nprob:
As an alternativ for consideration, I made #2724  with the same idea but:

* renamed to `hostname_replace` (since neither this nor mine are actually doing generic URL rewrites, just domain/netloc replacement)
* supports regular expressions
* reads configuration from settings instead of from harcoded file in source

searx/plugins/generic_url_rewrite.py
```diff
@@ -0,0 +1,73 @@
...
+from os.path import join as pathJoin
+from flask_babel import gettext
+
+
+name = gettext("Hits rewrite")
+description = gettext("Rewrite hits to custom endpoints")
+default_on = False
+preference_section = "general"
+
+configPath = pathJoin(searx_dir, "plugins/genericRewrites/config.yml")
```
asciimoo, line 15: Perhaps it would be better to create a section in the main `settings.yml` instead of adding a new config file without documentation and the option to configure the path. What do you think?

searx/plugins/generic_url_rewrite.py
```diff
@@ -0,0 +1,73 @@
...
+                key = splat[0].strip()
+                val = splat[1].strip()
+                ourDict[key] = val
+        return ourDict
+    except:
+        return {}
+
+
+def rewriteUrl(result, newHostname):
+    splat = result["url"].split("/")
```
asciimoo, line 52: please use proper url parsing.

searx/plugins/generic_url_rewrite.py
```diff
@@ -0,0 +1,73 @@
...
+    splat[2] = newHostname
+    newUrl = "/".join(splat)
+    result["url"] = newUrl
+
+    parsedUrl = result[parsed]
+    snu = "{url}:{port}".format(url=newUrl, port=parsedUrl.port)
```
asciimoo, line 58: What if the rewrited url has a different port number than the original one? I'd use the port number from the new url.

asciimoo (changes requested):
Thanks, great feature, please check my notes.

PR #2640, opened 2021-03-07
[fix] rewrite Yahoo-News engine

return42:
Many things have been changed since last review of this engine.  This patch fix xpath selectors and is a complete review / rewrite of the engine.

Yahoo News engine does not work anymore

     !yhn linux

### DIFF
--- a/searx/engines/yahoo_news.py
+++ b/searx/engines/yahoo_news.py
@@ -1,16 +1,35 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Yahoo (News)
+"""Yahoo (News)
+
+Yahoo News is "English only" and do not offer localized nor language queries.
+
 """
 
+# pylint: disable=invalid-name, missing-function-docstring
+
 import re
-from datetime import datetime, timedelta
 from urllib.parse import urlencode
-from lxml import html
-from searx.engines.yahoo import parse_url, language_aliases
-from searx.engines.yahoo import _fetch_supported_languages, supported_languages_url  # NOQA # pylint: disable=unused-import
+from datetime import datetime, timedelta
 from dateutil import parser
-from searx.utils import extract_text, extract_url, match_language
+from lxml import html
+
+from searx import logger
+from searx.utils import (
+    eval_xpath_list,
+    eval_xpath_getindex,
+    extract_text,
+)
+
+from searx.engines.yahoo import parse_url
+
+# pylint: disable=unused-import
+from searx.engines.yahoo import (
+    _fetch_supported_languages,
+    supported_languages_url,
+)
+# pylint: enable=unused-import
+
+logger = logger.getChild('yahoo_news engine')
 
 # about
 about = {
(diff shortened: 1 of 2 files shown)


searx/utils.py
```diff
@@ -607,3 +607,30 @@ def eval_xpath_getindex(elements, xpath_spec, index, default=NOTSET):
...
     return default
+
+
+class RE:
+    u"""regular expression that stores last match (like Perl's ``=~`` operator)"""
+
+    # pylint: disable=missing-function-docstring
+    def __init__(self, *args, **kwargs):
+        self.re = re.compile(*args, **kwargs)
+        self.last_match = None
```
dalf: In the current architecture (multi thread), it creates side effect:
* the RE instance is shared in the yahoo engine between the different requests
* two requests can create:
     * write from thread 1
     * write from thread 2
     * read from thread 2
     * read from thread 1  <--- error

With async everywhere and no thread, it is technically okay, but still.

searx/utils.py
```diff
@@ -607,3 +607,30 @@ def eval_xpath_getindex(elements, xpath_spec, index, default=NOTSET):
...
+
+    def search(self, *args, **kwargs):
+        self.last_match = self.re.search(*args, **kwargs)
+        return self.last_match
+
+    def __getattr__(self, attr):
+        return getattr(self.re, attr)
+
+    def __getitem__(self, group):
+        if group < 0 or group > self.groups - 1:
```
dalf: `self.groups` is never assigned

searx/utils.py
return42: Its a class instance like any other .. should I move the `RE_ago` into the `response` function?

https://github.com/searx/searx/pull/2640/files#diff-94618cb260d81eafbc56387ca5bc868415b94871879888e4c90f2c5f2049bc63R55

searx/engines/yahoo_news.py
```diff
@@ -22,90 +40,77 @@
...
+
+        item = {
+            'url': url,
+            'title': title,
+            'content': content,
+            'img_src' : img_src
+        }
+
+        pub_date = extract_text(result.xpath('.//span[contains(@class,"s-time")]'))
+        if RE_ago.search(pub_date):
```
dalf: I don't know perl, but I'm not sure to understand the `RE` class.

The code here seems to be identical to:
```python
ago = regex_ago.search(pub_date)
if ago:
	ago = int(ago.group(0))
	if ago.group(1) == 'minute':
		pub_date = datetime.now() - timedelta(minutes=c)
       ....
```

Even simpler
```python
AGO_TIMEDELTA = {
  'minute': timedelta(minutes=1),
  'hour': timedelta(hour=1),
  'day': timedelta(day=1),
  'month': timedelta(month=1),
  'year': timedelta(year=1),
}

...
raw_pub_date= regex_ago.search(pub_date)
if ago:
	number = int(raw_pub_date.group(0))
	delta = AGO_TIMEDELTA[raw_pub_date.group(1)]
        pub_date = datetime.now() - delta * number
else:
        try:
            pub_date = parser.parse(raw_pub_date)
        except parser.ParserError:
            pub_date = None
```

dalf (approved):
Thank you!

PR #2642, opened 2021-03-08
[fix] APKMirror engine - update xpath selectors and fix img_src

return42:
## What does this PR do?

Fix APKMirror engine - update xpath selectors and fix img_src

BTW: make the code slightly more readable

engine APKMirror is broken

    !apkm calculator

### DIFF
--- a/searx/engines/apkmirror.py
+++ b/searx/engines/apkmirror.py
@@ -1,13 +1,21 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+"""APKMirror
 """
- APK Mirror
-"""
+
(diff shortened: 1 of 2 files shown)


searx/engines/apkmirror.py
```diff
@@ -20,47 +28,41 @@
 # engine dependent config
 categories = ['it']
```
return42: The F-Droid search engine is in category `files` .. should we move APKMirror also into category `files`?

dalf: I agree it should be in the `files` category.

dalf (approved):
Thank you!

PR #2643, opened 2021-03-08
[fix] make targets engines.languages and useragents.update

return42:
## What does this PR do?

Fix make targets `engines.languages` and `useragents.update`

Since PR #2600 is merged the update scripts for languages and useragent has been moved to folder:

    searx_extra/update/

    make engines.languages useragents.update


dalf:
Thank you!

PR #2652, opened 2021-03-13
Add Apache Solr engine

kvch:
## What does this PR do?

This PR adds a new engine that is able to query Apache Solr collections. Administrators must configure `collection` to use the engine. Also, generic settings are exposed as well:
* `lf` as `field_list`
* `df` as `default_fields`
* `qf` as `query_fields`

The query part users can submit is substituted into `q`. Thus, the generic `*:*` and so on queries work as expected.

Example output on the UI:

solrengine_000

The engine has been tested with Apache Solr v8.8.1.

This is another "offline" engine that relies on a backend that can be hosted on your machine.

Install Solr: https://solr.apache.org/guide/8_8/solr-tutorial.html
Uncomment solr section in `settings.yml`.
Search. Example query: `*:*`, `{fieldname}:{value}`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,6 +192,7 @@ PYLINT_FILES=\
 	searx/engines/google_images.py \
 	searx/engines/mediathekviewweb.py \
 	searx/engines/solidtorrents.py \
+	searx/engines/solr.py \
 	searx/engines/google_scholar.py \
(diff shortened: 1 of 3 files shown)


asciimoo (approved):
Great, thanks!

PR #2655, opened 2021-03-14
[fix] remove unused import from yahoo-news engine

return42:
## What does this PR do?

remove unused import from yahoo-news engine


dalf (approved):
Thank you!

PR #2656, opened 2021-03-14
[fix] url_for(..., _external=True) in templates

return42:
## What does this PR do?

Fix `webapp.url_for_theme` which is used in templates by `url_for(..., _external=True)`

The `url_for` function in the template context is not the one from Flask, it is the one from `webapp`.  The `webapp.url_for_theme` is different from its namesake of Flask and has it quirks, when called with argument `_external=True`.

https://github.com/searx/searx/blob/9292571304376c2c463fb26ee0db89298bb00e6c/searx/webapp.py#L395-L396

The `webapp.url_for_theme` can't handle absolute URLs since it pokes a leading '/', here is the snippet of the old code::

https://github.com/searx/searx/blob/9292571304376c2c463fb26ee0db89298bb00e6c/searx/webapp.py#L311-L315

Next drawback of (Flask's) `_external=True` is, that it will not return the HTTP scheme when searx (the Flask app) listens on http and is proxied by a https server.

To get the right scheme `HTTP_X_SCHEME` is needed by Flask (werkzeug).  Since this is not provided in every environment (e.g. behind Apache mod_wsgi or the HTTP header is not fully set for some other reasons) it is recommended to get *script_name*, *server* and *scheme* from the configured `base_url`.  If `base_url` is specified, then these values from are given preference over any Flask's generics.

  The issue is discussed here https://github.com/searx/searx/discussions/2639

BTW this patch normalize to use `url_for` in the `opensearch.xml` and drop the need of `host` and `urljoin` in template's context.

To test the issue discussed in https://github.com/searx/searx/discussions/2639 you will need a proxied searx instance with a https proxy (where `HTTP_X_SCHEME` is not set).  I tested it on my instance at https://darmarit.org/searx

----

Prerequisite-of: https://github.com/searx/searx/pull/2716

### DIFF
--- a/searx/templates/__common__/opensearch.xml
+++ b/searx/templates/__common__/opensearch.xml
@@ -3,7 +3,7 @@
   <ShortName>{{ instance_name }}</ShortName>
   <Description>a privacy-respecting, hackable metasearch engine</Description>
   <InputEncoding>UTF-8</InputEncoding>
-  <Image>{{ urljoin(host, url_for('static', filename='img/favicon.png')) }}</Image>
+  <Image>{{ url_for('static', filename='img/favicon.png', _external=True) }}</Image>
   <LongName>searx metasearch</LongName>
   {% if opensearch_method == 'get' %}
     <Url rel="results" type="text/html" method="get" template="{{ url_for('search', _external=True) }}?q={searchTerms}"/>
@@ -13,7 +13,7 @@
     </Url>
   {% endif %}
   {% if autocomplete %}
-    <Url rel="suggestions" type="application/x-suggestions+json" template="{{ host }}autocompleter?q={searchTerms}"/>
+    <Url rel="suggestions" type="application/x-suggestions+json" template="{{ url_for('autocompleter', _external=True) }}?q={searchTerms}"/>
   {% endif %}
 
   <Url type="application/opensearchdescription+xml"
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -40,7 +40,7 @@ from datetime import datetime, timedelta
 from time import time
 from html import escape
 from io import StringIO
-from urllib.parse import urlencode, urljoin, urlparse
+from urllib.parse import urlencode, urlparse
 
 from pygments import highlight
 from pygments.lexers import get_lexer_by_name
@@ -269,14 +269,7 @@ def extract_domain(url):
 
 
 def get_base_url():
-    if settings['server']['base_url']:
-        hostname = settings['server']['base_url']
-    else:
-        scheme = 'http'
-        if request.is_secure:
-            scheme = 'https'
-        hostname = url_for('index', _external=True, _scheme=scheme)
-    return hostname
+    return url_for('index', _external=True)
 
 
 def get_current_theme_name(override=None):
@@ -309,10 +302,6 @@ def url_for_theme(endpoint, override_theme=None, **values):
         if filename_with_theme in static_files:
             values['filename'] = filename_with_theme
     url = url_for(endpoint, **values)
-    if settings['server']['base_url']:
-        if url.startswith('/'):
-            url = url[1:]
-        url = urljoin(settings['server']['base_url'], url)
     return url
 
 
@@ -812,7 +801,7 @@ def preferences():
 
     # save preferences
     if request.method == 'POST':
-        resp = make_response(redirect(urljoin(settings['server']['base_url'], url_for('index'))))
+        resp = make_response(url_for('index', _external=True))
         try:
             request.preferences.parse_form(request.form)
         except ValidationException:
@@ -1002,11 +991,11 @@ def opensearch():
     if request.headers.get('User-Agent', '').lower().find('webkit') >= 0:
         method = 'get'
 
-    ret = render('opensearch.xml',
-                 opensearch_method=method,
-                 host=get_base_url(),
-                 urljoin=urljoin,
-                 override_theme='__common__')
+    ret = render(
+        'opensearch.xml',
+        opensearch_method=method,
+        override_theme='__common__'
+    )
 
     resp = Response(response=ret,
                     status=200,
@@ -1027,7 +1016,7 @@ def favicon():
 
 @app.route('/clear_cookies')
 def clear_cookies():
-    resp = make_response(redirect(urljoin(settings['server']['base_url'], url_for('index'))))
+    resp = make_response(redirect(url_for('index', _external=True)))
     for cookie_name in request.cookies:
         resp.delete_cookie(cookie_name)
     return resp
@@ -1128,19 +1117,38 @@ class ReverseProxyPathFix:
     '''
 
     def __init__(self, app):
+
         self.app = app
+        self.script_name = None
+        self.scheme = None
+        self.server = None
+
+        if settings['server']['base_url']:
+
+            # If base_url is specified, then these values from are given
+            # preference over any Flask's generics.
+
+            base_url = urlparse(settings['server']['base_url'])
+            self.script_name = base_url.path
+            self.scheme = base_url.scheme
+            self.server = base_url.netloc
 
     def __call__(self, environ, start_response):
-        script_name = environ.get('HTTP_X_SCRIPT_NAME', '')
+
+        script_name = self.script_name or environ.get('HTTP_X_SCRIPT_NAME', '')
         if script_name:
             environ['SCRIPT_NAME'] = script_name
             path_info = environ['PATH_INFO']
             if path_info.startswith(script_name):
                 environ['PATH_INFO'] = path_info[len(script_name):]
 
-        scheme = environ.get('HTTP_X_SCHEME', '')
+        scheme = self.scheme or environ.get('HTTP_X_SCHEME', '')
         if scheme:
             environ['wsgi.url_scheme'] = scheme
+
+        server = self.server or environ.get('HTTP_X_FORWARDED_HOST', '')
+        if server:
+            environ['HTTP_HOST'] = server
         return self.app(environ, start_response)
 
 


searx/templates/__common__/opensearch.xml
```diff
@@ -6,9 +6,9 @@
   <Image>{{ urljoin(host, url_for('static', filename='img/favicon.png')) }}</Image>
   <LongName>searx metasearch</LongName>
   {% if opensearch_method == 'get' %}
-    <Url rel="results" type="text/html" method="get" template="{{ url_for('search', _external=True) }}?q={searchTerms}"/>
+    <Url rel="results" type="text/html" method="get" template="{{ host }}search?q={searchTerms}"/>
```
kvch: Is this URL going to be correct always? Is it required for host to include `/`? I haven't seen a check for this in the code.

return42: Good question I had also .. I had this trailing slash issue in my engine, have a look at the existing *autocompleter* implementation:

https://github.com/searx/searx/blob/11070e57449caf86fd6087116e24bed8bea54e63/searx/templates/__common__/opensearch.xml#L16

So I added a trailing slash to the base_url like it was given in the comment `https://your.custom.host/location/`

https://github.com/searx/searx/blob/11070e57449caf86fd6087116e24bed8bea54e63/searx/settings.yml#L26

What do you think, does it make sense to extend ..

https://github.com/searx/searx/blob/11070e57449caf86fd6087116e24bed8bea54e63/searx/webapp.py#L272-L280

and check if there is a trailing slash?

return42:
Can we prioritize this PR?

dalf:
@return42 can you try the starlette branch in your current configuration / setup: https://github.com/dalf/searx/tree/starlette and check if the issues are fixed (http instead of https for example) ?

The branch is not up to date but:
* the implementation of `url_for` is different:
  * https://github.com/dalf/searx/blob/e8bf00132f093eab9dba12d1de1ba79bdb8a0961/searx/web/templates.py#L88
  * https://github.com/encode/starlette/blob/1222e78a15336162d190060f76c8ccee4b011d6b/starlette/templating.py#L56-L59
* there is no [reverse proxy path fix](https://github.com/searx/searx/blob/ae122ea943f77600fd97556503c483dcd92e1e63/searx/webapp.py#L1113-L1147) ( https://github.com/wilbertom/flask-reverse-proxy )

return42:
>  can you try the starlette branch in your current configuration

I tried, but  the branch is about 476 commits behind master and I did not have luck to get it running in my darmarit.org/searx setup.

The patch `url_for_theme` function ..

https://github.com/dalf/searx/blob/e8bf00132f093eab9dba12d1de1ba79bdb8a0961/searx/web/templates.py#L88-L94

also drops the `del values['_external']` so I think this patch addresses the same problem I try to solve.  I would vote for my solution, where the _usage_ of the `_external` argument is dropped from the templates .. I think it is a better compared to drop arguments from an argument list.  

IMO the usage of the  the _usage_ of the `_external` argument and this usage should be fixed.

> there is no reverse proxy path fix ( https://github.com/wilbertom/flask-reverse-proxy )

Can't say anything about proxy path fix (haven't had time to analyse) and the http / https problem, but I think this problem is not related to this PR ... [The `webapp.url_for_theme` can't handle absolute (external) URLs since it pokes a leading `/`](https://github.com/searx/searx/pull/2656#issue-592577306)

> Next drawback of (Flask's) `_external=True` is, that it will not return the https scheme when searx (the Flask app) listens on http and is proxied by a https server.

this might be related to the proxy-path-fix .. but I think we should not use the `_external=True` at all, at least not in a way we use it ATM, with poking a  leading `/` in the `url_for_theme`  .. external URLs always have a leading `h` from `http...` :-)

dalf:
Reading https://flask.palletsprojects.com/en/1.1.x/api/?highlight=url_for#flask.url_for
I discover that there are 3 config variables:
* `APPLICATION_ROOT`
* `SERVER_NAME`  (note: if not set, Flask uses the value from the `Host` header of the request.)
* `PREFERRED_URL_SCHEME`: (note: Flask does not seems to use the `X-Forwarded-Proto` header.)

So here a patch based on master :
```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index b4466c71..ead475bd 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -122,6 +122,13 @@ app.jinja_env.lstrip_blocks = True
 app.jinja_env.add_extension('jinja2.ext.loopcontrols')  # pylint: disable=no-member
 app.secret_key = settings['server']['secret_key']
 
+if settings['server']['base_url']:
+	base_url = urlparse(settings['server']['base_url'])
+	app.config['PREFERRED_URL_SCHEME'] = base_url.scheme
+	app.config['SERVER_NAME'] = base_url.netloc
+	app.config['APPLICATION_ROOT'] = base_url.path
+
+
 # see https://flask.palletsprojects.com/en/1.1.x/cli/
 # True if "FLASK_APP=searx/webapp.py FLASK_ENV=development flask run"
 flask_run_development = \
@@ -270,14 +277,7 @@ def extract_domain(url):
 
 
 def get_base_url():
-    if settings['server']['base_url']:
-        hostname = settings['server']['base_url']
-    else:
-        scheme = 'http'
-        if request.is_secure:
-            scheme = 'https'
-        hostname = url_for('index', _external=True, _scheme=scheme)
-    return hostname
+    return url_for('index', _external=True)
 
 
 def get_current_theme_name(override=None):
@@ -309,12 +309,7 @@ def url_for_theme(endpoint, override_theme=None, **values):
         filename_with_theme = "themes/{}/{}".format(theme_name, values['filename'])
         if filename_with_theme in static_files:
             values['filename'] = filename_with_theme
-    url = url_for(endpoint, **values)
-    if settings['server']['base_url']:
-        if url.startswith('/'):
-            url = url[1:]
-        url = urljoin(settings['server']['base_url'], url)
-    return url
+    return url_for(endpoint, **values)
 
 
 def proxify(url):
@@ -824,7 +819,7 @@ def preferences():
 
     # save preferences
     if request.method == 'POST':
-        resp = make_response(redirect(urljoin(settings['server']['base_url'], url_for('index'))))
+        resp = make_response(url_for('index', _external=True))
         try:
             request.preferences.parse_form(request.form)
         except ValidationException:
@@ -1038,7 +1033,7 @@ def favicon():
 
 @app.route('/clear_cookies')
 def clear_cookies():
-    resp = make_response(redirect(urljoin(settings['server']['base_url'], url_for('index'))))
+    resp = make_response(redirect(url_for('index', _external=True)))
     for cookie_name in request.cookies:
         resp.delete_cookie(cookie_name)
     return resp
```

Does it fix the issues?

If not, is it possible to keep one function to create the external URL?

```python
def url_for_external(relative_url):
    # TODO check there is no double / (one at the end of base_url, one at the beginning of relative_url)   
    return settings['server']['base_url'] + relative_url
```

So it is easier to adjust the URL creation in one place, and it avoids mistakes. For example should a `/` by added just after `{{base_url}}` ?

I'm sorry for the late review.

return42:
> I'm sorry for the late review.

Hey, no problem .. you released v1 which was a lot of work!

> If not, is it possible to keep one function to create the external URL?

I think this is the best solution.

> Does it fix the issues?

I will give it a try .. and come back.

searx/templates/__common__/opensearch.xml
return42: > Is this URL going to be correct always? Is it required for host to include /? I haven't seen a check for this in the code.

@kvch I normalized the opensearch.xml template to make use of the `url_for(..., _external=True)` function from Flask (aka werkzeug) [1]. This implementations handles a leading `/` well, it doesn't matter if three is a leading `/` or not, both configurations of `base_url` are now valid..

[1] https://github.com/searx/searx/pull/2656/files#diff-db0875e7d660ca2d6de1bbf3f8e7a069922bf959a398ff1b5592d24d8df76445L6-R16

return42:
> So here a patch based on master

@dalf thanks for the patch.  It was not the completion solution but points me in the right direction.  Instead of setting `app.conf` I patched the `ReverseProxyPathFix`, see commit message  (or my updated https://github.com/searx/searx/pull/2656#issue-592577306)

dalf:
Why is it `X-Forwarded-Server` and not [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) ?
* The [nginx documentation](https://searx.github.io/searx/admin/installation-nginx.html?highlight=forwarded) makes a reference to `X-Forwarded-Host`.
* There is no `X-Forwarded-Host` for the [Apache documentation](https://searx.github.io/searx/admin/installation-apache.html), but a reference to [`ProxyPreserveHost`](https://httpd.apache.org/docs/2.2/mod/mod_proxy.html#ProxyPreserveHost).

(I was not aware of `X-Forwarded-Server`, I found this https://stackoverflow.com/questions/43689625/x-forwarded-host-vs-x-forwarded-server I'm not sure if it is revelant or not )

----

Side note about `X-Scheme` (the value is already in the master branch):
* The [nginx documentation](https://searx.github.io/searx/admin/installation-nginx.html?highlight=forwarded) makes a reference to it.
* Not the [Apache documentation](https://searx.github.io/searx/admin/installation-apache.html) documentation.
* [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) seems more standard, but `X-Scheme` is fine too.

return42:
> Why is it X-Forwarded-Server and not X-Forwarded-Host ?

fixed it to `..-Host`

dalf (approved):
Thank you!

PR #2659, opened 2021-03-15
Fix HTTP error in onion engines

MarcAbonce:
## What does this PR do?
Regression from https://github.com/searx/searx/pull/2641.
The .onion engines only serve HTTP, so the `enable_http` setting must be enabled for them to work.

1. Configure the settings to use TOR. For that, set the proxy setting like: `http : socks5h://127.0.0.1:9050` and enable `using_tor_proxy : True`.
2. Try to run a query on any of the onion engines, for example: `!ahmia test`.

Expected result: No HTTP error.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -82,6 +82,8 @@ outgoing: # communication with search engines
 #        https:
 #            - http://proxy1:8080
 #            - http://proxy2:8080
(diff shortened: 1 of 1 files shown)


dalf (approved):
Thank you!

PR #2661, opened 2021-03-16
[mod] preferences: a tooltip is shown when the mouse is over the engine names

dalf:
## What does this PR do?

Based on the about section of the engines, a tooltip is shown when the mouse is over the engine names:

image

It avoids to search for the name 

check it is works with all the themes (oscar and 3 differents styles, simple).

## Related issues

Another PR can display the engine descriptions using the result of https://github.com/searx/searx/blob/master/searx_extra/update/update_engine_descriptions.py


searx/templates/simple/preferences.html
```diff
@@ -18,6 +18,19 @@
...
 {%- endmacro -%}
 
+{% macro engine_about(search_engine) -%}
+{% if search_engine.about is defined %}
+{% set about = search_engine.about %}
+<div class="engine-tooltip" role="tooltip">{{- "" -}}
+    <p><a href="{{about.website}}" rel="noreferrer">{{about.website}}</a></p>
+    {%- if about.wikidata_id -%}<p><a href="https://www.wikidata.org/wiki/{{about.wikidata_id}}" rel="noreferrer">wikidata.org/wiki/{{about.wikidata_id}}</a></p>{%- endif -%}
+    {%- if search_engine.enable_http %}<p>{{ icon('exclamation-sign', 'No HTTPS') }}{{ _('No HTTPS')}}</p>{% endif -%}
+    {%- if about.require_api_key %}<p>{{ _('Require API key')}}</p>{% endif -%}
```
kvch: Why is it important for instance users to know if the engine requires and API key? They cannot access the instance and configure a key for an engine. Also, why is it important for instance users to know if the engine uses the official API? If it does not use the API they would not use it?

dalf: It makes sense.

Now, the tooltip contains:
* the website
* the wikidata URL
* the warning about "No HTTPS"

PR #2669, opened 2021-03-19
Update settings.yml to enable HTTP for yacy engine

jhigginbotham:
Added a line to the yacy entry to enable HTTP if the local yacy instance isn't using HTTPS. Otherwise, an error will be thrown in the logs: "No connection adapters were found for 'http://localhost:8090/yacysearch.json...'". This is likely related to ticket searx#2641 that forces HTTPS by default.


dalf (approved):
Thank you!

PR #2678, opened 2021-03-23
New search engine: wordnik.com

### LINKED ISSUE #1844: Feature suggestion: Add Wordnik
Hi there, would it be possible to add Wordnik? This is where DuckDuckGo gets its definitions from, and they're much better than Wikipedia/Wiktionary. It would be nice to have this for instant answers on the sidebar, with top priority above other services that offer definitions, such as Wikipedia and Wolfram|Alpha. Thanks!

plague-doctor:
## What does this PR do?

Adds a new search engine Wordnik.com to retrieve word definitions. Similar to DuckDuckGo query `define: word`.

Current version of searx missing the functionality known from DuckDuckGo. This PR eliminates this gap.

Run `make run`.

## Related issues

Closes #1844


searx/engines/wordnik.py
```diff
@@ -0,0 +1,129 @@
...
+        return[]
+
+    results = []
+    word_defs = []
+    word_defs_ib = []
+    urls = []
+    lines = ""
+    lines_ib = "<div>"
+
+    word_xpath = '//*[@id="headword"]/text()'
```
kvch: These four variables can be moved to the global level of this file.

searx/engines/wordnik.py
```diff
@@ -0,0 +1,129 @@
...
+
+def response(resp):
+    if resp.status_code == 404:
+        return []
+
+    try:
+        dom = fromstring(resp.text)
+    except Exception:
+        raise_for_httperror(resp)
+        return[]
```
kvch: Tyop:
```python
return []
```

searx/engines/wordnik.py
```diff
@@ -0,0 +1,129 @@
...
+    for definition in definitions:
+        for i in range(len(definition)):
+            definition_extra_ib = ""
+            definition_extra = ""
+
+            part_of_speach_xpath = f"./li[{i+1}]/abbr"
+            definition_extra_xpath = f"./li[{i+1}]/i"
+            definition_text_xpath = f"./li[{i+1}]"
+
+            part_of_speach = extract_text(definition.xpath(part_of_speach_xpath))
```
kvch: Nit: isn't `speach` supposed to be `speech`?

searx/engines/wordnik.py
```diff
@@ -0,0 +1,129 @@
...
+SEARCH_URL = URL + '/words/{query}'
+
+
+def request(query, params):
+    params['url'] = SEARCH_URL.format(query=query)
+    return params
+
+
+def response(resp):
+    if resp.status_code == 404:
```
kvch: I think HTTP 404 is handled by the online plugin, there is no need for this check.

kvch:
Awesome PR!

searx/engines/wordnik.py
plague-doctor: Fixed.

searx/engines/wordnik.py
plague-doctor: OK.

searx/engines/wordnik.py
plague-doctor: Now you know why do I need wordnik :-)

searx/engines/wordnik.py
```diff
@@ -0,0 +1,126 @@
...
+
+    results = []
+    word_defs = []
+    word_defs_ib = []
+    urls = []
+    lines = ""
+    lines_ib = "<div>"
+
+    word = extract_text(dom.xpath(word_xpath))
+    definition_from = extract_text(dom.xpath(from_xpath))
```
return42: There are multiple definition_from ...

- from The American Heritage® Dictionary of the English Language, 5th Edition.
- from The Century Dictionary.
- ...

searx/engines/wordnik.py
```diff
@@ -0,0 +1,126 @@
...
+    lines_ib += f"<small>{definition_from}</small></div>"
+
+    urls.append({'title': word, 'url': f"{URL}/words/{word}"})
+
+    results.append({
+        'infobox': word,
+        'content': lines_ib,
+        'urls': urls})
+
+    results.append({
```
return42, line 72: IMO this is not needed / _infobox_ should be enough I think.

searx/engines/wordnik.py
```diff
@@ -0,0 +1,126 @@
...
+    "results": 'HTML',
+}
+
+categories = ['general']
+paging = False
+
+word_xpath = '//*[@id="headword"]/text()'
+from_xpath = '//*[@id="define"]/div/h3[1]'
+definitions_xpath = '//*[@id="define"]/div/ul[1]'
+misspelling_xpath = '//*[@id="define"]/div/ul/li/xref'
```
return42: Do you have an example (search term) matching this xpath? .. I tried several typos, but I did not find a term where I got a misspelling_xpath in the response from wrodnick.

searx/engines/wordnik.py
```diff
@@ -0,0 +1,126 @@
...
+
+def request(query, params):
+    params['url'] = SEARCH_URL.format(query=query)
+    return params
+
+
+def response(resp):
+    try:
+        dom = fromstring(resp.text)
+    except Exception:
```
return42: Never catch general exception.

searx/engines/wordnik.py
```diff
@@ -0,0 +1,126 @@
...
+        raise_for_httperror(resp)
```
return42: This is only for HTTP errors (see my branch)

searx/settings.yml
```diff
@@ -1264,6 +1264,14 @@ engines:
     categories: videos
     disabled : True
 
+  - name: wordnik
+    engine: wordnik
+    shortcut: def
+    base_url: https://www.wordnik.com/
+    categories: general
+    timeout: 5.0
+    disabled: False
```
return42: I vote for disabling by default, since wordnik is english only and not all of our users are looking for english terms.

return42 (changes requested):
Thanks for this engine, I really like it but I think there are some quirks we should avoid, see my comments.  

In my branch https://github.com/return42/searx/commits/add-wordnik I rebased and squash your commits.

On top of this branch you find a commit that simplifies the implementation and fixes the quirks.

If it is OK for you merge my commit in this PR.

searx/engines/wordnik.py
plague-doctor: Sometimes the definitions take huge amount of space. I just wanted to mimic DuckDuckGo behavior and return only a few definitions. Just to get general meaning of the word not the whole etymology or every possible idiom. There was a link to the wordnik for. But I am quite happy with your approach.

searx/engines/wordnik.py
plague-doctor: For example `distopia`. It is easier for us - ESLs :-)

searx/settings.yml
plague-doctor: Absolutely! Left over from testing. Sorry.

searx/engines/wordnik.py
plague-doctor: Agree. Your take is much more simple and neat :-)

searx/engines/wordnik.py
plague-doctor: Yep.

searx/engines/wordnik.py
plague-doctor: Lots to be learnt. Will get there eventually.

searx/engines/wordnik.py
return42: Ah, Ok .. thanks .. I think it is OK to display it "as is" ..

image

plague-doctor:
Opss... I have screwed this up... Will create a new PR. Sorry.

PR #2702, opened 2021-03-29
Fix dark "expand" button from infobox

MarcAbonce:
I just noticed that the button at the bottom of the infoboxes is always light gray even in the dark theme. This should fix the issue.
It also uses a "finger" shaped cursor and a slightly different shade of gray when hovering over the button in both the light and dark themes.

### How to test:
1. Pick the logicodev dark theme.
2. Search anything on Wikipedia, for example: `:en !wp beatles`.
3. Hover and click over the "expand" button at the bottom part of the infobox.
4. Repeat steps with light theme.

Expected behaviour:
* When using the dark theme, the infobox should be entirely dark.
* This change shouldn't affect the color in the light theme.
* In both themes, the expand button should actually feel like a button when you hover over it.


return42 (approved):
Works like a charm / thanks!!

dalf (approved):

PR #2707, opened 2021-03-31
[fix] default_doi_resolver in preferences

return42:
## What does this PR do?

Instead of a hard-coded `oadoi.org` default, use the default value from
`settings.yml`.

Fix an issue in the themes: The replacement 'current_doi_resolver' contains the
doi_resolver_url, not the name of the DOI resolver.  Compare return value of::

    searx.plugins.oa_doi_rewrite.get_doi_resolver(...)

Fix a typo in `get_doi_resolver(..)`:  suggested by @kvch:

  *L32 should set doi_resolver not doi_resolvers*

To get in use of the setting `default_doi_resolver`.

In `settings.yml` change the value of `default_doi_resolver`, after

    make run

Open preferences and 

- reset all preferences
- Check Tab "General" pre-selected "Open Access DOI resolver"  .. should be the default from settings.yml 
- Save / run a query in category 'science' and check DOI URLs

### DIFF
--- a/searx/plugins/oa_doi_rewrite.py
+++ b/searx/plugins/oa_doi_rewrite.py
@@ -29,7 +29,7 @@ def get_doi_resolver(args, preference_doi_resolver):
     doi_resolvers = settings['doi_resolvers']
     doi_resolver = args.get('doi_resolver', preference_doi_resolver)[0]
     if doi_resolver not in doi_resolvers:
-        doi_resolvers = settings['default_doi_resolver']
+        doi_resolver = settings['default_doi_resolver']
     doi_resolver_url = doi_resolvers[doi_resolver]
     return doi_resolver_url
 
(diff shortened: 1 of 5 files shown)


searx/plugins/oa_doi_rewrite.py
```diff
@@ -40,9 +42,10 @@ def on_result(request, search, result):
 
     doi = extract_doi(result['parsed_url'])
     if doi and len(doi) < 50:
-        for suffix in ('/', '.pdf', '/full', '/meta', '/abstract'):
+        for suffix in ('/', '.pdf', '.xml', '/full', '/meta', '/abstract'):
             if doi.endswith(suffix):
                 doi = doi[:-len(suffix)]
+        logger.debug("found DOI: '%s'", doi)
```
kvch: Is it a leftover? If not, please move the import to the same line as settings:

```python
from searx import logger, settings
```

Otherwise delete both line.

kvch:
To me there is something else that is off in the DOI resolver plugin. I think this is not correct: https://github.com/searx/searx/blob/master/searx/plugins/oa_doi_rewrite.py#L31-L32

L32 should set `doi_resolver` not `doi_resolvers`:

```python
if doi_resolver not in doi_resolvers:
    doi_resolver = settings['default_doi_resolver']
```

WDYT? If you agree that's also an issue, do you mind fixing it in this PR?

return42:
> WDYT? If you agree that's also an issue, do you mind fixing it in this PR?

Wow, haven't seen that typo .. thanks a lot for the hint .. I fixed it and rebased.

searx/plugins/oa_doi_rewrite.py
return42: I decided to drop the logging (not really a need for).

kvch (approved):
Thank you!

PR #2717, opened 2021-04-04
Allow overriding env vars SEARX_SETTINGS_PATH,UWSGI_SETTINGS_PATH

3nprob:
l## What does this PR do?

Allows changing location of configuration files:

* `SEARX_SETTINGS_PATH` (default `/etc/searx/settings.yml`)
* `UWSGI_SETTINGS_PATH` (default `/etc/searx/uwsgi.ini`)

It may be desirable to separate static assets, runtime configuration, and runtime data.
For example when orhestrating with Nomad, templated files are generally available under `/secrets` or `/local` unless one hacks around it with volume mounts. This is therefore a big convenience for some scenarios.


dockerfiles/docker-entrypoint.sh
```diff
@@ -24,8 +24,8 @@ if [ -z "${BIND_ADDRESS}" ]; then
     export BIND_ADDRESS="${DEFAULT_BIND_ADDRESS}"
 fi
 
-export UWSGI_SETTINGS_PATH=/etc/searx/uwsgi.ini
```
kvch, line 27: Why is it needed to set the default values here as well if they had been set in the `Dockerfile` at an earlier point of execution?

3nprob: It's not necessary at all, but:

* There may exist users who currently rely on the value being set here in previous versions - this prevents that from breaking for such users

* Laziness - removes ned to check for value being set or not 

But if you prefer I'm happy to remove it to just have the default set in one place

kvch: Please remove it. If anyone complains about it, I am willing to take the blame. :D

3nprob: @kvch removed!

dalf:
The help message needs some adjustements:
https://github.com/searx/searx/blob/547478089f1a48bb22a7651435db6cf941afc2f5/dockerfiles/docker-entrypoint.sh#L17

----
Note: `docker-entrypoint.sh`:
* will create `${SEARX_SETTINGS_PATH}` and `${UWSGI_SETTINGS_PATH}` if they don't exist.
* will create `${SEARX_SETTINGS_PATH}.new` and `${UWSGI_SETTINGS_PATH}.new` if the configuration file embedded in the docker image are newer. See https://searx.github.io/searx/admin/installation-docker.html#command-line

kvch (approved):

Issue #1710, opened 2019-10-07
Option to rewrite YT and Twitter links


dimqua:
Searx has a plugin system and it should be possible to rewrite YouTube and Twitter links to use Invidious and Nitter respectively.

PrasannaVenkadesh:
Yes. I have implemented a plugin to rewrite youtube and twitter links with custom domain names that you provide in `settings.yml` but I am wondering whether to submit it as a single plugin or one plugin for each service like youtube, twitter, etc., because some might want just youtube links to be rewritten (or) just twitter links to be rewritten.

Once this is decided, I will issue pull request for the same.

dimqua:
> one plugin for each service

I think this is better.

return42:
I closed https://github.com/asciimoo/searx/pull/1726 for now.  After https://github.com/asciimoo/searx/pull/1938 gets merged, this could be implemented as external plugin. I leave that issue open for any further discussion on that topic ...

seniorm0ment:
Worth noting Bibliogram, which is an Instagram frontend.
Also maybe rewrite Google Maps/Apple Maps into OpenStreetMaps 

Edit: also there's Teddit.net now for Reddit

Privacy Redirect extension is good for redirects, however it does not rewrite.

3nprob:
Noting for others coming here that #1938 was superseded by the now merged #2074 - so there should be no blockers for such a plugin now from what I can tell.

Looks related to #2609

PR #2724, opened 2021-04-05
Add hostname_replace plugin

### LINKED ISSUE #2609: Capability to rewrite hits url

3nprob:
## What does this PR do?

This is an alternative to #2638 , borrowing from the existing work done by @anmathew and @evuraan.

It supports replacing the domain/hostname (specifically, `result['parsed_url'].netloc`) based on regular expressions.

* Adds new plugin `hostname_replace`
* Configuration is done via "normal" settings
* Supports regular expressions

Supports generic domain-replacements. Useful for redirecting sites to alternative frontends.

To rewrite youtube links to use a specific invidious instance, add to `settings.yml`:
```
hostname_replace:
  '(.*\.)?youtube\.com$': 'invidiou.site'
  '(.*\.)?youtu\.be$': 'invidiou.site'
  '(.*\.)?youtube-noocookie\.com$': 'invidiou.site'
```
(please use on sites you manage or have permission from the owner, this is for illustration only)

## Related issues

Closes #2609, #1710

### DIFF
--- a/searx/plugins/hostname_replace.py
+++ b/searx/plugins/hostname_replace.py
@@ -1,0 +1,28 @@
+import re
+from urllib.parse import urlunparse
+from searx import settings
+from searx.plugins import logger
+from flask_babel import gettext
+
+name = gettext('Hostname replace')
+description = gettext('Rewrite result hostnames')
+default_on = False
+preference_section = 'general'
+
+plugin_id = 'hostname_replace'
+parsed = 'parsed_url'
+
+replacements = {re.compile(p): r for (p, r) in settings[plugin_id].items()} if plugin_id in settings else {}
+
+logger = logger.getChild(plugin_id)
+
+
+def on_result(request, search, result):
+    if parsed not in result:
+        return True
+    for (pattern, replacement) in replacements.items():
+        if pattern.search(result[parsed].netloc):
+            result[parsed] = result[parsed]._replace(netloc=pattern.sub(replacement, result[parsed].netloc))
+            result['url'] = urlunparse(result[parsed])
+
+    return True
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -32,6 +32,7 @@ from searx.plugins import (oa_doi_rewrite,
                            https_rewrite,
                            infinite_scroll,
                            self_info,
+                           hostname_replace,
                            search_on_category_select,
                            tracker_url_remover,
                            vim_hotkeys)
@@ -168,6 +169,7 @@ plugins.register(hash_plugin)
 plugins.register(https_rewrite)
 plugins.register(infinite_scroll)
 plugins.register(self_info)
+plugins.register(hostname_replace)
 plugins.register(search_on_category_select)
 plugins.register(tracker_url_remover)
 plugins.register(vim_hotkeys)
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -104,6 +104,18 @@ outgoing: # communication with search engines
 #   - "HTTPS rewrite"
 #   - ...
 
+# Example to rewrite hostnames in external links
+#
+# enabled_plugins:
+#   - 'Hostname replace'
+# hostname_replace:
+#   '(.*\.)?youtube\.com$':           'invidious.example.com'
+#   '(.*\.)?youtu\.be$':              'invidious.example.com'
+#   '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
+#   '(.*\.)?reddit\.com$':            'teddit.example.com'
+#   '(.*\.)?redd\.it$':               'teddit.example.com'
+#   '(www\.)?twitter\.com$':          'nitter.example.com'
+
 checker:
     # disable checker when in debug mode
     off_when_debug: True


3nprob:
Room for improvement:

* Enable user configuration in frontend
  * it is not clear to me how this should be done without overcoupling the plugin to the backend
* Allow more complex rewrites: matching other parts of the URL, exempt-lists, etc
  * the current implementation should be good enough for major use-cases without being too complex

searx/plugins/hostname_replace.py
```diff
@@ -0,0 +1,28 @@
...
+
+logger = logger.getChild(plugin_id)
+
+
+def on_result(request, search, result):
+    if parsed not in result:
+        return True
+    for (pattern, replacement) in replacements.items():
+        if pattern.search(result[parsed].netloc):
+            result[parsed] = result[parsed]._replace(netloc=pattern.sub(replacement, result[parsed].netloc))
```
return42, line 25: I did not know, that `._replace()` is an [documented method of named tuples](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace).  Should we use it here / I don't now.

3nprob: Do you see any reason not to?
FWIW it's already in use by `tracker_url_remover`, and in `results.py`

3nprob:
Bringing over from https://github.com/searx/searx/pull/2638#issuecomment-813355895:
> @3nprob Do you like implement generic URL rewrites in #2724 (and rename it to generic_url_rewrite)?

Do you have any proposal of how that would look? There are several design directions to take and I figured by keeping this one smaller in scope there's still room for another more complete solution - the intention with this one is to server the use-case of replacing domains, not a generic rewrite DSL.

But - if you have a somewhat clear idea of how that would look I'm all ears!

return42:
> There are several design directions to take and I figured by keeping this one smaller in scope there's still room for another more complete solution 

Before implementing several plugins, I guess it is better to have one for all.

> But - if you have a somewhat clear idea of how that would look I'm all ears!

Very simple: A regular expression substitution ([re.sub](https://docs.python.org/3/library/re.html#re.sub)) 

A list of expressions can be configured in `settings.yml` at engine level (_untested example_):

    - name : youtube
      shortcut : yt
      engine : youtube_noapi
      url_rewrite :
        '(http.*\.)?youtube\.com(.*)$': '\1invidiou.site\2'
        '(http.*\.)?youtu\.be(.*)$': '\1invidiou.site\2'
        '(http.*\.)?youtube-noocookie\.com(.*)$': '\1invidiou.site\2'

Bonus: A list of expressions can be configured in `settings.yml` at global level


    default_doi_resolver : 'oadoi.org'
    ...
    url_rewrite :
      '(http.*\.)?youtube\.com(.*)$': '\1invidiou.site\2'
      '(http.*\.)?youtu\.be(.*)$': '\1invidiou.site\2'
      '(http.*\.)?youtube-noocookie\.com(.*)$': '\1invidiou.site\2'

3nprob:
I started doing that actually, but the issue there is that the regexes quickly become pretty complex - and it's very easy to shoot yourself in the shoot and miss edge-cases.

Example for your pattern (amended for python regex syntax):

```
>>> re.sub('(http.*\.)?youtube\.com(.*)$', '\g<1>invidiou.site\g<2>', 'http://other.site/youtube.com/foo')
'http://other.site/invidiou.site/foo'
```

even if this one is addressed, covering for all edge-cases like these correctly yields complex expressions even for simple use-cases like domain replacement that I think are beyond what an average user should be expected to deal with

Performance is another one - once you do get a correct expression, you never know when you're hit by URLs that make it eat up all your RAM or hog up the CPU

Basically, covering the 80-90% use-cases and making it 10-20% as difficult to use 

This is from someone who has spent way too many hours of their life debugging URL-parsing regexes made by myself as well as other people ;)

3nprob:
@asciimoo thoughts of this one vs #2638? :)

asciimoo:
I just realized that this aims to solve the same problem as #2638 . At first glance it looks simpler and more flexible. What disadvantages you see compared to #2638 ?

3nprob:
> I just realized that this aims to solve the same problem as #2638 . At first glance it looks simpler and more flexible. What disadvantages you see compared to #2638 ?

Honestly none really - meant to address the issues in that PR and with a more descriptive name.
I've been running it myself and it works fine for my use-cases (rewriting youtube/twitter/reddit/etc to alternative frontends)

asciimoo:
Could you add some commented-out example to the settings.yml to demonstrate the usage?

3nprob:
> Could you add some commented-out example to the settings.yml to demonstrate the usage?
Will do!

3nprob:
@asciimoo updated

kvch (approved):

PR #2736, opened 2021-04-08
Add new engine: SJP - Słownik języka polskiego

plague-doctor:
## What does this PR do?

Adds a new search engine: SJP - Słownik języka polskiego (The Great Dictionary of Polish).

The engine aggregates three major dictionaries of Polish language: 
- Wielki słownik ortograficzny PWN (The Great Spelling Dictionary of the Polish Language)
- Słownik języka polskiego PWN (Dictionary of Polish)
- Słownik poprawnej polszczyzny (Dictionary of Correct Polish)

The engine presents responses in the `infobox`.

At the moment Searx is missing the engines allowing users to improve their vocabulary. I am trying to patch this gap with wordnik.com for English language (already in PR: 2735) and now with SJP for Polish.

Run `make run` and search eg. `!sjp słownik`.

### DIFF
--- a/searx/engines/sjp.py
+++ b/searx/engines/sjp.py
@@ -1,0 +1,82 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Słownik Języka Polskiego (general)
+
+"""
+
+from lxml.html import fromstring
+from searx import logger
+from searx.utils import extract_text
+from searx.raise_for_httperror import raise_for_httperror
+
+logger = logger.getChild('sjp engine')
+
+# about
+about = {
+    "website": 'https://sjp.pwn.pl',
+    "wikidata_id": 'Q55117369',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+categories = ['general']
+paging = False
+
+URL = 'https://sjp.pwn.pl'
+SEARCH_URL = URL + '/szukaj/{query}.html'
+
+
+def request(query, params):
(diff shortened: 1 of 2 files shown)


searx/engines/sjp.py
```diff
@@ -0,0 +1,82 @@
...
+    logger.debug(f"query_url --> {params['url']}")
+    return params
+
+
+def response(resp):
+    results = []
+
+    raise_for_httperror(resp)
+    dom = fromstring(resp.text)
+    word = extract_text(dom.xpath('//*[@id="content"]/div/div[1]/div/div[1]/div[1]/div[2]/div/div/div[2]/div/div'))
```
kvch: Do you mind simplifying the XPATH expressions in the PR?
The page you are scraping is indeed very messy, but it would be nice to use simpler expressions so we could minimize the chance of this engine breaking if the page gets redesigned.

For starters, I think this line could be simplified to
```python
word = extract_text(dom.xpath('//div[@class="query"]'))
```

I have started to simplify the other expressions, but it is a bit hard as I do not speak Polish (besides "polak, wegier, dwa bratanki // i do szabli, i do szklanki" :D :D :D) so I have no idea if I am selecting the right parts of the page.

kvch (changes requested):
Please try to simplify XPATH expressions.

searx/engines/sjp.py
plague-doctor: Good advice @kvch.
Well, I can only say "egészségedre!" in your language. Probably the most important "do szklanki" :-)

kvch (approved):

kvch:
Thank you!

PR #2746, opened 2021-04-09
Add Wikiversity

WikipediaUserbox:
## What does this PR do?
Adds Wikiversity to searx.

It's a Wikimedia project.


WikipediaUserbox:
done in #1914, will find another wikimedia project not in #1914.

PR #2768, opened 2021-04-16
Fix typo

Zackptg5:
## What does this PR do?

Fix typo in go path for morty install

Was going through the install process step by setup when I found this typo. Took me a few minutes to figure out the problem cause I'm not as smart as I think I am. The tar gzip contains the go folder so that shouldn't be in the tar destination path. This same line in filtron install is correct btw

Ran the setup manually on my server


return42 (approved):

asciimoo (approved):
Thanks!

PR #2777, opened 2021-04-23
Add SpellChecker for requests

MykolaSauliak:
## What does this PR do?

Add new visual block (in results template) with a hint for correct request sentence
results.html and webapp.py are changed

image

other search engines have such feature
good to have I guess

make run


unixfox:
Did you test the behavior with the Google engine because if I remember well we already fetch the spellchecker of Google and show it as a suggestion?

MykolaSauliak:
I'm using python library, didn't find any google spellchecker or it doesn't
work now

сб, 24 квіт. 2021, 09:32 користувач Émilien Devos ***@***.***>
пише:

asciimoo:
Ohi! Thanks for the contribution. It is a useful and highly requested feature, however the implementation has some flaws. First, it doesn't support properly multiple languages currently, second it introduces a new dependency, what we try to avoid if it isn't necessary. I'm thinking about implementing it as a stand-alone plugin. In this case the extra dependency wouldn't cause an issue for all of our admins, anyone can decide if they want to install it. What do you think? If you agree, you can find an example stand-alone plugin here: https://github.com/asciimoo/searx_external_plugin_example

MykolaSauliak:
I didn't know about plugins for searx
I can try to create a stand-alone plugin and then the community will decide
to include or not this to the main codebase
Do you have any docs about plugins, how often they are used ?
or if there are any ways to conduct a survey we can do that, it will give
us more information about this popularity of this feature

пн, 3 трав. 2021 о 13:18 Adam Tauber ***@***.***> пише:

PR #2807, opened 2021-05-07
Cherry-pick initialization fixes to make master stable

kvch:


dalf:
For reference:
https://github.com/searx/searx/blob/5e53e9412d118a3a9b17eabd47f8e55612e17a7a/searx/network/network.py#L257-L258

https://github.com/searx/searx/blob/5e53e9412d118a3a9b17eabd47f8e55612e17a7a/tests/unit/network/test_network.py#L20-L26

PR #2819, opened 2021-05-16
Add Wikimini engine

CrocodileCroco:
## What does this PR do?

It adds Wikimini as a engine in Searx

Because Wikimini has 19k articles (in french version) and every pages are written in a way so it's easy to read and to understand

## Author's checklist

The fulltext and title parameters should not be removed from the search url because sometimes it will redirect instead of showing results

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -218,6 +218,24 @@ engines:
     engine : wikipedia
     shortcut : wp
     base_url : 'https://{language}.wikipedia.org/'
+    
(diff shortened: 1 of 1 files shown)


unixfox:
Can you squash all your commits into one?

CrocodileCroco:
> 
> 
> Can you squash all your commits into one?

@unixfox 

It is done now, i squashed all the commits into one

kvch (approved):

kvch:
Thank you!

PR #2829, opened 2021-05-28
Add json_engine configuration for ManKier

jordemort:
## What does this PR do?

This adds a `json_engine` configuration for [ManKier](https://www.mankier.com/)

ManKier is a very nice website for viewing manpages.

Add it to your `settings.yml`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -790,6 +790,22 @@ engines:
     shortcut : nt
     disabled : True
 
+  - name : mankier
+    engine : json_engine
+    search_url : https://www.mankier.com/api/v2/mans/?q={query}
+    results_query : results
+    url_query : url
+    title_query : name
+    content_query : description
+    categories : it
+    shortcut : man
+    about:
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -790,6 +790,23 @@ engines:
     shortcut : nt
     disabled : True
 
+  - name : mankier
+    engine : json_engine
+    paging : True
```
kvch: You enabled `paging`, yet I cannot see `{pageno}` in the `search_url`. Either add it to the URL or remove `paging: True`.

kvch (changes requested):
Thanks for the contribution! I have a minor comment regarding paging.

searx/settings.yml
jordemort: Oops, missed that in whatever engine I copy/pasted from; I've removed paging from the engine definition.

kvch (approved):
Thank you!

PR #2832, opened 2021-05-29
Add engine for Docker Hub

jordemort:
## What does this PR do?

This adds an engine for [Docker Hub](https://hub.docker.com/) (Docker's central repository of containers).

This is a lightly modified version of `json_engine.py`. I couldn't just use `json_engine` for this because the results from the API don't actually contain URLs. I considered adding more advanced formatting stuff to `json_engine` but I couldn't think of a sensible way to implement the logic of "add `_/` in front if there's no `/` in the name, otherwise add `r/`" logic in a way that seemed reasonable to express in YAML, so I just made it my own engine.

I search Docker Hub a lot but I don't like using their website much. I'd rather search it with searx.

`!dkr ubuntu`
`!dkr searx`

### DIFF
--- a/searx/engines/dockerhub.py
+++ b/searx/engines/dockerhub.py
@@ -1,0 +1,131 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from collections.abc import Iterable
+from json import loads
+from urllib.parse import urlencode
+from searx.utils import to_string
+
+about = {
+    "website": "https://hub.docker.com",
+    "wikidata_id": "Q100769064",
+    "official_api_documentation": "https://docs.docker.com/docker-hub/api/latest/",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["it"]
+paging = True
+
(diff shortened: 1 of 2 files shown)


unixfox:

Someone already did a PR for adding Docker Hub support to Searx here: https://github.com/searx/searx/pull/2543
How is yours better than this one?

jordemort:
@unixfox The API used by this one returns both "official" and "community" images; when I was playing with the API used by the other one, I could only get it to return one or the other in a single call.

jordemort:
To be clear, I was unaware of the other PR when I made this one; I think I must have missed it since it was closed; if I had known it existed I probably would have just used the code over there instead of rolling my own.  I haven't tried that one but the screenshot shows both an official and a community image, it also has thumbnails, which this PR does not. The main advantage of this PR over the other one is that it is open.

No hard feelings if you choose to revive and merge the other one instead; if that one gets merged, I will happily use it.

jordemort:
Cleaning this up, since it's been open for a while. I've come around to thinking the implementation in https://github.com/searx/searx/pull/2543 is better in every way and would recommend taking that one instead.

PR #2861, opened 2021-06-23
[enh] Add whaleslide.com

### LINKED ISSUE #1955: (Request) Add whaleslide.com as a Search Engine
### Request structure:
`https://whaleslide.com/search/web/{query}?p={pageno}`
### Language support too.
Example in FR ( yes, I am french ^^ ):
`https://whaleslide.com/search/web/{query}?p={pageno}&locale=fr-FR`
### Example in practice:
[https://whaleslide.com/search/web/searx?p=1&locale=en-US](https://whaleslide.com/search/web/searx?p=1&locale=en-US)
### Support also for secure search parameters on the search page 'TOOLS option' (but I did not find how to include it in command line)

allendema:
Upstream example: https://whaleslide.com/search/web/runlevels

## What does this PR do?

This adds whaleslide.com as another optional engine.

Added whaleslide.com using the ``` json_engine ```

The API used here is required and can be found with Developer Optiones > Network Tab

More engines to choose from.

Try different things using Python and make Searx better.

Do a search with:
``` !whale epal dimensions ```

## Related issues

Closes #1955

### CODE REVIEW

File: searx/settings.yml [hunk 1]
```diff
@@ -1347,6 +1347,27 @@ engines:
       require_api_key: false
       results: html
 
+  - name : whaleslide
+    shortcut : whale
+    engine : json_engine
+    paging : True
+    # This API Token is needed to work
+    search_url : https://search.whaleslide.com/api/v1/search_v2/web/{pageno}?q={query}&api_token=f8OhUDEYKUICHDnIxEgI7Cb4uYyTBqT4nO8iueNbfTO3devS24yElGqM7nCm
```
kvch, line 1385: Is this your development token? How can others acquire their own tokens? In general I am not a huge fan of hardwiring tokens into searx codebase. I would rather let instance admins get their own tokens and set it for themselves.

allendema: **How can others acquire their own tokens?**

> The API used here is required and can be found with Developer Optiones > Network Tab

No not my Token, they seem to use it everytime.

Maybe different users will get a different one but this has worked for me.
Its the same today as the one from weeks ago.

PR #2955, opened 2021-08-30
[enh] Add azlyrics.com

allendema:
Upstream example query: https://search.azlyrics.com/search.php?q=The+verbal+acupunture&w=lyrics&p=1

## What does this PR do?

Add azlyrics.com.

Have an engine to search for lyrics.

Added azlyrics.com using the ``` xpath ``` engine.

More engines.

Make Searx better.

Do a search with:
``` !lyrics The verbal acupunture ```

## Author's checklist

```content_xpath``` shows a part of the lyrics.

Optimally the ```title_xpath``` would include the performer too.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -742,7 +742,7 @@ engines:
       results: HTML
 
   - name : azlyrics
-    shortcut : lyrics
+    shortcut : azl
     engine : xpath
     timeout : 4.0
     disabled : True
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -741,6 +741,23 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name : azlyrics
+    shortcut : lyrics
+    engine : xpath
+    timeout : 4.0
+    disabled : True
+    categories : general
+    paging : True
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: searx/settings.yml [hunk 1]
```diff
@@ -741,6 +741,23 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name : azlyrics
+    shortcut : lyrics
```
kvch: I am a bit on the fence with this shortcut because it is not future-proof. We haven't got any lyrics search engines, but if we get some in the future it would look a bit weird to have one engine with this shortcut.  Maybe `azl` would be a better shortcut.

kvch (commented):
I have tested it, works fine, I just have to sleep on the shortcut.

kvch (changes requested):
I don't have to sleep on it. Please change the shortcut to something less generic e.g. `azl`.

searx/settings.yml
allendema: You are right, I changed it now.

kvch (approved):

kvch:
Thank you!

PR #2980, opened 2021-09-23
Add imdb support

### LINKED ISSUE #1145: Possible to add IMDB as an engine ?
Hey,

I think it's a good engine to add to !Videos or even !General.

What do you think ?

PS : [for those who don't know IMDB](http://imdb.com/)

Guilvareux:
## What does this PR do?

1 file addition - engines/imdb.py
The imdb engine gives a limited but decent mix of results (names, tvshows, movies) which are more forgiving than 'general' (if the name is not exact). e.g. 'Guardians' in default 'general' only returns 'The Guardians (2017)' whereas imdb returns pages sorted by popularity: Guardians of the galaxy, vol 2, vol 3, Rise of the Guardians etc. 

The actual search page has lackluster results so this engine takes the results from the 'suggested' dropdown (via the json endpoint) on imdb.com

I usually use imdb to search for names in films, tv shows or vice versa. This engine gives some limited support but provides 'more popular' results.

Enable imdb in settings and run "!imdb <query>"

## Author's checklist

Open to any feedback and especially with how the listings are formatted

## Related issues
Closes #1145

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -165,3 +165,4 @@ generally made searx better:
 - @jhigginbotham
 - @xenrox
 - @OliveiraHermogenes
+- Paul Alcock @Guilvareux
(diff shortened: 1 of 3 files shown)


kvch (approved):
Thank you!

PR #2981, opened 2021-09-23
Verify that Tor proxy works every time searx starts

MarcAbonce:
## What does this PR do?

Every time searx starts, if the `using_tor_proxy` setting is true, verify that outgoing HTTP(s) requests are actually passing through Tor. This is done by sending a request to https://check.torproject.org/api/ip which is Tor Project's official API for this purpose.

Avoid any future regression for #2968 and ensure that searx instance maintainers are configuring their Tor proxies correctly.

Default scenario:
1. Run searx with default settings.
1. No verification should be made. Everything should work as usual.

Correctly configured Tor:
1. Add a proxy such as `all:// : socks5h://127.0.0.1:9050` and run Tor in that port.
1. Run searx.
1. Logs should show "outgoing requests are using Tor" and everything should work as usual.

Incorrectly configured Tor:
1. Keep the proxies section empty or proxy to Tor using `socks5` instead of `socks5h`.
1. Run searx.
1. Logs should show an explanatory error and searx should stop running.

## Related issues
Continuation of #2969

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -7,9 +7,11 @@ from itertools import cycle
 
 import httpx
 
+from searx import logger
 from .client import new_client, get_loop
 
 
+logger = logger.getChild('network')
 DEFAULT_NAME = '__DEFAULT__'
 NETWORKS = {}
 # requests compatibility when reading proxy settings from settings.yml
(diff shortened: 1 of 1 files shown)


kvch:
@MarcAbonce Could you please check the CI failures? I would love to merge this.

MarcAbonce:
I didn't know that you could have more than one proxy configured in the settings, so I opened #3015 (cherry-picked from https://github.com/searxng/searxng/pull/341) that actually verifies that Tor is used in all the proxies and not just the first one.

PR #2986, opened 2021-09-27
[mod] Fix Libgen + Uncomment Ebay and Urbandictionary

allendema:
## What does this PR do?

- Change Libgen provider and use https by default.
- Umcomment Urbandictionary but disable it by default, it is working.
- Uncomment Ebay as it is working correctly.
- 
(For ebay in the future: base_url should be changed from settings.yml just like peertube or invidious)

Working Libgen results.

Have a working Urbandictionary engine.

!eb test
!ud lgtm
!lg permanent record

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -434,11 +434,11 @@ engines:
       require_api_key: false
       results: HTML
 
-#  - name : ebay
-#    engine : ebay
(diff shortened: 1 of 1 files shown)


kvch (approved):

kvch:
Thanks!

PR #2987, opened 2021-09-27
[enh] Add T-mobile.de and Zapmeta autocompleters

allendema:
## What does this PR do?

Adds to more engines to autocomplete queries.

Sometime !go autocompleter is blocked, take a look at [this](https://github.com/searx/searx/discussions/2977).

We now have t-online which itself uses !go and zapmeta which uses !bing.

They should do less blocking.

Change the autocomplete engine from preferenes to any of those.
## Author's checklist
N/A

## Related issues
https://github.com/searx/searx/discussions/2977

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -125,8 +125,8 @@ def zapmeta(query, lang):
     return results
 
 
-def tmobile(query, lang):
-    # suche.t-mobile.de autocompleter
+def tonline(query, lang):
+    # suche.t-online.de autocompleter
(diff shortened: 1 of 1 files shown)


unixfox:
Not useful because Google autocomplete is never blocked nor Bing is.

Both can't restrict/block their autocomplete because the request done by the autocomplete engine of the web browser doesn't pass any cookies and no other personal identifiers that can allow them to know with certainty if the request was done by a bot or not.

PR #2991, opened 2021-10-01
[enh] Fix uppercase ip query

### LINKED ISSUE #2888: Only a lower case "ip" displays the IP address
When the feature is enabled to show a user's IP address when "ip" is entered into the search bar, it only does so when it is all lowercase. Querying "IP" does not return an IP. This seems like a bug, apologies if this was intended.

Thanks

nav1s:
## What does this PR do?

Fixes the self_info plugin to support uppercase ip queries.

This PR solves the mild annoyance of retyping IP in lowercase.

http://127.0.0.1:8888/search?q=IP

## Related issues

Closes #2888


kvch (approved):
Thank you!

PR #3000, opened 2021-10-05
[enh] Add Pagination to Wiby / Fix Kaufland

allendema:
## What does this PR do?

Adds Pagination support to Wiby engine.

Fixes XPath expression for Kaufland and removes tracking parameter from returned URLs.

Make Searx better.

Do a example search with the engine:

!wiby blog

!kaufland mag

## Author's checklist
N/A

Added a regex to tracker remover plugin.

## Related issues
N/A


searx/settings.yml
```diff
@@ -1513,8 +1514,8 @@ engines:
     timeout : 3.0
     paging : True
     search_url : https://www.kaufland.de/item/search/?search_value={query}&page={pageno}
-    title_xpath : //li[@class="product__title"]
-    url_xpath : //div[@class="product"]//a/@href
+    title_xpath : //div[@class="product__title"]/text
```
kvch:
```suggestion
    title_xpath : //div[@class="product__title"]/text()
```

kvch (changes requested):
In this form, Kaufland does not return any results when I tested. After I fixed the XPATH for `title_xpath`, it was ok again. I have added a suggestion to the PR.

allendema:
Updated it, thanks for the review.
Can you check the regex if it is corret too?

kvch (approved):
Perfect! The regex works as well. Thanks.

PR #3015, opened 2021-10-13
Verify that Tor proxy works every time searx starts

MarcAbonce:
This is a cherry-pick of https://github.com/searxng/searxng/pull/341 which is meant to replace my other PR #2981.
The problem is that I didn't know that you could have more than one proxy configured in the settings, so this version actually verifies that Tor is used in all the proxies and not just the first one.

## What does this PR do?

Every time searx starts, if the `using_tor_proxy` setting is true, verify that outgoing HTTP(s) requests are actually passing through Tor. This is done by sending a request to https://check.torproject.org/api/ip which is Tor Project's official API for this purpose.

Avoid any future regression for #2968 and ensure that searx instance maintainers are configuring their Tor proxies correctly.

Default scenario:
1. Run searx with default settings.
1. No verification should be made. Everything should work as usual.

Correctly configured Tor:
1. Add a proxy such as `all:// : socks5h://127.0.0.1:9050` and run Tor in that port.
1. Run searx.
1. Tor check should run successfully. Everything should work as usual.

Incorrectly configured Tor:
1. Keep the proxies section empty or proxy to Tor using `socks5` instead of `socks5h`.
1. Run searx.
1. Logs should show an explanatory error and searx should stop running.

## Related issues
Rewrite of #2981
Continuation of #2969

### DIFF
--- a/searx/network/__init__.py
+++ b/searx/network/__init__.py
@@ -8,7 +8,7 @@ from time import time
 import httpx
 import h2.exceptions
 
-from .network import get_network, initialize
+from .network import get_network, initialize, check_network_configuration
 from .client import get_loop
 from .raise_for_httperror import raise_for_httperror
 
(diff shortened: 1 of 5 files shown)


kvch (approved):

kvch:
Thank you!

PR #3035, opened 2021-10-24
run tests under python 3.10

rooterkyberian:
## What does this PR do?

run tests under python 3.10

Keeping up to date with current python stable 3.10

Tested under CI


.github/workflows/integration.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
rooterkyberian: don't do `3.10` in yaml as it will be interpreted as number `3.1`, while you want `'3.10'` string

kvch: :+1:

kvch: Could you please update your branch?

PR #3053, opened 2021-11-06
[enh] Improve ranking based on language

### LINKED ISSUE #206: using language of result to process better ranking
If someone has specified a search-language, we could improve ranking using the language of the result and rank it higher, if it is matching.

**for example: wikipedia**

the subdomain in the wikipedia-url is clearly showing what language the result has. So we can use it to rank wikipedia-results in the specified search-language higher as the "normal" english results.

0xhtml:
## What does this PR do?

Add configurable setting to rank search results higher when part of the domain (e.g. `en` in `en.wikipedia.org` or `de` in `beispiel.de`) matches the selected search language. Does not apply to e.g. `be` in `youtube.com`.

The weight of a result matching the language is multiplied by `1.1`.

Some engines return results not necessarily matching the selected language. When searching for `alpha beta suche` with language set to `de`, there is a result for `de.wikipedia.org` as well as for the same article on `en.wikipedia.org`. With this PR the result for `de.wikipedia.org` would get a boost in ranking.

To test optionally apply the following patch for debugging, enable the setting `prefer_configured_language` in the `settings.yml` and search with language set to `de` for e.g. `!wikidata alpha beta suche`.

```patch
--- a/searx/results.py
+++ b/searx/results.py
@@ -136,11 +136,15 @@ def result_score(result, language_parts):
         if hasattr(engines[result_engine], 'weight'):
             weight *= float(engines[result_engine].weight)

+    print(weight, '->', end='')
+
     if settings['search']['prefer_configured_language']:
         domain_parts = result['parsed_url'].netloc.split('.')
         if any(part in domain_parts for part in language_parts):
             weight *= 1.1

+    print(weight)
+
     occurences = len(result['positions'])

     return sum((occurences * weight) / position for position in result['positions'])
```

Results should be `https://de.wikipedia.org/wiki/Alpha-Beta-Suche` and `https://en.wikipedia.org/wiki/Alpha%E2%80%93beta_pruning`. The results for `de.wikipedia.org` should be ranked higher and it's weight should be changed `2.0 -> 2.2` compared to the other results weight `2.0 -> 2.0`.

## Author's checklist

~~To be discussed: Should this feature be configurable through `settings.yml` or preferences.~~

Todo for a future PR: Consider other indicators for the results language and give engines that support the selected language a ranking boost.

## Related issues

Closes #206

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -6,6 +6,7 @@ from urllib.parse import urlparse, unquote
 from searx import logger
 from searx.engines import engines
 from searx.metrology.error_recorder import record_error
+from searx import settings
 
 
 CONTENT_LEN_IGNORED_CHARS_REGEX = re.compile(r'[,;:!?\./\\\\ ()-_]', re.M | re.U)
@@ -129,12 +130,17 @@ def merge_two_infoboxes(infobox1, infobox2):
             infobox1['content'] = content2
 
 
-def result_score(result):
+def result_score(result, language):
     weight = 1.0
 
     for result_engine in result['engines']:
         if hasattr(engines[result_engine], 'weight'):
             weight *= float(engines[result_engine].weight)
+
+    if settings['search']['prefer_configured_language']:
+        domain_parts = result['parsed_url'].netloc.split('.')
+        if language in domain_parts:
+            weight *= 1.1
 
     occurences = len(result['positions'])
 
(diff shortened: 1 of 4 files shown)


kvch:
Personally, I like this change. However, I would make it configurable by instance admins. Do you mind adding it to `settings.yml`? Maybe call it `prefer_configured_language` under `search` section?

0xhtml:
I updated the PR with the new setting `prefer_configured_language`.

kvch (approved):

kvch:
Awesome!

PR #3065, opened 2021-11-17
Onesearch engine

israelyago:
## What does this PR do?

Add a new search engine https://www.onesearch.com/

Add another search engine to have a more resilient pool of engines in case of failures

1. Run `sudo make run`
1. Go to `/preferences`
1. Disable all search engines with the exception of onesearch (the last one)
1. Make some search
1. Assert that it works as expected
1. Assert that the pagination works

## Author's checklist

This search engine is ON by default, since they claim to be privacy friendly

## Related issues

[/e/ foundation spot - issue 3676](https://gitlab.e.foundation/e/backlog/-/issues/3676)


searx/settings.yml
```diff
@@ -1622,6 +1622,18 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name: onesearch
+    shortcut: onesearch
+    engine: onesearch
+    paging: true
```
kvch: This should be set only in the engine.
```suggestion
```

searx/settings.yml
```diff
@@ -1622,6 +1622,18 @@ engines:
...
+    categories: general
```
kvch: Same as above.
```suggestion
```

kvch (commented):
Minor comments, awesome addition!

kvch (approved):

Issue #2958, opened 2021-09-01
URL preferences persistance


marcin-szmagara:
I use the URL preferences feature and have it set up as my browsers default search URL. When I resubmit the query, this time not through browser search box, but rather the one in searx website, default URL is used rather than one with the preferences. In other words, `?preferences=<long_string>` gets removed from the URL. This is especially painful because it switches to light theme.

I would like my preferences to be propagated to each subsequent query.

I do not use cookies.

crocket:
Search engines aren't preserved, either without preferences URL parameter.

kvch:
I am afraid we will not implement this feature or fix this bug. To persist the preferences this way searx has to keep a session for your searches. It has to know that you first submitted your search term with a given preferences string and persist it between later searches. It is contrary to the principles of searx. We are linking two searches together and we do not want to change this.

The alternative option that came to my mind is saving the preferences as cookies. However, it can lead to unexpected issues, like if you share a search URL with preferences string with someone who already uses cookies on that searx instance, their settings will be overwritten.

If someone else has a different approach I am willing to consider it. But now I do not see any other way to support this use case. You have to choose between persisting your preferences in cookies or using preferences strings in the URL that do not apply to later searches from the UI. Unfortunately, I do not see a third way at the moment.

crocket:
> The alternative option that came to my mind is saving the preferences as cookies.

Searx already does this.

kylian-002:
> > The alternative option that came to my mind is saving the preferences as cookies.
> 
> Searx already does this.

this doesn't work if your browser is setup to always be in incognito. saving your preferences then is extremely hard to do.

crocket:
I use my browser in incognito mode.

kvch:
> > The alternative option that came to my mind is saving the preferences as cookies.
> 
> Searx already does this.

I meant that once someone submits a query with a preferences string, searx could preserve those settings in cookies. Searx does not store settings from preferences strings in cookies. One can only save their settings to cookies if they want to by clicking on a button.

crocket:
That's a good workaround.

crocket:
Is searx going to add an option to preserve preferences URL parameter as cookies?
The option can be enabled in preferences URL parameter.

PR #3068, opened 2021-11-18
Preserve cookies after query

israelyago:
## What does this PR do?

Implements the "preserve preferences URL parameter as cookies" https://github.com/searx/searx/issues/2958 

After making a search with the url preferences, the user expects it's settings to persist across more searches. The current behavior does not persist the user configuration.

This branch allows the user to have it's, e.g. dark mode ON even after the first query

1. Set your theme to dark mode
1. Go to the /preferences page
1. Copy the preferences URL
1. Open an anonymous browser tab
1. Paste the preferences URL and update the last q=%s to q=dog

At the search results page, confirm that you are in dark mode
Go to the images page, confirm that you still have the dark mode

## Related issues

https://github.com/searx/searx/issues/2958
https://gitlab.e.foundation/e/backlog/-/issues/1154

### DIFF
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -367,6 +367,16 @@
 	  <small class="text-muted">({{ _('Note: specifying custom settings in the search URL can reduce privacy by leaking data to the clicked result sites.') }})</small>:
         </p>
 
+        <div class="row">
+            <div class="col-sm-1" style="margin-right: 2rem;">
+                {{- checkbox_toggle("save_preferences", cookies["save_preferences"] != "on") -}}
+            </div>
+            <div class="col-sm-10">
+                <p> Save my preferences with cookies </p>
+            </div>
+
+        </div>
+
 	<div class="tab-pane">
           <input readonly="" class="form-control select-all-on-click cursor-text" type="url" value="{{ url_for('index', _external=True) }}?preferences={{ preferences_url_params|e }}{% raw %}&amp;q=%s{% endraw %}">
           <input type="submit" class="btn btn-primary" value="{{ _('save') }}" />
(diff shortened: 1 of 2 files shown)


kvch (changes requested):
I have slept on the original issue. My main concern, cookies being accidentally overwritten, is still there. However, as a compromise we could merge this functionality. But we should add a new checkbox to the preferences URL, maybe called "save to cookies", and pass it through with the preferences string as is. Example `http://localhost:8888/?preferences={string}&save=true` and this would instruct searx to save the preferences to cookie. If `save` is set to false, it would behave as before, not save the preferences. By default, it should be set to false, so we are not breaking the expected behaviour. WDYT?

israelyago:
> I have slept on the original issue. My main concern, cookies being accidentally overwritten, is still there. However, as a compromise we could merge this functionality. But we should add a new checkbox to the preferences URL, maybe called "save to cookies", and pass it through with the preferences string as is. Example `http://localhost:8888/?preferences={string}&save=true` and this would instruct searx to save the preferences to cookie. If `save` is set to false, it would behave as before, not save the preferences. By default, it should be set to false, so we are not breaking the expected behaviour. WDYT?

That seems good to me; I will come back with the new updates

israelyago:
@kvch I thought the requested changes would be pretty simple to make, but I was surprised with the way the system works.
Our toggles send the inverted value to the server and somehow it works.

This search engine, which is DISABLED and have a UI id of "engine_archive_is__general" have a internal value of checked as "true":

image

image

And when submitting the form:

image

Is this the expected and right behavior of the system?

kvch:
Yes, unfortunately, this is the expected behaviour. It was implemented a long time ago, probably one of the first features of searx. We never changed it because we were afraid to break existing user preferences preserved in cookies.

kvch:
I have tested the PR  and the functionality works. But the display if the preferences are supposed to be saved stays grey on reload.

israelyago:
> ... But the display if the preferences are supposed to be saved stays grey on reload.

Yes, that's how I found out the little problem we have in the preferences thing; I'm still unsure how to fix the "Not updating the toggle" on reload

PR #3099, opened 2021-12-15
Disable onesearch by default

unixfox:
onesearch is not available everywhere and thus display an error by default in searx

Related to https://github.com/searx/searx/pull/3065

image
image


kvch (approved):
Thanks!

PR #3115, opened 2022-01-06
New files engine: Prowlarr

narcoticfresh:
## What does this PR do?

Gives the user the possibility to search their own prowlarr instances.

Info: https://wiki.servarr.com/en/prowlarr
Github: https://github.com/Prowlarr/Prowlarr

Prowlarr searchs multiple upstream search providers, thus allows to use that functionality through searx.

* Run sudo make run
* Go to /preferences
* Enable prowlarr search engine
* Make some search
* Assert that it works as expected

## Author's checklist

This uses the prowlarr search API; it does not support pagination.

## Related issues

(none)


searx/settings.yml
```diff
@@ -908,6 +908,26 @@ engines:
     engine : openstreetmap
     shortcut : osm
 
+  - name : prowlarr
```
kvch: Could you please comment out the engine? If I understood correctly it requires config from instance admins.

searx/engines/prowlarr.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+from json import loads
+from urllib.parse import urlencode
+from searx.exceptions import SearxEngineAPIException
+
+categories = ['files']
+
+paging = False
+api_key = 'None'
```
kvch: Why not set it to empty string?

kvch (commented):
Minor comments

searx/engines/prowlarr.py
narcoticfresh: @kvch 
ok, addressed your comments..

i wasn't quite sure what the correct approach for the "default values" are - every engine seems a bit different there.. but i changed it to empty strings now..

narcoticfresh:
@kvch i closed this in favor of #3118 as I had several unintended commits in this branch and I didn't want to poison the git history nor did i want to force push.. hope #3118 is alright now..

PR #3121, opened 2022-01-14
Fix searx up to work with modern httpx

andyljones:
## What does this PR do?

This is a slapdash upgrade of searx to a modern httpx version, since the one that's pinned in reqs is 18 months old.

The particular changes are 
* replace `allow_redirects` with `follow_redirects`
* fix up the `client.py` module

The old httpx version has some fun heisenbugs bugs in it that have been removed in the time since.

I'm not actually fussed about whether you merge this PR, but since I've done the work I figured I'd share it as a template if nothing else.

No tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ dist/
 local/
 gh-pages/
 searx.egg-info/
+.env
+geckodriver.log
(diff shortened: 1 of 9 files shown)


kvch (approved):
Thanks! Awesome update.

Issue #3110, opened 2022-01-03
HTTPX doesn’t seem to use the default root CA file/path locations


cyclaero:
I use searx as local instances on my desktop computers. I installed python together with `certifi` and the `DefaultVerifyPaths`and environement values are all set up correctly:

    python3 -c "import ssl; print(ssl.get_default_verify_paths())“

    DefaultVerifyPaths(cafile='/Library/Frameworks/Python.framework/Versions/3.9/etc/openssl/cert.pem', \
    capath='/Library/Frameworks/Python.framework/Versions/3.9/etc/openssl/certs', openssl_cafile_env='SSL_CERT_FILE‘, \
    openssl_cafile='/Library/Frameworks/Python.framework/Versions/3.9/etc/openssl/cert.pem‘, \
    openssl_capath_env='SSL_CERT_DIR', openssl_capath='/Library/Frameworks/Python.framework/Versions/3.9/etc/openssl/certs‘)

The file `cert.pem`is actually a symbolic link to the CA file which comes with certifi. So far so good. Now the problem is, that HTTPX does not use the configured CA paths, but directly picks the CA file from the certifi installation directory, which is `../../lib/python3.9/site-packages/certifi/cacert.pem`.

Besides that this is non-standard behaviour, additional root CA certificates cannot be simply added to the `capth`, actually we can add it, but it don’t become recognized. I need this when I want to use the local searx instance over my transparent HTTPS proxy.

As a workaround I simply cat’ed my `proxy.local` CA cert to the upstream CA file of certifi. I know, this will work only until the next certifi update. Best would be, if HTTPX could be configured to use the default SSL paths.

If this is not possible, can we have an option by `settings.yml` to ditch HTTPX and use the former HTTP client, this would be good anyway, because HTTPX is incompatible with Windows.

PR #3165, opened 2022-02-17
Fix Tor proxy when using httpx 0.21.x

MarcAbonce:
## What does this PR do?

This should fix #3164.
The problem is that `httpx` keeps making breaking changes to their library, so we just have to adjust the code a little bit to make it work with the new version of the library.

1. In the `settings.yml` add an outgoing proxy such as `all:// : socks5h://127.0.0.1:9050` and run Tor in that port.
1. Increase the `request_timeout` to make it up for Tor's slowness.
1. Run searx.
1. Try to run a query like `!onions searx`.
1. It should work.

## Related issues
Closes  #3164

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -167,8 +167,7 @@ class Network:
         for transport in client._mounts.values():  # pylint: disable=protected-access
             if isinstance(transport, AsyncHTTPTransportNoHttp):
                 continue
-            if not getattr(transport, '_pool') and getattr(transport._pool, '_rdns', False):
-                result = False
+            if getattr(transport, '_pool') and getattr(transport._pool, '_rdns', False):
                 continue
             return False
         response = await client.get("https://check.torproject.org/api/ip", timeout=10)


MarcAbonce:
Hmmm, this doesn't make any sense to me, but it looks like the value of `using_tor_proxy` is False inside the `settings_outgoing` dict but True outside that dict. This only happens here with Github CI, in my local that weird behaviour didn't happen.

For example, adding this code inside the initialize function in [searx/network/network.py](https://github.com/searx/searx/blob/master/searx/network/network.py#L251):
```python
    raise Exception({
        'extracted value': settings_outgoing.get('using_tor_proxy'),
        'settings': settings_outgoing,
    })
```
Prints [this](https://github.com/searx/searx/runs/5227699946?check_suite_focus=true):
`Exception: {'extracted value': True, 'settings': {'request_timeout': 3.0, 'useragent_suffix': '', 'pool_connections': 100, 'pool_maxsize': 20, 'enable_http2': True, 'using_tor_proxy': False, 'extra_proxy_timeout': 0}}`

kvch:
Oh, good old httpx. I will throw out httpx and revert back to requests...

kvch:
I am toying with the idea of reverting back to requests. httpx has issues on Windows https://github.com/searx/searx/issues/3110 and using async does not really differ from running requests in threads. The performance improvement we got is negligible.

kvch:
I have updated master to fix errors after httpx upgrade. Is this PR needed now?

MarcAbonce:
Thanks!
There's only one small fix that's left, though. I just rebased this PR to fix that.

kvch (approved):

kvch:
Thank you!

Issue #1622, opened 2019-06-18
Thoughts on Baidu, Sogou, Yandex via searx?


dm17:
I'm guessing this is the list of search engines searx can currently use as sources:
https://github.com/asciimoo/searx/wiki/possible-search-engines

"Possible" throws me off a bit; not sure if that means they can be used now because at the top it says, "which could be useful as engine and waiting for an implementation." So potentially unimplemented?

Anyway, I'd like to look into adding Baidu, Sogou, and Yandex. Thoughts?

rachmadaniHaryono:
i create yandex image but only tested for first page here

https://github.com/rachmadaniHaryono/searx/blob/master/searx/engines/yandex_images.py

i have no idea how to get next page

also there is also thread here https://github.com/asciimoo/searx/issues/430

virtadpt:
Those are indeed possible search engines, and not necessarily implemented at this time.  If you want to see an up-to-date list of the search engines supported by Searx, check out the settings.yml file: https://github.com/asciimoo/searx/blob/master/searx/settings.yml

jackTaw88:
@virtadpt I could not understant this topic. I can search on searx via yandex. How you say that is not implemented?

rachmadaniHaryono:
yandex image and yandex video are still not implemented yet

dalf:
Baidu : this issue is related to https://github.com/asciimoo/searx/issues/1432 : it is not possible to implement this engine because all links are encrypted. Yes searx can proxy the encrypted linked, but searx merges links between different search engine, and this step won't be possible for the Baidu engine.

Sogou : same as Baidu. Note : The encrypted links can't be compared with the one provided by Baidu.

dalf:
I close this issue: 
* Baidu / Sogou : look at comment https://github.com/asciimoo/searx/issues/1432#issuecomment-511137295
* Yandex : https://github.com/asciimoo/searx/issues/430

PR #3182, opened 2022-03-10
Add a new engine: Baidu

ShevonKuan:
## What does this PR do?
Add a new search engine https://www.baidu.com/
Baidu is a chinese search engine which is famous in China, so I think it is necessary to add it. 
Run `sudo make run`
Go to `/preferences`
Disable all search engines with the exception of onesearch (the last one)
Make some search
Assert that it works as expected
Assert that the pagination works
## Author's checklist
This search engine is OFF by default.
HTML parser from [python-baidusearch](https://github.com/wcadaydayup/python-baidusearch)
## Related issues
[#1622 ](https://github.com/searx/searx/issues/1622)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 certifi==2021.10.8
 babel==2.9.1
+beautifulsoup4==4.10.0
 flask-babel==2.0.0
 flask==2.0.2
 jinja2==3.0.3
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -398,7 +398,11 @@ engines:
     timeout : 3.0
     weight : 2
     tests: *tests_infobox
-
+  - name: baidu
+    engine : baidu
+    shortcut : bd
+    timeout : 2.0
+    disabled : True
   - name : duckduckgo
     engine : duckduckgo
     shortcut : ddg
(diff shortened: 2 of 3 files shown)


searx/settings.yml
```diff
@@ -398,7 +398,11 @@ engines:
     timeout : 3.0
     weight : 2
     tests: *tests_infobox
-
+  - name: baidu
```
kvch, line 401:
```suggestion

  - name: baidu
```

searx/settings.yml
```diff
@@ -398,7 +398,11 @@ engines:
...
+    engine : baidu
+    shortcut : bd
+    timeout : 2.0
+    disabled : True
```
kvch, line 405:
```suggestion
    disabled : True
    
```

requirements.txt
```diff
@@ -1,5 +1,6 @@
 certifi==2021.10.8
 babel==2.9.1
+beautifulsoup4==4.10.0
```
kvch, line 3: I would rahter not add a new dependency. Is there any reason why you used this lib instead of `lxml` like in other engines?

Example:
```python
from lxml import html
```

PR #3278, opened 2022-07-07
Add engine for Emojipedia

obfuscated-loop:
## What does this PR do?
This adds an [Emojipedia](https://emojipedia.org/) engine to searX, making it easy to find whatever emoji you want under a certain search term.

Emojis are commonly used for various types of expression, and aren't language specific. The problem is that not all systems have a place to search for an exact emoji - or even copy them. This is where Emojipedia comes in; it's essentially a Wikipedia of different emojis, which also, showcases how the emojis will look on different systems.

By using it as an engine within searX, you can copy from the direct search result or visit the page to learn more about the emoji as summarised.

`!em person`
This should return a search result including emojis with people in them.


kvch:
@james-still Do you mind fixing the error in the Python 3.9 job?

obfuscated-loop:
> @james-still Do you mind fixing the error in the Python 3.9 job?

Sorry for the late response. Just made the commit.

kvch:
I've tested the engine with the example query you provided, but I get no results back. Am I missing something?

obfuscated-loop:
> I've tested the engine with the example query you provided, but I get no results back. Am I missing something?

That's odd, maybe try a different query like 'math' - as you can see it's been merged into searXNG and seems to be working fine.

obfuscated-loop:
I think they might have updated their site and it's affecting the xPath values. I'll have a closer look in an hour.

obfuscated-loop:
All working now but leave it unmerged. I'll implement a relative xPath tonight so this doesn't happen again.

obfuscated-loop:
Should be good to merge now by the way :+1:

kvch:
Thank you!

kvch (approved):

PR #3285, opened 2022-07-13
[fix] make autofocus configurable (#1984)

### LINKED ISSUE #1984: Results page: having text cursor still in search field is annoying
If the search field is focussed (active) in the results page, any attempt to scroll down one page using the spacebar results in typing a space in the search field instead. None of the other search engines I use do that, and it's rather counter-intuitive.

bencollerson:
"Results page: having text cursor still in search field is annoying #1984"

## What does this PR do?

This patch allows the text area autofocus on the results page to be toggled either in the UI or in the searx server config.

This allows users to configure searx to not focus the text area on the results page allowing for more easy keyboard navigation when viewing searx results.

Run a local instance. Set the "Autofocus search field" setting to "Off". Perform a search and check the text area is not focused.

Also:

* test that default existing behaviour still works so that the search area on the results page is still focused when "Autofocus search field" is set to on
* test that the main searx page has its text field focused independent of this setting.

This setting's default value can also be tested by changing the `ui/autofocus` setting to `False` in the `searx/settings.yml` file.

## Related issues

Closes #1984


searx/static/plugins/js/vim_hotkeys.js
```diff
@@ -258,7 +258,7 @@ $(document).ready(function() {
     }
 
     function searchInputFocus() {
-        $('input#q').focus();
+        //$('input#q').focus();
```
kvch: Why is this commented out?

bencollerson: Oops will fix

kvch (approved):

kvch:
Thank you!

PR #3308, opened 2022-07-31
[enh] archive.today option for results page

bencollerson:
## What does this PR do?

Adds an option to the oscar theme that puts an archive.today link next to the web.archive.org cache link.

I find it useful to easily access archive.today links and thought other people might find it useful as well.

By default the option is off so the extra link should not appear on the results page.

In preferences change 'Show archive.today links' to 'On'

Do a search and on the results page test that the 'archive' links work.

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -354,6 +354,16 @@ class Preferences:
                     'True': True
                 }
             ),
+            'archive_today': MapSetting(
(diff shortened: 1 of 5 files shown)


kvch (approved):
Thank you!

bencollerson:
Haha cool. That's all I've got for now. Thanks for accepting all my patches.

PR #3332, opened 2022-08-08
Deepl

cy8aer:
## What does this PR do?

This implements the Deepl Translation engine

It works nearly like lingva but directly to the deepl api. There is a free option to use. See https://www.deepl.com/pro-api?cta=header-pro-api for registering a free account.

This api only needs a to-lang, from-lang is a fake by now

You will have deepl in searx with full phrase and sentence translations

Api key needed. Put api key into settings.yml - enable deepl. 

de-en Dies ist ein großartiger Test
en-es This is deepl.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1401,6 +1401,15 @@ engines:
     # See : https://mymemory.translated.net/doc/usagelimits.php
     # api_key : ''
 
+- name : deepl
+  engine : deepl
+  shortcut : dpl
+  # You can use the engine using the official stable API, but you need an API key
+  # See: https://www.deepl.com/pro-api?cta=header-pro-api
+  # api_key : '' # required!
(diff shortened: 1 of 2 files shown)


searx/settings.yml
```diff
@@ -1401,6 +1401,15 @@ engines:
     # See : https://mymemory.translated.net/doc/usagelimits.php
     # api_key : ''
 
+- name : deepl
```
kvch: As the engine cannot function without an API key, please comment it out.

cy8aer:
Uh, I hope I did the merging right. I need some routine for doing this.
For the patch: It is now commented out.

cy8aer:
Removed update branch from upstream.

cy8aer:
and rebased it to master. This should be the right way.

PR #3347, opened 2022-08-19
Tag privacy Violators

### LINKED ISSUE #1987: Filtering or tagging Cloudflare and other Reverse Proxy hoster
ATM I do not have time, but  at least I want to left a remark here:

It might be a good to have a plugin or any other option to tag and/or filter reverse proxy hoster like Cloudflare

This issue was inspired by the discussion at https://github.com/asciimoo/searx/issues/1965#issuecomment-632548230

SepehrRasouli:
## What does this PR do?
This PR adds the ability to tag privacy violating websites, Fixing Issue #1987. I grabbed the ASN data from [here](https://github.com/searxng/searx-space/blob/master/searxstats/data/asn.py), I also used IPWhois as it was used in [here](https://github.com/searx/searx-stats2/blob/45097dc7ebbaf284b2dc04b7f5051afb19571980/searxstats/fetcher/network.py#L120-L137). I used a dictionary as a cache so the IPWhois queries wouldn't slow down the searches drastically. 
I also changed the templates to show the tagged websites correctly.
This PR is motivated by Issue #1987.
```bash
make run
```
## Related issues
Closes #1987


searx/tagprivacyviolators.py
```diff
@@ -0,0 +1,225 @@
+import socket
+import ipwhois
+from searx import logger
+
+ASN_PRIVACY = {
+    # Akamai
+    "55770": "Akami",
```
unixfox: it's akamai not Akami

unixfox:
When doing a search this adds some latency for each new IP address right? I mean, it has to do a WHOIS before rendering the results to the user screen.

If so, IMO this feature just be configurable and disabled by default. It already takes quite an amount of time to do the requests to the configured engines, so adding even more latency will degrade the user experience.

SepehrRasouli:
I've implemented some kind of caching so when the user sends a query and sees some results, the privacy violators' URLs would be cached, not adding more latency the next time the user sends another query. 
But i think adding this a s a feature, disabled by default would be good.

unixfox:
> I've implemented some kind of caching so when the user sends a query and sees some results, the privacy violators' URLs would be cached, not adding more latency the next time the user sends another query. But i think adding this a s a feature, disabled by default would be good.

Adding some caching is minor in a case where the URL is always going to be different, when the user search on SearX it is very unlikely it will always search for the same exact thing. Also, most of the engines give different result URLs when you search for the same keyword.

The cache will hardly ever get used.

On top of that, due to the cache, this feature may decrease the privacy of the user because literally every search the user is doing is going to be saved in memory.

As discussed with @dalf about this feature, the best option for the WHOIS is to instead use a local database like this one: https://lite.ip2location.com/database-asn. But it will require a manual intervention from the user in order to setup this database. But using this database will be so much quicker to access than doing a WHOIS request to an external server.
This same WHOIS server can also log every website that you get in the results on SearX, harming even more the privacy of the user.

SepehrRasouli:
I agree with you, caching will harm the user's privacy and i will remove it.
But what do you think should we do about this ? I added it as a preference, disabled by default, so the requests wouldn't take long by default.

PR #3348, opened 2022-08-22
No change to variable name

Niveshkrishna:
There's no need to change variable name in translation

## What does this PR do?
Bug fix: Someone in translations team has changed variable name during translation. Because of this change in variable name, the app throws an exception KeyError 'minutos`


kvch:
We maintain the translations on Transifex. I've updated the string as you did and I am pulling it from Transifex now. Thanks for finding this issue!

PR #3361, opened 2022-09-12
[fix] brave engine: no search results

### LINKED ISSUE #3362: Brave engine (brave only) not working, yields no results.
Using official version 1.1.0 (latest) on OpenBSD and downloaded via git.

Using this how-to: https://www.tumfatig.net/2021/running-the-searx-metasearch-engine-on-openbsd/

**What happened?**
I don't know if it is a bug, but when I select Brave as the only search engine and do a search, it yields nothing and this message:
"Sorry! we didn't find any results. Please use another query or search in more categories."

**How To Reproduce**
Enable only brave engine, then do a search.

**Expected behavior**
Some results would be returned. =)

**Screenshots & Logs**
If needed, I can provide but it's basic.

**Additional context**
Question: Can I change something in settings.yml to resolve this?

rob4226:
There were never any results from the Brave search engine so fixed url_xpath and now Brave search results are working.

fixes: #3362

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1661,7 +1661,7 @@ engines:
     paging: true
     first_page_num: 0
     search_url: https://search.brave.com/search?q={query}&offset={pageno}&spellcheck=1
-    url_xpath: //div[@class="snippet fdb"]/a/@href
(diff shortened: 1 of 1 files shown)


kvch:
Thanks!

PR #3367, opened 2022-09-28
allow engines to override pretty_url and use this in bing to show mea…

### LINKED ISSUE #3335: Bing: don't show bing.com urls
bing shows these really useless bing.com urls in which it encodes the actual url. Google does this too but Google replaces the actual url with a Google url only when you visit (or copy) it. So these bing urls are really useless seeing in searx. I realize the code uses the same url for display and the href but there could be a property to override it, like `display_url` which could then be used in the bing engine

br4nnigan:
…ningful urls

## What does this PR do?

it parses the url that bing shows to the user and uses it in searx too as display url (or pretty_url as its called). To achieve this we just need to allow engines to override the pretty_url property that's created for every result

as I laid out in the linked issue I think not making the actual url visible to the user makes the bing engine unusable. I want to see the actual href before I click. 

Honestly I have not tested this and I can't even code python but it seems valid to me? please check it out, just a small change.

## Related issues

Closes #3335

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -68,11 +68,13 @@ def response(resp):
     for result in eval_xpath(dom, '//div[@class="sa_cc"]'):
         link = eval_xpath(result, './/h3/a')[0]
         url = link.attrib.get('href')
+        pretty_url = extract_text(eval_xpath(result, './/cite'))
         title = extract_text(link)
         content = extract_text(eval_xpath(result, './/p'))
 
         # append result
         results.append({'url': url,
+                        'pretty_url': pretty_url,
                         'title': title,
                         'content': content})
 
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
Great, thanks!

br4nnigan:
if 'url' in result and 'pretty_url' not in result:

pushed a fix, told you I cant code python ;)

br4nnigan:
@asciimoo did you see? there is a bug in the code, a test failed

br4nnigan:
so the patch is live apparently (including the fix) but it seems it doesn't work

can you tell me how to debug? If I make a change in the code (on my live instance) and rebuild the python cache the change doesn't seem to be used. I also restarted uwsgi service. Is there anything else I have to do?

PR #3397, opened 2022-10-30
Fix scraping of 'sc' value from homepage

fancypantalons:
Looking at the current HTML for the Startpage front page, the previous footer logo element is no longer present.  This change scrapes the "sc" parameter from one of the hidden HTML form elements, which should (hopefully) be a bit more stable long term, since that form is used by Startpage to submit requests to the engine.

## What does this PR do?

Prior to this change, the Startpage engine simply doesn't work because the scraping of the 'sc' value fails and raises an exception.  This change corrects the issue.

Without this change the engine crashes.

A basic search against Startpage triggers the issue.

### DIFF
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -91,8 +91,7 @@ def get_sc_code(headers):
         dom = html.fromstring(resp.text)
 
         try:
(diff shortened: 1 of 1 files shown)


kvch (approved):

kvch:
Thank you!

Issue #3441, opened 2023-01-03
DuckDuckGo blocked


Pistos:
**Version of Searx, commit number if you are using on master branch and stipulate if you forked Searx**

Running off `master` at commit 117dbd462f39fb4c

**How did you install Searx?**

Installed and run manually from source (not docker)

**What happened?**

After starting searx, the first search says in the UI that duckduckgo was blocked. The `debug: True` logging shows:

```
DEBUG:urllib3.connectionpool:https://lite.duckduckgo.com:443 "POST /lite HTTP/1.1" 403 None
DEBUG:urllib3.connectionpool:https://www.bing.com:443 "GET /search?q=language%3AEN+hello+world&first=1 HTTP/1.1" 200 None       
DEBUG:urllib3.connectionpool:https://search.yahoo.com:443 "GET /search?p=hello+world&ei=UTF-8&fl=1&vl=lang_en&btf=&fr2=time&age=&b=1&xargs=0 HTTP/1.1" 200 None                                                                                                 
DEBUG:searx:duckduckgo: ErrorContext('/home/searx/searx/searx/search/processors/online.py', 96, "response = req(params['url'], **request_args)", 'searx.exceptions.SearxEngineAccessDeniedException', None, ('HTTP error 403',))
ERROR:searx.search.processor.online:engine duckduckgo : Searx is blocked                                                        
Traceback (most recent call last):                                                                                              
  File "/home/searx/searx/searx/search/processors/online.py", line 144, in search                                               
    search_results = self._search_basic(query, params)                                                                          
  File "/home/searx/searx/searx/search/processors/online.py", line 124, in _search_basic                                        
    response = self._send_http_request(params)                                                                                  
  File "/home/searx/searx/searx/search/processors/online.py", line 96, in _send_http_request
    response = req(params['url'], **request_args)                                                                               
  File "/home/searx/searx/searx/poolrequests.py", line 223, in post                                                             
    return request('post', url, data=data, **kwargs)                                                                            
  File "/home/searx/searx/searx/poolrequests.py", line 202, in request                                                          
    raise_for_httperror(response)                                                                                               
  File "/home/searx/searx/searx/raise_for_httperror.py", line 62, in raise_for_httperror
    raise SearxEngineAccessDeniedException(message='HTTP error ' + str(resp.status_code),
searx.exceptions.SearxEngineAccessDeniedException: HTTP error 403, suspended_time=86400
```

[...]

PR #3444, opened 2023-01-07
DuckDuckGo fixes

### LINKED ISSUE #2035: "Page not found": displayed the requested path.
Currently, when a unknown page is requested, searx displays this:
image

It would be helpful to show the requested path.

Example: https://github.com/asciimoo/searx/issues/2033 . If the requested path is shown on the HTML page, it would help to adjust the NGINX configuration accordingly.

### LINKED ISSUE #2080: Brainstorming: on_result plugin / https_rewrite plugin
Different plugin ideas require to filter / modify the results according to the host of the URL. 

Usage examples:
* the https_rewrite plugin, and the current implementation fails to do so: https://github.com/asciimoo/searx/tree/master/searx/plugins/https_rules are outdated, and the number of rules is a fraction of the actual list to avoid an overhead. Here there is one lookup per .xml files: https://github.com/asciimoo/searx/blob/6e37255e59693d7a3768491b2b8d06441051131f/searx/plugins/https_rewrite.py#L140
* [Add support for Youtube and Twitter links rewrite PR #1726](https://github.com/asciimoo/searx/pull/1726) rewrite some URL too.
* [feature request: private user blocklists/blacklists #2001](https://github.com/asciimoo/searx/issues/2001) asks to remove some domain names from the results.

Currently, ```on_result``` plugin scan for each URLs and process them. 
What if we provide provide each plugin register for a list of domain.

Some technical solutions to create a static list of domains:
* a [trie](https://en.wikipedia.org/wiki/Trie). (This issue superseed this one https://github.com/asciimoo/searx/issues/936 )
* one hash table per label in the domain name: https://github.com/sethmlarson/hstspreload

According to https://www.eff.org/https-everywhere/rulesets#wildcard-targets the wildcard in HTTPS everywhere can be either:
* *.eff.org
* www.eff.*

A solution: two lookups in two different structures, one in reversed order (org -> eff -> *), one in normal order (www -> eff -> *).

Prototype: https://gist.github.com/dalf/26462c1627c8b00fdfacdd55525b08b6

The idea is provide: 
* a format to store the rules (either a binary or plain text format), with a module to create this output format.
* a plugin can register this structure: the searx core calls the plugin where there is a match.
* it is difficult to merge all trie without a lot of repetition, some most probably one "double trie" per plugin; but in simple cases I guess it can be done.

Note about the issue #2001:
* the choice of domains to remove is a preference, then the issue is not solved.
* if we can make some maintained list, then the issue is solved.

cyclaero:
## What does this PR do?

Fix duckduckgo blocking: add trailing slash to duckduckgo url (= back ported fix #2035 from https://github.com/searxng/searxng/)

Fix duckduckgo blocking: Add HTTP Referer header to DuckDuckGo requests (= back ported fix #2080 from https://github.com/searxng/searxng/)

This makes duckduckgo work again.

See: DuckDuckGo blocked #3441

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -73,6 +73,7 @@ def request(query, params):
     # link again and again ..
 
     params['headers']['Content-Type'] = 'application/x-www-form-urlencoded'
+    params['headers']['Referer'] = 'https://lite.duckduckgo.com/'
 
     # initial page does not have an offset
     if params['pageno'] == 2:
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -18,7 +18,7 @@ from searx.poolrequests import get
 
 # about
 about = {
-    "website": 'https://lite.duckduckgo.com/lite',
+    "website": 'https://lite.duckduckgo.com/lite/',
     "wikidata_id": 'Q12805',
     "official_api_documentation": 'https://duckduckgo.com/api',
     "use_official_api": False,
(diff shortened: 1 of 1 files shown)


kvch (approved):

kvch:
Thank you!

PR #3465, opened 2023-02-09
Add pagination to Wiby

wibyweb:
## What does this PR do?

Makes pagination work on Wiby by adding the page (p) parameter and removing the old offset (o) parameter that wasn't compatible with {pageno}.

Makes pagination work for Wiby.

See if the pages change when searching Wiby.


wibyweb:
@kvch thank you!

PR #3472, opened 2023-02-22
Use packaging instead of distutils

### LINKED ISSUE #3462: remove distutils usage
In Python 3.10 and 3.11, distutils has been formally marked as deprecated. Code that imports distutils will no longer work from Python 3.12.

`searx_extra/update/update_firefox_version.py` still uses distutils. Maybe its use of `distutils.version` can be replaced by the `packaging` module?

See https://peps.python.org/pep-0632/

glanham-jr:
## What does this PR do?

Removes `distutil` usage in `searx_extra/update/update_firefox_version.py`

`distutil` will become deprecated and no longer in use in Python 3.12

I personally ran the file directly in python, only running the necessary functions and returning the results.

## Author's checklist

Closes  #3462

### DIFF
--- a/searx_extra/update/update_firefox_version.py
+++ b/searx_extra/update/update_firefox_version.py
@@ -5,7 +5,7 @@ import requests
 import re
 from os.path import dirname, join
 from urllib.parse import urlparse, urljoin
-from distutils.version import LooseVersion, StrictVersion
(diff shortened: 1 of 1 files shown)


kvch:
Thanks!

PR #3477, opened 2023-02-27
Reword the TLDR as it is misleading

unixfox:
The current TLDR in the README is misleading as it applies that SearXNG is not suited for users that self-host, but that is not true. There are plenty of ways to easily self-host SearXNG: https://docs.searxng.org/admin/installation.html

I reworded, so it reflects the current situation between the development of SearXNG and Searx. Please suggest any modification if the words used doesn't suit you.

Please see the examples of the misleading:
- https://github.com/searx/searx/discussions/3475#discussioncomment-5106229
- https://github.com/searx/searx/discussions/3457#discussioncomment-4884694

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,8 +85,8 @@ of turning it off.
 What is the difference between searx and SearxNG?
 #################################################
 
-TL;DR: If you want to run a public instance, go with SearxNG. If you want to
-self host your own instance, choose searx.
+TL;DR: SearXNG is for users that want more features and bugs getting fixed quicker.
+If you prefer a minimalist software and stable experience, use searx.
 
 SearxNG is a fork of searx, created by a former maintainer of searx. The fork
 was created because the majority of the maintainers at the time did not find


kvch (commented):
It is true that searx is moving slower than searxng. But I suggest people to use searx who want to run it locally because searxng comes with many irrelevant features to self hosters. We can add that searxng moves faster, but I do not want to change my recommendation regarding self hosting.

README.rst
```diff
@@ -85,8 +85,8 @@ of turning it off.
 What is the difference between searx and SearxNG?
 #################################################
 
-TL;DR: If you want to run a public instance, go with SearxNG. If you want to
-self host your own instance, choose searx.
+TL;DR: SearXNG is for users that want new features and bugs fixed quicker, whereas
```
kvch: Maybe this would be better:
```
TL;DR: SearXNG is for users that want more features (even at the expense of privacy) and bugs getting fixed quicker. If you want a minimalist and stable experience, use searx.
```

Or
```
TL;DR: SearXNG is for users that want more features aimed at instance administrators and bugs getting fixed quicker. If you prefer a minimalist software and stable experience, use searx.
```
WDYT?

return42: We at the SearXNG team have always held back and never rated searx or made any claim as to what audience it may or may not be suitable for.  We would be very grateful if this could be practiced similarly in the searx team.

The whole chapter is nonsense, the ratings do not reflect the experience of SearXNG users and the statements about target groups are the very personal opinion of the searx team.  We from the SearXNG team have a completely different opinion.

Please stop confusing people, everyone can make his own picture and will then realize which of the products is more suitable for him and with which team he gets the fastest help.

unixfox:
@kvch I've chosen your first suggested sentence without the privacy part nor the instance administrator part, this way we stay "general":
> TL;DR: SearXNG is for users that want more features and bugs getting fixed quicker. If you want a minimalist and stable experience, use searx.

I think it should suit everyone and seem to represent well the current state of searx and searxng in a very few words.

README.rst
glanham-jr: As someone who has recently started looking at the two projects - I would say it's important to highlight what each project does, as my decision to choose which one was slightly confusing.

I think a mutually agreed upon comparison table might be nice in determining which one to choose. Below is an example that could be expanded on. Again, below is an example of what could be expanded on.


|   Project Structure    | [searx](https://github.com/searx/searx) | [SearxNG](https://github.com/searxng/searxng) |
| ----------- | ----------- | ------ |
| Release Cycle      | Stable       | Fast |
| New Features   | Yes[^1] |  Yes|
| New Engines | Yes | Yes |



|   Features     | [searx](https://github.com/searx/searx)  | [SearxNG](https://github.com/searxng/searxng) |
| ----------- | ----------- | ------ |
|  Meta-Searching   | Yes| Yes |
| Telemetry | No | Yes[^2]
| Translations | Yes | Yes - More Coverage |
|  JavaScript | Not Required | Required / Highly Recommended (?) |

[^1]: Excludes possible User Privacy Violations deemed by searx maintainers

[^2]: Opt-in by default. Can be [disabled on the server](https://docs.searxng.org/admin/engines/settings.html#general)

I also think further describing what aspects of privacy is deemed acceptable vs encroaching on perceived users would also be good (and what can be disabled). Seems the main difference is in telemetry, but there could be more I'm not aware of. These clear definitions would definitely help administrators determine which project is more acceptable. Privacy isn't binary, and vague descriptions of "at the cost of privacy" doesn't really help inform an individual of which choice to make effectively.

unixfox: Telemetry is the wrong word here, nothing is sent to a distant server, everything stay locally on the searxng instance.
See the definition of telemetry on Wikipedia: https://en.wikipedia.org/wiki/Telemetry
Metrics or (a similar word) is the best because all of what it does is to show based on the usage of the user, the reliability of an engine.

You do not need to enable JavaScript on searxng, it works fine without it. (Did you test it?)

And what define "new features" and "new engines"? To this day searx stopped adding new features a few years ago and there are many more engines in searxng than searx.

You also have not really imply the time to fix a bug (mainly engines) in both projects, usually when a new engine is broken it is fixed under a week on searxng and searx merge the bug fix from searxng usually after 3-4 weeks.

glanham-jr: It was just an example - but I think your reply provides a great reason why a comparison table would be quite nice! My quick table was derived from exclusively reading the READMEs of the two projects and has ambiguity.

1. I would consider it "telemetry" if JavaScript was used to send information back to the server from the client. If the metrics are just gathered server-side only, then I agree its only "metrics". For me personally, this isn't a privacy violation if fingerprinting like information isn't gathered, like IP address and etc.
2. The searx README has the following clause which lead to the statement on JavaScript[^1]. SearxNG has not specifically stated they do not rely on JavaScript, hence the (?) in the original table.
3. Adding "New Features" would be any additional functionality that is currently not implemented. "New Engines" is adding new engines that are currently not implemented. The searx README claims to still accept new features and engines, but based on your comment: in practice it appears that feature adoption rate is extremely low if at all, and engines picked from SearxNG are also extremely low (?)
4. Time to Engine Issue Resolution - would be a good thing to add to the table. Average < 1 week vs < 1 month

Some further things to add:
| Features | Searx | SearxNG |
|----------|--------|----------|
| Image Proxy | Requires Morty | Built in (can't tell from README) |
| Bot Blocker | Requires Filtron | Built in |
| URL Blocker / Replacer | No  | Yes |
| Dark Mode | No  | Yes |
| Right-to-Left Language Support | No | Yes |

I really like a comparison table, and for more details / philosophy, the README can be consulted.

[^1]: We are picking engine changes from SearxNG, but we are not implementing engine detailed monitoring and not adding a new UI that relies on Javascript.

kvch (approved):

unixfox:
I unfortunately don't have the permissions to merge.
You will have to do it for me.

unixfox:
@kvch could you look into merging this PR? thank you

PR #3483, opened 2023-03-09
Initial integration tests

glanham-jr:
## What does this PR do?

Creates a base integration test suite to quickly test search results of a set of listed engines.

This is a PR in progress and is meant to aid in diagnosing certain engine issues.

Allows for quick iteration of engine development for selected engines (more could be added).

Run the following command in your terminal

`make test.int`

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ search.checker.%: install
 	$(Q)./manage pyenv.cmd searx-checker -v "$(subst _, ,$(patsubst search.checker.%,%,$@))"
 
 PHONY += ci.test test test.sh
-ci.test: test.pep8 test.pylint test.unit test.robot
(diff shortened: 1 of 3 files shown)


kvch:
Is this a similar checker to searx-checker? https://github.com/searx/searx-checker

glanham-jr:
Yes it is, didn't know that existed. While these integrations tests don't have to be ran for CI/CD, it provides a way to quickly test engines locally at the run of a command.

Some things I want to do:
1. Pass in what engine(s) you want to run on the command line.
2. Only a subset of engines are supported, specifically those that require additional setup. Haven't planned how to structure that, but was thinking to pull setting from a config file.

With this, if the integration tests are stable, it might be nice to include them in CI for the most common engines (even if not failing the build) so we are alerted on engine issues much faster.

PR #3501, opened 2023-04-25
Support flask-babel 3.0

eryg-kai:
## What does this PR do?

Allows searx to run with flask-babel 3.0.  The driving motivation was to fix the package on NixOS which is not honoring dependency versions (https://github.com/NixOS/nixpkgs/issues/216695).

The error:

```
AttributeError: 'Babel' object has no attribute 'localeselector'
```

The fix:

The way the locale selector is added has changed somewhat, and I needed to move `get_locale` up to make use of it.

I just want it to work with NixOS but keeping dependencies up to date is good in general.

I just ran `make test` with and without the updated requirements to make sure searx runs with both 2.0 and 3.0 without any errors.

## Related issues

Obviates https://github.com/searx/searx/pull/3497

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -141,11 +141,47 @@ if not werkzeug_reloader\
    or (werkzeug_reloader and os.environ.get("WERKZEUG_RUN_MAIN") == "true"):
     search_initialize(enable_checker=True)
 
-babel = Babel(app)
+ui_locale_codes = [l.replace('_', '-') for l in settings['locales'].keys()]
+
+
+def get_locale():
+    if 'locale' in request.form\
+       and request.form['locale'] in settings['locales']:
+        # use locale from the form
+        locale = request.form['locale']
+        locale_source = 'form'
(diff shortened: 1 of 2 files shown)


searx/webapp.py
```diff
@@ -141,11 +141,47 @@
...
+
+    return locale
+
+
+babel = Babel()
+if hasattr(babel, "localeselector"):
+    babel.init_app(app)
+    babel.localeselector(get_locale)
+else:
+    babel.init_app(app, locale_selector=get_locale)
```
jecarr, line 181: *Only commenting as I came across this recently too

If you're changing the versions in requirements.txt to be fixed (i.e. `3.1.0` which throws the AttributeError), is it worth having this if-else?

I'd think you either keep this approach but cater for (include) a lower version in the requirements file or have a cleaner diff (mine was [this](https://github.com/arachne-threat-intel/searx/commit/96c3dff)).

